Issue #108, opened 2013-05-01
Support for HTTP CONNECT proxy


OferE:
Hi,
When I use a proxy server to access EC2 I get "Signature" problem error.
When I use the same code without the proxy it works fine.
I also checked other functionality of EC2  and the behaviour is the same.
Thanks.

Here is my code (written as a mocha test) - with the proxy settings:

```js
var AWS = require('aws-sdk');
AWS.config.update({
   region: 'us-east-1',
   'httpOptions': {
      proxy: 'http://myProxy.com:8080' // I changed the proxy details for the github issue
   },
   sslEnabled: true
});

describe('checking amazon API', function() {

   it('should launch instances on amazon', function(done) {
      this.timeout(120000);
       var params = {
          ImageId: 'ami-1a249873', // changed the ami for the github issue
          MinCount: 1,
          MaxCount: 1,
          InstanceType: 't1.micro'
       },
       ec2 = new AWS.EC2({maxRetries: 15});

      ec2.runInstances(params, function(error, result) {
         console.log(this.request.httpRequest);
         console.log('result: ', result);
         console.log('error: ', error);
         done();
      });
   });
});
```

Here is the output I get from the console.log:

```js
{ method: 'POST',
  path: '/',
  headers: 
   { 'User-Agent': 'aws-sdk-nodejs/v0.9.8-pre.9 linux/v0.8.18',
     'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8',
     'Content-Length': 283 },
  body: 'AWSAccessKeyId=***&Action=RunInstances&ImageId=ami-00fe0569&InstanceType=t1.micro&MaxCount=1&MinCount=1&Signature=***&SignatureMethod=HmacSHA256&SignatureVersion=2&Timestamp=2013-05-01T07%3A33%3A00.120Z&Version=2013-02-01',
  endpoint: 
   { protocol: 'https:',
     slashes: true,
     host: 'ec2.us-east-1.amazonaws.com',
     hostname: 'ec2.us-east-1.amazonaws.com',
     href: 'https://ec2.us-east-1.amazonaws.com/',
     pathname: '/',
     path: '/',
     port: 443,
     constructor: { [Function: Endpoint] __super__: [Function: Object] } },
  region: 'us-east-1',
  params: 
   { Version: '2013-02-01',
     Action: 'RunInstances',
     ImageId: 'ami-00fe0569',
     MinCount: '1',
     MaxCount: '1',
[...]

Issue #26, opened 2013-01-07
Proxy support


gramakri:
Allow the API to be used with  HTTP and Socks Proxies.

yamsellem:
Using `https-proxy-agent` worked for me.

```
var ProxyAgent = require('https-proxy-agent');

var options = {
  accessKeyId: '..', 
  secretAccessKey: '..',
  httpOptions: {agent: new ProxyAgent({host, port})}
};
client = s3.createClient({s3Options: options});
```

restebanez:
http://blogs.aws.amazon.com/javascript/post/Tx3CV0OSVN2TBPS/Using-the-AWS-SDK-for-JavaScript-from-Behind-a-Proxy

ivarrian:
@yamsellem This is the solution that helped me! Thanks.

license2e:
@yamsellem Thank you! This helped me as well...

alain-odea-vgh:
Got hit by this too.  My Squid proxy was refusing access, returning HTTP 403 itself, and presenting a user-facing access denied HTML page.  Not sure if there's a reliable way to differentiate that from a AWS API error. The Content-Type is text/html regardless of Accept headers, but I suppose that isn't enforceable for S3 since each Object can have a custom Content-Type including text/html.

PR #355, opened 2014-09-10
Proxy

nickdk:
Had some problems with using a proxy while uploading code through the grunt-aws-s3 module, apparently the proxy at our company was trying to use the aws secret in the Authorization header as the proxy credentials which of course always fails and results in a http 407. This seems to be fixed if the Proxy-Authorization header is set with the actual proxy credentials. Not entirely sure but it could be that these issues are related and possibly fixed:

https://github.com/aws/aws-sdk-js/issues/108
https://github.com/aws/aws-sdk-js/issues/26

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ doc
 test/browser/build
 test/browser/sample/appinfo.js
 dist/aws-sdk-all.js
+.idea/
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -36,9 +36,8 @@ AWS.NodeHttpClient = AWS.util.inherit({
       options.agent = this.sslAgent();
     }
 
-    if (httpOptions.proxy) {
-        options.headers['Proxy-Authorization'] = 'BASIC ' +new Buffer(endpoint.auth).toString('base64');
-        console.log('headers: ' +JSON.stringify(options.headers));
+    if (httpOptions.proxy && endpoint.auth !== undefined) {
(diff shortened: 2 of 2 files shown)


AdityaManohar:
@nickdk thanks for the pull request!

We've deliberated on this issue for some time, and given the many different proxy server implementations and protocols, we think the best way to support proxies is to configure the SDK to use an [`http.Agent`](https://nodejs.org/api/http.html#http_class_http_agent) that implements proxy support.

There a number of `http.Agent` implementations available like [proxy-agent](https://www.npmjs.com/package/proxy-agent) and [tunnel-agent](https://www.npmjs.com/package/tunnel-agent) that do implement proxy support. The SDK can easily be configured to use these implementations instead of the default native `http.Agent`.

We've documented how to configure the SDK with an `http.Agent` implementation that supports proxies in our [developer guide](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html#Configuring_a_Proxy). We've also written a blog post that describes [using the SDK from behind a proxy](http://blogs.aws.amazon.com/javascript/post/Tx3CV0OSVN2TBPS/Using-the-AWS-SDK-for-JavaScript-from-Behind-a-Proxy).

I'm going to mark this as closed because there is no plan to build proxy support in the SDK at this time.

Feel free to open an issue if you have any questions, and thanks again!

nickdk:
Thanks @AdityaManohar, forgot that this pull request was actually still open. I think we're still using my fork at my company to support our proxy, so thanks for the developer guide to do this properly. We'll start using this solution so we can ditch my fork as soon as possible.

PR #448, opened 2014-12-18
With a lot of concurrent requests to metadata service, credentials can fail

### LINKED ISSUE #445: CredentialsError: Missing credentials in config
Originally at https://github.com/aws/aws-sdk-js/issues/205#issuecomment-66229958.

Here is the error I am seeing when using the SDK in high concurrency applications:

```
/usr/local/src/metadata-creds/node_modules/aws-sdk/lib/request.js:32
          throw err;
                ^
CredentialsError: Missing credentials in config
    at createHangUpError (http.js:1476:15)
    at Socket.socketOnEnd [as onend] (http.js:1572:23)
    at Socket.g (events.js:180:16)
    at Socket.emit (events.js:117:20)
    at _stream_readable.js:943:16
    at process._tickDomainCallback (node.js:463:13)
```

I made an isolated test case for this: https://gist.github.com/willwhite/6afe2d333a002c33413f. The trick to replicating appears to be in adjusting `http.globalAgent.maxSockets` to a value around 100 (the default is 5 but that's too low for high concurrency applications).

It seems like we need a locking cache around [`MetadataService.loadCredentials()`](https://github.com/aws/aws-sdk-js/blob/6fc999b0644eb71518eb00a3445f3e1472b834e0/lib/metadata_service.js#L86-L107). To avoid bombarding the metadata service with requests the moment credentials expire.

Happy to work on a PR if there is agreement.

mick:
This is encountered as `CredentialsError: Missing credentials in config` because the request for IAM Role can timeout before all the requests are answered. fixes #445 

The fix:
 - add a locking so the metadata service isnt flooded with requests.
 - add setImmediate so IO can still be processed even with a lot of request to the metadata service being generated.

This is a bit hard to add a test for. I did [add a test](https://github.com/mick/aws-sdk-js/compare/aws:master...mick:creds_lock?expand=1#diff-c5a12adba23f3b82dcb6d33d6980c069R42) for multiple concurrent requests to `loadCredentials`.

On an ec2 of the same type and size: [This test case](https://gist.github.com/willwhite/6afe2d333a002c33413f) passed with these changes, even with increasing concurrency up to 10000.

cc: @lsegal @willwhite

### DIFF
--- a/lib/metadata_service.js
+++ b/lib/metadata_service.js
@@ -67,11 +67,18 @@ AWS.MetadataService = inherit({
     var httpRequest = new AWS.HttpRequest('http://' + this.host + path);
     httpRequest.method = 'GET';
 
-    http.handleRequest(httpRequest, this.httpOptions, function(httpResponse) {
-      httpResponse.on('data', function(chunk) { data += chunk.toString(); });
-      httpResponse.on('end', function() { callback(null, data); });
-    }, callback);
+    process.nextTick(function() {
+      http.handleRequest(httpRequest, this.httpOptions, function(httpResponse) {
+        httpResponse.on('data', function(chunk) { data += chunk.toString(); });
+        httpResponse.on('end', function() { callback(null, data); });
+      }, callback);
+    });
   },
+
+  /**
+  * @api private
+  */
+  loadCredentialsCallbacks: [],
 
   /**
    * Loads a set of credentials stored in the instance metadata service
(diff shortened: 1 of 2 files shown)


mick:
@lsegal I think this is all set to go, unless there is anything else you think is blocking. Thanks! :octocat:

lsegal:
Looks good, thanks!

mick:
@lsegal great! Just curious, what does the release / publish cycle for aws-sdk look like?  I'd like to be able to use this from npm.

lsegal:
We have a release scheduled this week. FWIW you can use this from npm already if you `npm install git://github.com/aws/aws-sdk-js`.

paritosh16:
I have a number of node modules that are running as cron jobs. The modules use the aws-sdk for javascript (Installed using `npm install aws-sdk`, current version being `2.1.38`). It is giving the following error randomly,

    { [TimeoutError: Missing credentials in config]
      message: 'Missing credentials in config',
      code: 'CredentialsError',
      time: Mon Jul 13 2015 09:20:02 GMT+0000 (UTC),
      originalError:
       { message: 'Could not load credentials from any providers',
         code: 'CredentialsError',
         time: Mon Jul 13 2015 09:20:02 GMT+0000 (UTC),
         originalError:
          { message: 'Connection timed out after 1000ms',
            code: 'TimeoutError',
            time: Mon Jul 13 2015 09:20:02 GMT+0000 (UTC) } } }

This occurs randomly for all the modules. After the error, it again works fine for some time with another occurence of the error. Please help.

PR #450, opened 2014-12-20
Do not use ternary in a require call

jjjjw:
Webpack has trouble with this: https://github.com/webpack/webpack/issues/627

This could be seen as a bug in webpack, but a fix here would be very helpful.


joearasin:
Is there any reason not to remove multiRequire entirely? Unless I'm missing something, it looks like dead code.

PR #456, opened 2015-01-06
add error handling to list buckets example

jaxgeller:
Hey great work on the library, loving it so far, just one minor change. I ran the example of listing all buckets off of http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-examples.html#Amazon_S3__List_All_of_Your_Buckets__listBuckets_ and noticed if you don't have any buckets it will throw a type error because buckets will return as null, so I added some simple error checking to the documentation so that it will handle a situation with no buckets.
I didn't see a pattern in the way you formatted the control flow, so if there's a problem with inline if-statements let me know and I'll be happy to fix it. 

Also cleared a trailing whitespace.

### DIFF
--- a/doc-src/guide/node-examples.md
+++ b/doc-src/guide/node-examples.md
@@ -4,7 +4,7 @@
 
 All of these examples assume that the AWS library is required,
 credentials are loaded via environment variables (`AWS_ACCESS_KEY_ID`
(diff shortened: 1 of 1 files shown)


jaxgeller:
Changed that up, should be good to go.

lsegal:
Thanks for the doc update!

PR #472, opened 2015-01-20
Fix `AWS.S3.ManagedUpload.fillBuffer()` to use correct buffer indices.

### LINKED ISSUE #470: ManagedUpload.fillBuffer doesn't seem to work with Node v0.8
I'm using v2.1.6 of this library. Locally, I was using Node v0.10, and could use `s3.upload` to upload binary images. However, in production, we're using v0.8, and upload would fail with:

```
Error: oob
  at Buffer.slice (buffer.js:552:32)
  at ManagedUpload.fillBuffer (node_modules/aws-sdk/lib/s3/managed_upload.js:328:30)
  at ManagedUpload.AWS.S3.ManagedUpload.AWS.util.inherit.send (node_modules/aws-sdk/lib/s3/managed_upload.js:172:33)
  at [object Object].upload (node_modules/aws-sdk/lib/services/s3.js:480:50)
```

After much trial and error, I narrowed it down to a `sliceFn` call: https://github.com/aws/aws-sdk-js/blob/fc04887905f4a8dca1a96e1799c2f8bedab1cd01/lib/s3/managed_upload.js#L328

I was able to work around this by setting up my own `ManagedUpload` and overwriting `sliceFn`. Here it is in CoffeeScript:

``` coffee
bufferImg = new Buffer(blob.content, 'base64')
data =
  Key:  "#{S3_IMAGES_PATH}/#{image.path}"
  Body: bufferImg
  ContentEncoding: 'base64'
  ContentType: 'image/png'
  ContentLength: bufferImg.length
  ACL: 'public-read'

upload = new AWS.S3.ManagedUpload({service: @s3, params: data})
upload.sliceFn = {}
upload.sliceFn.call = (a, b, c) => bufferImg
upload.send(callback)
```

AdityaManohar:
Fixes #470

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -324,7 +324,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
     var self = this;
     var bodyLen = byteLength(self.body);
     while (self.activeParts < self.queueSize && self.partPos < bodyLen) {
-      var endPos = self.partPos + self.partSize;
+      var endPos = Math.min(self.partPos + self.partSize, bodyLen);
       var buf = self.sliceFn.call(self.body, self.partPos, endPos);
       self.partPos += self.partSize;
 


AdityaManohar:
Did not add a test case for this because this is an internal api implementation.

PR #474, opened 2015-01-21
Use `getCredentialsForIdentity()` when a `RoleArn` is not passed in as a parameter

AdityaManohar:
Use `getCredentialsForIdentity()` when a `RoleArn` is not passed in as a parameter to `AWS.CognitoIdentityCredentials`


AdityaManohar:
Any chance you could get some eyes on this?

Issue #383, opened 2014-10-13
Support using the SDK with browserify


konklone:
Right now, the AWS SDK for JavaScript _is_ usable by browserify, but with constraints:
- You have to download the minified version from the CDN, not use npm. If you use the version from npm, you get an error:

![error-1](https://cloud.githubusercontent.com/assets/4592/4608791/6e0a38b0-5289-11e4-949e-4a15d7ed62be.png)

From this place:

![error](https://cloud.githubusercontent.com/assets/4592/4608792/72d836f8-5289-11e4-941d-b68b2e9c8b2e.png)
- When freezing aws-sdk into your project, you can only use the minified version, not the full version. I'm not sure what the processing is for the minified version, though I can see it uses many less `require()` calls.
- When freezing the minified version of aws-sdk into your project, you have to add the following to the bottom of the .js file:

``` javascript
module.exports = AWS;
```

The fact that I can get it working via browserify like this suggests it shouldn't be a huge lift for the project to support browserify via npm.

The benefit of using it via browserify, compared to using it in a more traditional way, is that it makes it possible to combine the AWS SDK with other Node-packaged modules in npm (including those designed to work with this library in Node-land). 

In my case, I'm connecting [`filereader-stream`'](https://github.com/maxogden/filereader-stream) to [`s3-upload-stream`](https://github.com/nathanpeck/s3-upload-stream) in the browser to pipe data from disk directly into S3 using the Multipart Upload API. Stream syntax is incredibly helpful at gluing complicated workflows together in a simple way, and I wouldn't be able to pull this off without using browserify.

doapp-ryanp:
@lsegal I'm working on using browserify in nodejs, with the output being targeted at a lambda function.  FWIW I'm doing this as part of my work on the new [JAWS project](https://github.com/jaws-stack/JAWS/tree/v1.0-deploy-lambda) - essentially making it easier to use Lambda/api gateway etc.

Anyways, we'd like to use browserify to keep the footprint small (to combat the well documented cold start problems).  We also want to bundle aws-sdk for node so we can target specific versions and not rely on what is exposed in the lambda nodejs runtime.
 
We are concerned with the dynamic require brought up in this issue, however it looks like you have it solved in >= v2.0.20

[...]

PR #506, opened 2015-02-12
Fix path matching for non-*nix environments.

### LINKED ISSUE #505: JavaScript SDK build on WIndows does not include any AWS service
Building the SDK on a Windows 8 system with node.js v0.12.0 does not include the AWS services in the output SDK js file.

For further details see https://forums.aws.amazon.com/thread.jspa?threadID=171478&tstart=0

AdityaManohar:
Fixes #505


AdityaManohar:
Referencing #383

Issue #528, opened 2015-03-10
SQS Inaccessible host


ottopoika:
For some reason sqs fails time to time with error message:
Inaccessible host: `sqs.eu-west-1.amazonaws.com'. This service may not be available in the `eu-west-1' region.

This is weird because the script starts fine and gets messages correctly but after 1 or 2 days it throws  the mentioned message.

AdityaManohar:
This error is thrown when the SDK encounters a lower level `ENOTFOUND ` error. This is usually because of a transient network or DNS issue.

Would you be able to verify network / DNS connectivity of the environment running your code? I'd be interested to know what you find.

Hope this helps!

mick:
I have got this error to on a few requests to DynamoDB, from an several EC2s in the same region. 

```
UnknownEndpoint: Inaccessible host: `dynamodb.us-east-1.amazonaws.com'. This service may not be available in the `us-east-1' region.
    at Request.ENOTFOUND_ERROR (/usr/local/src/app/node_modules/aws-sdk/lib/event_listeners.js:358:46)
    at Request.callListeners (/usr/local/src/app/node_modules/aws-sdk/lib/sequential_executor.js:100:18)
    at Request.emit (/usr/local/src/app/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/usr/local/src/app/node_modules/aws-sdk/lib/request.js:604:14)
    at ClientRequest.error (/usr/local/src/app/node_modules/aws-sdk/lib/event_listeners.js:204:22)
    at ClientRequest.<anonymous> (/usr/local/src/app/node_modules/aws-sdk/lib/http/node.js:62:19)
    at ClientRequest.emit (events.js:117:20)
    at CleartextStream.socketErrorListener (http.js:1551:9)
    at CleartextStream.emit (events.js:95:17)
    at Socket.onerror (tls.js:1455:17)
```

Since this seems caused by intermittent DNS request failures, shouldnt this error be `retryable:true`? https://github.com/aws/aws-sdk-js/blob/master/lib/event_listeners.js#L362

cc @lsegal

lsegal:
That's a fair point, regarding retryability. The primary intention of this error is to deal with clients that are incorrectly configured to talk to services in regions that they are not available on (as displayed in the message), in those cases retrying would be wasteful. We could probably toggle retryable on for this class of errors without a significant behavioral difference for those users if it means being able to potentially recover from intermittent networking issues.

[...]

PR #529, opened 2015-03-11
set ENOTFOUND_ERROR errors to be retryable

mick:
Retries on dns related errors.  Sometimes it wont be worth retrying in the case of invalid dns names, but it will in the case of intermittent dns / network issues.

fixes: #528


AdityaManohar:
Merging this in. Tests coming up in just a minute.

PR #534, opened 2015-03-15
Do not ignore bower.json itself

joscha:
In order to allow webpack and other build tools that rely on a properly defined `"main"` section in `bower.json` to work with this SDK without using any aliases or other workarounds, the bower.json itself should not be excluded from the installed bower package.


AdityaManohar:
Pulling this in. Thanks for the contribution @joscha!

Issue #412, opened 2014-11-18
Throttling: Rate exceeded error thrown


rclark:
We've written a library that polls CloudFormation to determine a stack's status after a `createStack` or `updateStack` call. We've been encountering throttling errors like this one:

```
/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/request.js:32
          throw err;
                ^
Throttling: Rate exceeded
    at Request.extractError (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/protocol/query.js:39:29)
    at Request.callListeners (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/sequential_executor.js:100:18)
    at Request.emit (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/request.js:604:14)
    at Request.transition (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/request.js:21:12)
    at AcceptorStateMachine.runTo (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /.../cfn-stack-event-stream/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/request.js:22:9)
    at Request.<anonymous> (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/request.js:606:12)
    at Request.callListeners (/.../cfn-stack-event-stream/node_modules/aws-sdk/lib/sequential_executor.js:104:18)
```

It appears that the error is not being handled in aws-sdk and so we can't handle it downstream with a back-off routine. I'm wondering why this error seems to be sneaking through aws-sdk?

Thanks!

patrick-motard:
I'm getting this same error for Cloudwatch `putMetricAlarm`. Is this expected?

```
{ Throttling: Rate exceeded
    at Request.extractError (/home/han/code/aws-utilities/node_modules/aws-sdk/lib/protocol/query.js:47:29)
    at Request.callListeners (/home/han/code/aws-utilities/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/home/han/code/aws-utilities/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
[...]

PR #545, opened 2015-03-25
Fix retry after ThrottlingException

daguej:
This is an additional fix for #412.  The patch by @lsegal (3677cd57de1c89dadb909d2f9029dc498a9e267f) checks for `err.code == 'Throttling'`, but some AWS services return `ThrottlingException`.

### DIFF
--- a/lib/service.js
+++ b/lib/service.js
@@ -341,6 +341,7 @@ AWS.Service = inherit({
     // this logic varies between services
     switch (error.code) {
       case 'ProvisionedThroughputExceededException':
+      case 'Throttling':
       case 'ThrottlingException':
         return true;
       default:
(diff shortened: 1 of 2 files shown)


AdityaManohar:
@daguej 
The error code `Throttling` is correct and is used by some services including AWS CloudFormation. The [CloudFormation API documentation for common errors](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/CommonErrors.html) elaborates on this. I agree that certain services may return a `ThrottlingException` error code and consequently this should be an additive change. Out of curiosity, which service is returning this exception?

daguej:
Tweaked to check for both `Throttling` and `ThrottlingException`.

I saw `ThrottlingException` when calling Lambda's [`deleteFunction`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Lambda.html#deleteFunction-property).

AdityaManohar:
@daguej 
Merging this in. Thanks for contributing!

Issue #158, opened 2013-09-17
S3.putObject RangeError: Maximum call stack size exceeded node v0.10.15


Meekohi:
_This bug does not exist with node v0.10.13, but does on v0.10.15, I have not tested v0.10.18_

When doing putObject with a large file (100MB+), node issues this warning many time:

`(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.`

and then crashes:

`RangeError: Maximum call stack size exceeded`

Best quick-fix solution is to use a stream instead (probably what you wanted anyway):

``` javascript
var stream = fs.createReadStream('filename.zip');
var uploadOptions = {
                Bucket: s3bucket,
                Key: s3key,
                ACL: "public-read",
                ContentType: "application/zip",
                Body: stream
};
s3.putObject(uploadOptions, function(err, data){ ... });
```

PR #551, opened 2015-03-29
Add highwatermark to readable in buffer.tostream

rclark:
I ran into this issue with recursive `nextTick` calls again and thought I'd just throw in a PR that helps: see https://github.com/aws/aws-sdk-js/issues/158#issuecomment-58612464. Again, while this problem does represent a problem in the node.js stream api, adding this is a very simple workaround.

p.s. this time I encountered the error when uploading a ~25MB zip file to lambda.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -120,7 +120,7 @@ var util = {
     toStream: function toStream(buffer) {
       if (!util.Buffer.isBuffer(buffer)) buffer = new util.Buffer(buffer);
 
-      var readable = new (util.nodeRequire('stream').Readable)();
+      var readable = new (util.nodeRequire('stream').Readable)({ highWaterMark: 5 * Math.pow(1024, 2) });
       var pos = 0;
       readable._read = function(size) {
         if (pos >= buffer.length) return readable.push(null);


lsegal:
I'm not sure we can pull this in. This change causes some of our progress event tests to fail because adding a HWM causes Node.js to stop emitting granular progress events, and based on the documentation, it looks like this change would cause Node.js to buffer 5mb chunks into memory, which would potentially incur a significant memory performance regression into the SDK.

Out of curiosity, are you running into this on a small EC2 instance? We've noticed this behavior on t2.micro type instances because of the known issue in 0.10.x that you mentioned in #158. I think the better solution here might be to upgrade to 0.12.x now that it is an officially supported version, which should resolve your nextTick issue.

Let me know what you think.

rclark:
> I think the better solution here might be to upgrade to 0.12.x

This sounds fine to me. We encounter this problem rarely and can usually find a way around it. I've seen it crop up on all kinds of EC2s. I've never run a t2 instance so I'm afraid I can't speak to that.

lsegal:
@rclark in that case I'm going to mark this PR as closed, but thanks for the extra info!

PR #574, opened 2015-04-14
fix for code example in node-examples topic

EronHennessey:
This is a simple (but significant) fix for a code example in the node-examples topic.


lsegal:
Thanks!

PR #584, opened 2015-04-17
Detach SHA_256 handler for AWS.Glacier

AdityaManohar:
CC: @lsegal 

The `AWS.Glacier` client was adding the `X-Amz-Content-Sha256` header twice.
Unit tests pass
Glacier integ tests pass

### DIFF
--- a/lib/services/glacier.js
+++ b/lib/services/glacier.js
@@ -10,7 +10,8 @@ AWS.util.update(AWS.Glacier.prototype, {
     } else {
       request.on('validate', this.validateAccountId);
     }
-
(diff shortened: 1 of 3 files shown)


AdityaManohar:
I've added a couple more tests to ensure that the checksum is computed only once. These fail if the casing of the `X-Amz-Content-Sha256` is not honored. 

The unit test case specifically for the Glacier client is somewhat difficult to write given that `crypto.sha256` is invoked in a loop which depends on the size of the payload.

Let me know if you have other ideas. And thanks much for your time in reviewing this!

Issue #554, opened 2015-04-01
Performance Regression in aws-sdk


sboora:
We recently upgraded the SDK from 1.15.0 to 2.1.9 and noticed a performance regression in 2.1.9. 

Here are the benchmark numbers from PUT and GET operations on a DynanmoDB test table in us-east-1.

AWS SDK Version 1.15.0 
Sample Size 10000
Record Size (byte) 2048
Waiting for the table to be ready in Dynamo
[putItem] lantecy (ms): mean/std_dev/median 147.27/52.28/139, throughput (/sec): 142.23, errors: 0
[getItem] lantecy (ms): mean/std_dev/median 133.89/49.19/130, throughput (/sec): 146.39, errors: 0

AWS SDK Version 2.1.9
Sample Size 10000
Record Size (byte) 2048
Waiting for the table to be ready in Dynamo
[putItem] lantecy (ms): mean/std_dev/median 164.55/49.28/157.5, throughput (/sec): 136.12, errors: 0
[getItem] lantecy (ms): mean/std_dev/median 150.67/45.28/146.5, throughput (/sec): 140.39, errors: 0

Can you let us know if this is already a known and an acceptable issue?

AdityaManohar:
@sboora 
Thanks for reporting! Would you be able to share details of your sampling method? Are you factoring out network latency in your measurement?

lsegal:
Note that aws-sdk@2.x does more than 1.x did. Specifically, we began to convert types for blobs and timestamps on output responses, which does affect performance. You can toggle this off if you don't need the behavior via the [`AWS.Config.convertResponseTypes`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#convertResponseTypes-property) property, which should improve performance on GetItem calls.

Also what version of Node.js are you using?

sboora:
@AdityaManohar both these benchmarks are done on the same ec2 instance and to the same DynamoDB table, so not sure if factoring out network latency helps.  (this is assuming there were no issues in the network since I ran this benchmark couple of times and the results were consistent)

@lsegal, we are

lsegal:
@sboora let us know once you have updated numbers. I'm only seeing a ~4% difference in both cases which could be explained by the SDK adding more robustness / functionality to its feature set over this major version, though it could also just be explained by networking and I/O bottlenecks. Without seeing the benchmark code it would be difficult to know what to attribute this to.

sboora:
[...]

PR #592, opened 2015-04-27
Add a new disableProgressEvents for better read performance.

chaosgame:
Fixes https://github.com/aws/aws-sdk-js/issues/554

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -80,6 +80,8 @@ require('./credentials/credential_provider_chain');
  *     * **xhrWithCredentials** [Boolean] &mdash; Sets the "withCredentials"
  *       property of an XMLHttpRequest object. Used in the browser environment
  *       only. Defaults to false.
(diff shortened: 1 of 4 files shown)


lsegal:
Thank you for taking the time to implement this change, @chaosgame. Unfortunately I'm not sure this is the right PR to take for a few reasons:

1. Turns out we can detect whether progress events have been subscribed to prior to sending the request, which means we can intelligently disable these events without an explicit switch. Avoiding explicit switches like these is typically something we optimize for. In this case we were able to.

2. That said, all the testing I've done indicates that disabling progress events would not have any impact on the performance of the requests. If there is a performance regression from 1.x (as per #554), it is likely not coming from these emitted events. 

For these reasons, pulling this in would add complexity but not actually resolve the root issue-- at least from my testing. If it really does help on your end, providing that data would be helpful in #554, but also, re-implementing this to intelligently disable the events would be a more appropriate solution (just checking if there are no event listeners and disabling it that way). Thanks!

PR #599, opened 2015-05-08
Link with markdown

adrianblynch:
Without the MD link, npm doesn't make the URL a link

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ bower install aws-sdk-js
 
 You can find a getting started guide at:
 
-http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/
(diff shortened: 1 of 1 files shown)


lsegal:
Would it be reasonable to open an issue with npm to auto-link http URLs? This is a pretty common behavior in most markdown parsers.

lsegal:
Thanks for opening this PR. Even though this change is simple, I'd rather see this behavior changed in npm since it is very common for markdown parsers to support a feature called "autolink" which should turn bare URLs into links. Although we only use one right now, we may add more bare URLs in the future (forgetting about the limitation on npm), and all of our other tools support this. npmjs.org should too.

PR #601, opened 2015-05-08
AWS.S3.ManagedUpload fixes for zero-byte streams and buffering performance

### LINKED ISSUE #549: Can't upload file to s3 when file size is 0
When file size is 0,i can't upload file to s3,Use wireshark tools i found sdk not send http request to s3.
my code:
s3obj.upload({Body: body}).send(cb);

i modify two code in l lib/s3/managed_upload.js file and it't work:

  /**
   * @api private
   */
  fillStream: function fillStream() {
    var self = this;
    if (self.activeParts >= self.queueSize) return;

    var buf = self.body.read(self.partSize - self.partBuffer.length) ||
              self.body.read();
    if (buf) {
      self.partBuffer = AWS.util.Buffer.concat([self.partBuffer, buf]);
      self.totalChunkedBytes += buf.length;
    }

    if (self.partBuffer.length >= self.partSize) {
      self.nextChunk(self.partBuffer.slice(0, self.partSize));
      self.partBuffer = self.partBuffer.slice(self.partSize);
    } else if (self.isDoneChunking) {
      self.totalBytes = self.totalChunkedBytes;
      if (self.partBuffer.length >= 0) {    // change self.partBuffer.length > 0 to self.partBuffer.length >= 0
        self.numParts++;
        self.nextChunk(self.partBuffer);
      }
      self.partBuffer = new AWS.util.Buffer(0);
    }

    self.body.read(0);
  }

  /**
   * @api private
   */
  adjustTotalBytes: function adjustTotalBytes() {
    var self = this;
    try { // try to get totalBytes
      self.totalBytes = byteLength(self.body);
    } catch (e) { }

    // try to adjust partSize if we know payload length
    if (self.totalBytes) {
      var newPartSize = Math.ceil(self.totalBytes / self.maxTotalParts);
      if (newPartSize > self.partSize) self.partSize = newPartSize;
    }else if(self.totalBytes ==0){     //When self.totalBytes is 0, it's not boolean false; so care  of here
    	self.totalBytes =0;
    } 
    else {
      self.totalBytes = undefined;
    }
  }

#===========#
Please update to github source files

### LINKED ISSUE #594: ManagedUpload callback is never invoked if body is 0 bytes, nor an error thrown.
If the input body is size 0, the callback is never called, because nextChunk() is never called. I would expect the callback to be called with an err value. Otherwise the upload is kicked off and no other events are triggered, nor Error's thrown, it just goes dark.

### LINKED ISSUE #597: CPU 100% when using ManagedUpload with many small writes
When using a stream that sends very small amounts of data with ManagedUpload the cpu will go to 100% while buffering. The culprit is

`self.partBuffer = AWS.util.Buffer.concat([self.partBuffer, buf]);`

Imagine if each write is only a 100 bytes, there are too many concat operations. Much more efficient to add "buf" to an array of buffers and then do one final concat when enough data has been queued up.

As a work around I had to build a buffering stream that does exactly that, and it fixed the problem.

lsegal:
Fixes #549 
Fixes #594 
Fixes #597

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -158,7 +158,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
       if (self.body instanceof Stream) {
         runFill = false;
         self.fillQueue = self.fillStream;
-        self.partBuffer = new AWS.util.Buffer(0);
+        self.partBuffers = [];
         self.body.
           on('readable', function() { self.fillQueue(); }).
           on('end', function() {
@@ -315,7 +315,12 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
   /**
    * @api private
    */
-  partBuffer: null,
+  partBuffers: null,
+
+  /**
+   * @api private
+   */
+  partBufferLength: 0,
 
   /**
    * @api private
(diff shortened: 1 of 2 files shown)


dconnolly:
Awesome, thank you! :rocket:

PR #607, opened 2015-05-14
Add notes to the contributing guidelines about running the tests locally.

wearhere:
There are none currently, and `npm test` does not work out of the box without installing phantomjs.

[Note: the original version of this request added phantomjs as a dev dependency in `package.json`, but @lsegal indicates below that that is undesirable.]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,30 @@ Please be aware of the following notes prior to opening a pull request:
    yet available in the SDK, please talk to us beforehand to avoid any
    duplication of effort.
 
+### Testing
+
+To run the tests locally, install `phantomjs`. You can do so using [Homebrew][homebrew]:
+
+```
+brew install phantomjs
+```
+
+Then, to run all tests:
+
+```
+npm test
+```
+
+To run a particular test subset e.g. just the unit tests:
+
+```
+npm run-script unit
+```
+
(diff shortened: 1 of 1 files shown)


lsegal:
Unfortunately we don't want to lock to the npm version of phantomjs as this disables us from testing with phantomjs 2.x, which is not currently in npm. It also causes problems when phantom is available through brew. The best way to bring in non-JS-code deps is to build them through system package managers like Homebrew, yum, apt, etc.

wearhere:
No trouble. If you don't want the project to bundle `phantomjs` it'd be nice to offer some other way to get the tests running, though—might we add something to the README? What's your suggested method of installation & what versions do you test with?

wearhere:
Speaking more broadly, I was surprised that [the contributing guidelines](https://github.com/aws/aws-sdk-js/blob/master/CONTRIBUTING.md) talk about adding code—but not about testing it. I had to do a little bit of digging through `package.json`, the test folder structure, etc. to figure out how to get things running and where I might add tests e.g. for #608.

lsegal:
@wearhere We would be open to additions to the contributing guidelines and/or README.

wearhere:
Ok @lsegal, I've fixed up the request to modify the contributing guidelines instead.

lsegal:
Looks good!

PR #608, opened 2015-05-14
Clear `AWS.CognitoIdentityCredentials` instance information alongside the cached ID.

wearhere:
Without this change, an instance will keep using the cached information even after the
cache is cleared.


lsegal:
This looks like the right thing to do, thanks for the pull request!

wearhere:
You're welcome, thanks for the speedy response!

PR #612, opened 2015-05-19
Clean up createReadStream logic to use a PassThrough stream.

### LINKED ISSUE #605: Out of memory when downloading a large file from S3
Hi there! I am using node.js (v0.12.1) and the aws-sdk (latest version, 2.1.27) to download a large S3 file (~4.8GB). I am using the following code, taken from the documentation (http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-examples.html):

var AWS    = require ('aws-sdk');
var s3 = new AWS.S3 ();
var params = {Bucket: 'BUCKET NAME', Key: 'FILE NAME'};
var output = fs.createWriteStream ('output.tar.gz');
s3.getObject(params).createReadStream().pipe(output);

When I start the script, CPU and memory usage skyrockets, to the point that other processes die (ie: redis) and finally the process is killed by the OS (this in an ubuntu instance with 2 CPUs and 2GB of memory, doing next no nothing in the background). This happens well before downloading 2GBs of the file.

We need to download large S3 files for performing backup restores. Even with a beefier machine (which should be unnecessary considering the performance of node streams), we would be worried about the download taking up all available memory and disrupting other processes (for example, the script that generates and uploads those same backups).

Please let me know if I can contribute further information to replicate the bug.

Thanks!

lsegal:
Also fixes incorrect detection of streaming API in Node 0.12.x and
resolves a memory leak when not using pipe.

Fixes #605 

Ran against new S3 integration tests and all updated tests pass with these changes.

### DIFF
--- a/features/s3/step_definitions/objects.js
+++ b/features/s3/step_definitions/objects.js
@@ -69,9 +69,12 @@ module.exports = function () {
     var params = {Bucket: this.sharedBucket, Key: key};
     var world = this;
     this.result = '';
-    this.service.getObject(params).createReadStream().
-      on('end', function() { callback(); }).
-      on('data', function(d) { world.result += d.toString(); });
+    var s = this.service.getObject(params).createReadStream();
+
+    setTimeout(function() {
+      s.on('end', function() { callback(); });
+      s.on('data', function(d) { world.result += d.toString(); });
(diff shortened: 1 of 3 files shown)


lsegal:
CC @AdityaManohar

PR #634, opened 2015-06-23
Support multiple send's with leavePartsOnError

### LINKED ISSUE #633: Resume multipart upload with 'leavePartsOnError'?
As far as I can tell from the current implementation of `ManagedUpload`, the user has to do a lot of work/bookkeeping to resume an upload that partially completed and then left parts behind. Trying to start a `ManagedUpload` from where it left off will not kick off the chunks that failed prior to the current `partPos`, and the 'complete' request will be incomplete or malformed. Is there any way to resume a multipart upload after stopping it while leaving parts behind with the current sdk api, or will you accept PR's trying to make this possible?

dconnolly:
Track which parts have been uploaded and if we have an `ETag` for them, don't try to upload them again. This works well with `leavePartsOnError`, so that you can restart a failed upload and only upload the missing parts.

This passes existing sdk tests and works well with buffer/file-type uploads. I don't think this is a use case for uploading streams, but it doesn't conflict with that case.

Fixes #633

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -148,6 +148,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
    */
   send: function(callback) {
     var self = this;
+    self.failed = false;
     self.callback = callback || function(err) { if (err) throw err; };
 
     var runFill = true;
@@ -411,6 +412,10 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
       return null;
     }
 
+    if (self.completeInfo[partNumber] && self.completeInfo[partNumber].ETag !== null) {
+      return null; // Already uploaded this part.
+    }
+
     self.activeParts++;
     if (!self.service.config.params.UploadId) {
 
@@ -438,6 +443,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
    */
   uploadPart: function uploadPart(chunk, partNumber) {
     var self = this;
+
     var partParams = {
       Body: chunk,
       ContentLength: AWS.util.string.byteLength(chunk),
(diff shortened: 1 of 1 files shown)


dconnolly:
@lsegal thoughts?

dconnolly:
@AdityaManohar @lsegal bump? We've been using this in production for almost a month with success.

lib/s3/managed_upload.js
```diff
@@ -522,7 +532,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
    */
   finishMultiPart: function finishMultiPart() {
     var self = this;
-    var completeParams = { MultipartUpload: { Parts: self.completeInfo } };
+    var completeParams = { MultipartUpload: { Parts: self.completeInfo.slice(1) } };
```
AdityaManohar: @dconnolly Is there a reason that we are slicing out the first index here?

AdityaManohar:
@dconnolly I will look into adding some tests for this and then merge it in if everything looks good.

lib/s3/managed_upload.js
AdityaManohar: CC @dconnolly

dconnolly: We are slicing as we are following the pattern of keeping track of parts by the correctly indexed key, ie: part 1 has key value 1, and key 0 is null, so we slice it out.

dconnolly:
That's great, thank you.

lonormaly:
+1(!!) Let's support pause / resume

lonormaly:
@dconnolly I've been using your version for some time now, works great!
Question: is there a way to abort an upload and to force clean the uploaded parts even when using the leavePartsOnError?

PR #636, opened 2015-06-24
Reset `request.httpRequest` on request restart.

### LINKED ISSUE #585: Retried S3 request uses a bogus hostname (the bucket name is repeated)
I have a node service using the aws-sdk that gets thousands of files from S3 per day.  It is using node v0.10.26 and aws-sdk 2.0.1 (although aws-sdk 2.1.24 still appears to have this bug based on a code inspection).

The vast majority of the time, s3.getObject() succeeds.  However, very infrequently -- approximately once every month -- I see a call to s3.getObject fail with a certificate error like this one:
```
 { [NetworkingError: Hostname/IP doesn't match certificate's altnames]
     message: 'Hostname/IP doesn\'t match certificate\'s altnames',
     code: 'NetworkingError',
     region: 'us-east-1',
     hostname: 'hurley-config-us-west-2.hurley-config-us-west-2.s3-us-west-2.amazonaws.com',
     retryable: true,
     time: Fri Apr 17 2015 17:39:38 GMT+0000 (UTC),
     statusCode: 200 } }
```
What really got my attention was the hostname above.  The name of the bucket is repeated in the hostname.  The bogus hostname resolves in DNS, but the certificate used by the HTTPS/SSL server has a common name of "*.s3-us-west-2.amazonaws.com" which does not match the bogus hostname being used.  This explains the cert error.

But, why is this bogus hostname being used?  The service code passes in the region, bucket name, and key to the aws-sdk this way:
```
var AWS = require('aws-sdk');
var s3 = new AWS.S3({region: 'us-west-2'});
var params = {
  Bucket: 'hurley-config-us-west-2',
  Key: various valid paths
};
s3.getObject(params, function(err, data) {
  ...
```
...so the hostname is computed by the aws-sdk code.

In aws-sdk/lib/services/s3.js, I see this code in populateURI:
```
  populateURI: function populateURI(req) {
    var httpRequest = req.httpRequest;
    var b = req.params.Bucket;

    if (b) {
      if (!req.service.pathStyleBucketName(b)) {
        if (!req.service.config.s3BucketEndpoint) {
          httpRequest.endpoint.hostname =
            b + '.' + httpRequest.endpoint.hostname;

          var port = httpRequest.endpoint.port;
          if (port !== 80 && port !== 443) {
            httpRequest.endpoint.host = httpRequest.endpoint.hostname + ':' +
              httpRequest.endpoint.port;
          } else {
            httpRequest.endpoint.host = httpRequest.endpoint.hostname;
          }
        }
      ...
```
The first time populateURI is run for my request, it would set req.httpRequest.endpoint.hostname to:
```
hurley-config-us-west-2.s3-us-west-2.amazonaws.com
```
This is correct.
However, if populateURI were to be called a second time for the same request, it would now reset req.httpRequest.endpoint.hostname to:
```
hurley-config-us-west-2.hurley-config-us-west-2.s3-us-west-2.amazonaws.com
```
This is incorrect -- the name of the bucket is repeated in the hostname.

Further examining the code, it appears that the request's state machine allows it to go through the 'build' state more than once in some failure/retry scenarios.  This would cause populateURI to be called more than once, which then uses this bogus hostname and causes the certificate validation error.

The aws-sdk code should not do this.  The unwanted effects are:
- The retry logic for s3 in the aws-sdk will always fail for this scenario.
- Whatever the original error was is masked by the bogus certificate error.

Instead, when retrying s3 requests, the aws-sdk should use the same hostname it used the first time, for example:

bucketName.s3-region.amazonaws.com
(NOT bucketName.bucketName.s3-region.amazonaws.com).

AdityaManohar:
Fixes #585

### DIFF
--- a/lib/event_listeners.js
+++ b/lib/event_listeners.js
@@ -122,6 +122,11 @@ AWS.EventListeners = {
       var err = this.response.error;
       if (!err || !err.retryable) return;
 
+      this.httpRequest = new AWS.HttpRequest(
+        this.service.endpoint,
+        this.service.region
+      );
+
       if (this.response.retryCount < this.service.config.maxRetries) {
         this.response.retryCount++;
       } else {
--- a/test/event_listeners.spec.coffee
+++ b/test/event_listeners.spec.coffee
@@ -150,6 +150,21 @@ describe 'AWS.EventListeners', ->
           sendRequest file, (err) ->
             done()
 
+  describe 'restart', ->
+    request = null
+
+    it 'constructs a fresh httpRequest object', ->
+      request = makeRequest()
+      httpRequest = request.httpRequest
+      request.on 'build', ->
+        if !@threwSimulatedError
+          @threwSimulatedError = true
+          err = new Error('simulated error')
+          err.retryable = true
+          throw err
+      request.build()
+      expect(request.httpRequest).not.to.eql(httpRequest)
+
   describe 'sign', ->
     it 'takes the request object as a parameter', ->
       helpers.mockHttpResponse 200, {}, ''


AdityaManohar:
All integ tests pass
All browser integ tests pass

CC @lsegal

PR #655, opened 2015-07-09
Upgrade to 0.4.9 of xml2js in order to properly resolve xml entities.

stigerland:
The previously used version (0.2.8) handles xml entities improperly.

This specifically affected the simpledb apis for retrieving attributes or selecting several items for a domain, but it might also affect other aws apis that use xml as the wire protocol.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     },
     "dependencies": {
       "sax": "0.5.3",
-      "xml2js": "0.2.8",
+      "xml2js": "0.4.9",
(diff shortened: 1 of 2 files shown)


bruun:
Would it be possible to get this PR looked at? Right now we are depending on a fork of this project in order for our code to work correctly. Thanks!

chrisradek:
@stigerland and @bruun, I'll take a look at this PR this week. Especially since this is such a big jump in the dependency version, we want to be very sure that nothing else breaks if we update.

PR #657, opened 2015-07-09
Support null values in JSON response list entries.

lsegal:


lsegal:
cc @AdityaManohar

lsegal:
Fixed up some stuff, thanks for review @AdityaManohar!

PR #667, opened 2015-07-22
Generate shared examples

DanielNi:
Adds functionality and tests for generating shared documentation examples from the language-agnostic JSON format. These new examples will live alongside the JSON apis.

The code in this PR loads the examples from the plugin.rb, which passes them into the model_documentor.rb. From there the SharedExampleVisitor class parses the JSON and creates a JS example immediately after the automatically generated examples in the documentation.

### DIFF
--- a/doc-src/templates/api-versions/model_documentor.rb
+++ b/doc-src/templates/api-versions/model_documentor.rb
@@ -80,7 +80,7 @@ class MethodDocumentor
 
   attr_reader :lines
 
-  def initialize(operation_name, operation, api, klass)
+  def initialize(operation_name, operation, api, klass, examples)
     desc = documentation(operation)
     desc ||= "Calls the #{method_name(operation_name, false)} API operation."
     desc = desc.gsub(/^\s+/, '').strip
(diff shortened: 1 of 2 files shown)


doc-src/templates/api-versions/model_documentor.rb
```diff
@@ -93,11 +93,19 @@ def initialize(operation_name, operation, api, klass)
     @lines += shapes(api, operation['input']).map {|line| "  " + line }
 
     ## @example tag
-
     @lines << "@example Calling the #{method_name(operation_name)} operation"
```
lsegal: I would suggest moving any hand-written examples above the generated example, since they are likely to be more correct / useful.

doc-src/templates/api-versions/model_documentor.rb
```diff
@@ -93,11 +93,19 @@ def initialize(operation_name, operation, api, klass)
...
     @lines << generate_example(api, klass, method_name(operation_name),
                 operation['input']).split("\n").map {|line| "  " + line }
-    @lines << ""
+    if examples
+      examples.each do |example|
+        @lines << "@example Example: #{example['title']}"
```
lsegal: Example tags should not be prefixed with "Example:" text, let's just list the title.

chrisradek:
Closed by #860

PR #668, opened 2015-07-24
Android Users have localStorage as null

tonyho1992:
On android phones, we have seen that there are times when `window.localStorage` is null. Since it's an object, we're actually using `null` as our storage. This should not be the case.


AdityaManohar:
@tonyho1992 Thanks for your contribution! I'm happy to merge this in.

PR #679, opened 2015-08-11
add more details to 'httpUploadProgress' event emitter to support tra…

mohamed-kamal:
when trying to upload multiple files in parallel, you get progress by listening on 'httpUploadProgess' event. but the payload you get is like:

**{loaded: 147456, total: 13566183, part: 1}**
**{loaded: 114688, total: 415560925, part: 1}**

so you don't know which object is for which file.

we added file name and key to the event emitter payload to be like this:
**{loaded: 147456, total: 415560925, part: 1, name: "12.mp4", key: "idOn4M86"}**
**{loaded: 180224, total: 13566183, part: 1, name: "122345.mp4", key: "ShMOAxJN"}**

so you can easily track the progress of each file.

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -555,7 +555,9 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
       info = {
         loaded: upload.totalUploadedBytes,
         total: upload.totalBytes,
-        part: this.params.PartNumber
+        part: this.params.PartNumber,
+        key: this.params.key,
+        name: upload.body.name
       };
     }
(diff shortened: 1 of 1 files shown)


chrisradek:
@mohamed-kamal 
I like adding the part number and the key, but do you have some example code for how you're getting the name to be set? What you have as the name, I'd expect to be the key. I'm curious how you're setting the name because based on the commit, it looks like it's based on a `name` property on the `body` you pass in to be uploaded.

chrisradek:
Closed by #788

PR #697, opened 2015-08-31
Retry S3 uploads with different region when using signature V4.

rolftimmermans:
When uploading objects to S3 (or making other requests), it is possible to use the `us-east-1` region (the default) so that a region does not have to be specified. However, this fails with signature version v4 (which is required for some regions), because the region is included in the `Authorization` header.

The returned error is `AuthorizationHeaderMalformed`. It actually contains the region the bucket is located in.

This patch uses the `<Region>` element in the returned error to update the region to the suggested region in order to be able to retry the request. This ensures that it is always possible to omit the region when using S3 in combination with signature v4.

### DIFF
--- a/lib/protocol/rest_xml.js
+++ b/lib/protocol/rest_xml.js
@@ -43,7 +43,8 @@ function extractError(resp) {
   if (data.Code) {
     resp.error = util.error(new Error(), {
       code: data.Code,
-      message: data.Message
+      message: data.Message,
+      region: data.Region
     });
   } else {
     resp.error = util.error(new Error(), {
--- a/lib/protocol/rest_xml.js
+++ b/lib/protocol/rest_xml.js
@@ -43,8 +43,7 @@ function extractError(resp) {
   if (data.Code) {
     resp.error = util.error(new Error(), {
       code: data.Code,
-      message: data.Message,
-      region: data.Region
+      message: data.Message
     });
   } else {
     resp.error = util.error(new Error(), {
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -272,7 +272,7 @@ AWS.util.update(AWS.S3.prototype, {
       return true;
     } else if (error && error.code === 'RequestTimeout') {
       return true;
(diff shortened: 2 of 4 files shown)


lib/protocol/rest_xml.js
```diff
@@ -43,7 +43,8 @@ function extractError(resp) {
   if (data.Code) {
     resp.error = util.error(new Error(), {
       code: data.Code,
-      message: data.Message
+      message: data.Message,
+      region: data.Region
```
chrisradek: There shouldn't be any changes needed in this file since the `extractError` in lib/services/s3.js will handle setting the region.  This functionality is specific to S3 so would like to keep the logic contained there.

lib/services/s3.js
```diff
@@ -272,6 +272,10 @@ AWS.util.update(AWS.S3.prototype, {
       return true;
     } else if (error && error.code === 'RequestTimeout') {
       return true;
+    } else if (error.code === 'AuthorizationHeaderMalformed' &&
```
chrisradek: Just to be safe, can you also add a check that error exists, similar to the conditional before this one?

test/protocol/rest_xml.spec.coffee
```diff
@@ -214,6 +214,7 @@ describe 'AWS.Protocol.RestXml', ->
         <Error>
           <Code>InvalidArgument</Code>
           <Message>Provided param is bad</Message>
+          <Region>eu-west-1</Region>
```
chrisradek: If you remove the changes from the rest_xml.js file, you won't need these tests.

chrisradek:
@rolftimmermans 
Thank you very much for your pull request!  I've added a few comments, but overall your changes look good. Once the requested changes are made I'll merge this in.
Thanks!

rolftimmermans:
Thanks! I've addressed your comments.

chrisradek:
@rolftimmermans 
Thank you! #697 is now merged and will be available through NPM in the next release. We really appreciate your contribution!

PR #707, opened 2015-09-10
Disable signature cache

jippeholwerda:
We use the AWS SDK both for our own purposes and for requests of users. In the former case, it is good that the v4 signatures are cached. In the latter, however, the Access Key Id is mostly different between subsequent requests, making caching unnecessary.

What's even more important, when users make a request with an invalid secret access key, the signature is cached as well.

Therefore, this pull requests allows one to disable caching of the signature for a specific configuration.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -390,7 +390,8 @@ AWS.Config = AWS.util.inherit({
     convertResponseTypes: true,
     dynamoDbCrc32: true,
     systemClockOffset: 0,
-    signatureVersion: null
+    signatureVersion: null,
+    signatureCache: true
   },
 
   /**
--- a/lib/config.js
+++ b/lib/config.js
@@ -95,6 +95,11 @@ require('./credentials/credential_provider_chain');
  * @!attribute signatureVersion
  *   @return [String] the signature version to sign requests with (overriding
  *     the API configuration). Possible values are: 'v2', 'v3', 'v4'.
+ *
+ * @!attribute signatureCache
+ *   @return [Boolean] whether the signature to sign requests with (overriding
+ *     the API configuration) is cached. Only applies to the signature version 'v4'.
+ *     Defaults to `true`.
  */
 AWS.Config = AWS.util.inherit({
   /**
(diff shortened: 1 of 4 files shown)


lib/config.js
```diff
@@ -390,7 +390,8 @@ AWS.Config = AWS.util.inherit({
     convertResponseTypes: true,
```
chrisradek: Would you mind adding some docs around this new config parameter?
Something like this at line 98:
```
 * @!attribute signatureCache
 *   @return [Boolean] whether the signature to sign requests with (overriding
 *     the API configuration) is cached. Only applies to the signature version 'v4'.
 *     Defaults to `true`.
```
and something like this at line 185:
``` 
   * @option options signatureCache [Boolean] whether the signature to sign
   *   requests with (overriding the API configuration) is cached. Only applies
   *   to the signature version 'v4'. Defaults to `true`.
```

lib/signers/v4.js
```diff
@@ -87,17 +88,27 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
   },
 
   signature: function signature(credentials, datetime) {
-    var cache = cachedSecret[this.serviceName];
+    var cache = null
```
chrisradek: Very minor, missing semicolon.

lib/config.js
```diff
@@ -390,7 +390,8 @@ AWS.Config = AWS.util.inherit({
...
     dynamoDbCrc32: true,
     systemClockOffset: 0,
-    signatureVersion: null
+    signatureVersion: null,
+    signatureCache: true
```
chrisradek: Would you mind adding some docs around this new config parameter?
Something like this at line 98:
```
 * @!attribute signatureCache
 *   @return [Boolean] whether the signature to sign requests with (overriding
 *     the API configuration) is cached. Only applies to the signature version 'v4'.
 *     Defaults to `true`.
```
and something like this at line 185:
``` 
   * @option options signatureCache [Boolean] whether the signature to sign
   *   requests with (overriding the API configuration) is cached. Only applies
   *   to the signature version 'v4'. Defaults to `true`.
```

chrisradek:
Thank you again @jippeholwerda for the pull request, it is very much appreciated!

PR #718, opened 2015-09-16
Fix/crypto browser unit test

chrisradek:
@AdityaManohar @jeskew


jeskew:
Any reason the middle check isn't verifying that data is an array (as per the comment)?

chrisradek:
I'll need to update the comment.  I originally added a check ArrayBuffer.isView(data), but that isn't supported in nodejs/phantomjs.  Comparing data's prototype to a typed array's prototype leads to false positives in some browsers as well.  I can add a check that the data's buffer field is an instance of ArrayBuffer for some added safety.

jeskew:
:shipit:

PR #731, opened 2015-09-23
fix issue when output is configured to return list

DrMegavolt:
configure output as list 
example
```json
"Search": {
      "http":{
        "method":"POST",
        "requestUri":"/account/search"
      },
      "name":"Search",
      "input":{"shape":"SearchAccountRequest"},
      "output":{"shape":"AccountList"}
    },

"AccountList":{
      "type":"list",
      "member":{"shape":"Account"}
    }
```

use protocol rest-json
when it is returned in method extractData line 57 rep.data is {}

after parsing rep.data will be array (returned as body from server)
but  util.merge(data, resp.data); will merge ({}, [a,b])
as result it will be { '0': a, '1':b}
instead of array

### DIFF
--- a/lib/protocol/rest_json.js
+++ b/lib/protocol/rest_json.js
@@ -56,7 +56,9 @@ function extractData(resp) {
   } else {
     var data = resp.data;
     Json.extractData(resp);
-    resp.data = util.merge(data, resp.data);
+    resp.data = Array.isArray(resp.data) ?
+      resp.data :
+      util.merge(data, resp.data);
   }
 }
 


chrisradek:
Thanks @DrMegavolt for the pull request. Do you know of any rest-json operations that have a list as an output? When I did a quick scan through our APIs, I only saw outputs defined as structures, which is what we'd expect.

DrMegavolt:
@chrisradek default SDK has only structures. 
I have a fork that does list output and faced with the issue.
If by design list is not supported and only structures can be used as root element, then this PR is not needed

chrisradek:
@DrMegavolt 
I just merged in #766 that should hopefully address what you're trying to accomplish.

We strongly prefer structures as a root element, and in fact don't have any rest-json outputs that are a list. A big reason for this is future-compatibility. If we had an operation that returned a list, if we ever wanted to return something else as well, we would likely have to create a new operation, or break the current one for existing users. By using structures, we just need to add another field to the output.

That said, with the change made in #766, you'll be able to accept a response where the body is a list, but still return an output that's a structure. This is done by including the `payload` field in an operation's `output`.  

For example, imagine a response where the body was simply a list of strings: `['a', 'b', 'c']`.
We may have an operation configured something like this:
```javascript
"Search": {
    "http": {
        "method": "POST",
        "requestUri": "/account/search"
    },
    "input":{"shape":"SearchAccountRequest"},
    "output":{
        "type": "structure",
        "payload": "AccountList", //treat the response's body as this shape
        "members": {
             "shape":"AccountList"
        }
    }
}
//...
"AccountList": {
    "type": "list",
    "member": {
        "shape": "AccountName"
    }
},
"AccountName": {
    "type": "string"
}
```

The result of calling our imaginary `Search` operation with the list above would be:
```
{
    "AccountList": ["a", "b", "c"]
}
```

I hope this change helps you accomplish what you were trying to do above. Let me know if this solve your use case or you have any questions about the change.

chrisradek:
@DrMegavolt 
I'm going to close this pull request due to pr #766.  Feel free to comment if you'd like to discuss your request, or the details of #766 further.

PR #738, opened 2015-09-28
Added ability to specify a LoginId field when creating new CognitoIdentity

chrisradek:
@AdityaManohar @jeskew

### DIFF
--- a/test/credentials.spec.coffee
+++ b/test/credentials.spec.coffee
@@ -600,6 +600,14 @@ describe 'AWS.CognitoIdentityCredentials', ->
       setupCreds(Logins: {provider1: 'Token1', provider3: 'Token3'})
       expect(creds.params.IdentityId).to.equal('MYID')
 
+    it 'uses IdentityId if it is cached with the same LoginId', ->
+      setupCreds({LoginId: 'LOGINID1'})
+      creds.setStorage('id', 'MYID1')
+      creds.loadCachedId()
+      expect(creds.params.IdentityId).to.equal('MYID1')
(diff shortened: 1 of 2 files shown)


test/credentials.spec.coffee
```diff
@@ -832,3 +840,26 @@ describe 'AWS.CognitoIdentityCredentials', ->
...
+              SecretAccessKey: 'SECRET'
+              SessionToken: 'TOKEN'
+          cb null
+        creds.refresh(->)
+        creds.getId (err, id) ->
+          expect(id).to.equal('IDENTITY-ID2')
+          expect(creds.getStorage('id')).to.equal('IDENTITY-ID2')
+        setupCreds({LoginId: 'LOGINIDB'})
+        expect(creds.getStorage('id')).not.to.equal('IDENTITY-ID2')
+        setupCreds({LoginId: 'LOGINIDA'})
```
jeskew: Should there be another `expect` check after this line?

chrisradek: These last 2 steps are actually a clean-up step.  When testing in a browser environment, these credentials would be cached, so the tests would pass the first time it was run in a browser, but not subsequent times. I could explicitly delete the credentials form localstorage instead, so it's more clear why I'm doing this.

jeskew: OIC; I thought that's what the next line was doing.

chrisradek:
I added a minor simplification so it is hopefully clearer which LoginId is getting deleted when the unit test completes.

jeskew:
:shipit:

PR #748, opened 2015-10-08
Lowercase the key when setting mapped httpRequest headers

### LINKED ISSUE #747: s3.getSignedUrl does not normalize Metadata keys properly
When `s3.getSignedUrl` method is given a `Metadata` object in its `params` argument, the keys of the metadata object are (correctly) lowercased when canonicalizing the amazon metadata headers (`x-amz-meta-propertyname`) to generate the StringToSign (resulting in a correct signature), but when generating the query parameter string to append to the signed URL the property names are (incorrectly) not lowercased, resulting in a URL that will be rejected by S3.

Example:
```javascript
s3.getSignedUrl('putObject', {Bucket:'somebucket', Key:'somekey', Metadata: {someProp: 'someVal'});
```
This will (correctly) canonicalize the amazon headers into `x-amz-meta-someprop:someVal` when calculating the StringToSign, but the resulting URL will have the query parameter `x-amz-meta-someProp=someVal` (the "P" is uppercase) instead of the correct `x-amz-meta-someprop=someVal` ("someprop" is all lowercase) appended to it.

The [REST documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html) explains that the meta header should be fully lowercased:

> Notice how the 'x-amz-' headers are sorted, trimmed of whitespace, and converted to lowercase. Note also that multiple headers with the same name have been joined using commas to separate values.

In the [User-Defined Metadata](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#UserMetadata) section of the S3 docs it also explains:

> Amazon S3 stores user-defined metadata in lowercase.

The [documentation for `s3.putObject` params](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#putObject-property) indicates that it should be ok to include Metadata with a non-lowercased property name:

class aws s3 aws sdk for javascript 2015-10-08 10-56-19

bantic:
This ensures that a metadata property like "someVal" will correctly
get mapped to an all-lowercased header "x-amz-meta-someval".

Fixes #747 

This will lowercase all mapped header keys which may not be ideal. I'm not sure what other important cases there may be, but none of the tests failed after making this change. If there is a way to explicitly target lowercasing only the Metadata keys that may be a better fix. Happy to change this PR if that is warranted.

### DIFF
--- a/lib/protocol/rest.js
+++ b/lib/protocol/rest.js
@@ -68,7 +68,7 @@ function populateHeaders(req) {
 
     if (member.location === 'headers' && member.type === 'map') {
       util.each(value, function(key, memberValue) {
-        req.httpRequest.headers[member.name + key] = memberValue;
+        req.httpRequest.headers[member.name + key.toLowerCase()] = memberValue;
       });
     } else if (member.location === 'header') {
       value = member.toWireFormat(value).toString();
--- a/test/services/s3.spec.coffee
+++ b/test/services/s3.spec.coffee
@@ -642,6 +642,10 @@ describe 'AWS.S3', ->
       url = s3.getSignedUrl('putObject', Bucket: 'bucket', Key: 'key')
       expect(url).to.equal('https://bucket.s3.amazonaws.com/key?AWSAccessKeyId=akid&Expires=900&Signature=J%2BnWZ0lPUfLV0kio8ONhJmAttGc%3D&x-amz-security-token=session')
 
+    it 'gets a signed URL for putObject with Metadata', ->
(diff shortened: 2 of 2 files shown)


chrisradek:
@bantic 
Thanks again for your pull request! I ended up moving the logic to lowercase the query parameters into the presigner. Since this was ultimately an issue with the presigner, I wanted to keep the change localized there. Extra thanks for also providing a unit test, that helps tremendously.
#756 has been merged and includes this fix.

bantic:
@chrisradek Awesome, thank you! Glad to see that make it in. Keeping the change to the presigner query params makes sense to me.

PR #751, opened 2015-10-12
Populates requestId on response objects regardless of the response status code.

chrisradek:
@jeskew


jeskew:
LGTM. Do error responses always have request IDs?

chrisradek:
The only time they wouldn't is if the request id is returned in the response's metadata, but in that case, there's no way to get the request IDs if we get an error.

jeskew:
:shipit:

PR #756, opened 2015-10-16
Fixes issue where S3 would reject presigned URLs that contained mixed-case Metadata

### LINKED ISSUE #747: s3.getSignedUrl does not normalize Metadata keys properly

chrisradek:
Fixes issue where S3 would reject presigned URLs that contained mixed-case Metadata

Includes tests from pull request #748 by @bantic
Fixes #747 

cc @jeskew @trevorrowe

### DIFF
--- a/lib/signers/presign.js
+++ b/lib/signers/presign.js
@@ -49,6 +49,9 @@ function signedUrlSigner(request) {
 
   AWS.util.each(request.httpRequest.headers, function (key, value) {
     if (key === expiresHeader) key = 'Expires';
+    if (key.indexOf('x-amz-') === 0) {
+      key = key.toLowerCase();
+    }
     queryParams[key] = value;
   });
   delete request.httpRequest.headers[expiresHeader];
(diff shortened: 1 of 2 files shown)


jeskew:
LGTM.

PR #770, opened 2015-10-31
Performance and a few readability improvements

RobGraham:
Yes, caching the length of arrays is a micro-optimization, something that V8 does automatically, however, IE, older versions of browsers as well as other iterable collections (non-array) can see a considerable speed improvement.


chrisradek:
Thanks for the pull request, we're inclined not to merge this at this time.

PR #775, opened 2015-11-04
Fix typo in document client example

AdityaManohar:
Fix example documentation in DynamoDB document client.

CC  @chrisradek


chrisradek:
Looks good. Merging into master.

Issue #785, opened 2015-11-10
S3: tracking overall upload progress for multiple files.


andrey-bahrachev:
Hey there!
Currently it's possible to keep track of upload progress only on per file basis, but there's no way of calculating the overall upload progress when multiple files are uploaded in parallel. It would be really great to have such ability. Thanks.

andrey-bahrachev:
Ah, found a way around using `this` in the upload progress callback.
```javascript
s3.upload(params, function (err, data) {
    ...
}).on('httpUploadProgress', function(progress) {
    // Here you can use `this.body` to determine which file this particular
    // event is related to and use that info to calculate overall progress.
});
```
I hope this will help someone else.
This wasn't obvious though and it still would be nice to have an easier way of doing this.

chrisradek:
@andrey-bahrachev 
We do have a pull request, #679 that would expose the object's `key` as part of the progress. I had a comment on it, but can look into expediting merging it in.

andrey-bahrachev:
@chrisradek Thanks for quick response.
That would be awesome!

chrisradek:
I've merged in #788. This will be available via npm with the next release of the SDK, otherwise you can install the SDK directly from github.

andrey-bahrachev:
Brilliant! Thanks a lot!

rossthedevigner:
@andrey-bahrachev Do you have an example how you're monitoring overall progress based on `this.body` info?

andrey-bahrachev:
Oh, it's been a while, so chances are that things might have changed since then, but here it is:
```
/**
 * Uploads file(s) directly to AWS S3.
 * 
 * @param form -  jQuery object representing the form.
 * @param {array} files -  Files to be uploaded (derived from HTML5 File API).
 * @param successCallback - Callback function to be called upon successful file(s) upload.
 */
uploadFiles: function(form, files, successCallback) {
    var filesData = [];
    var sizeTotal = 0;
    var loaded = [];
    for (var i = 0; i < files.length; i++) {
        filesData.push({name: files[i].name, size: files[i].size, type: files[i].type});
        sizeTotal += files[i].size;
        loaded[files[i].name] = 0;
    }
    app.checkStorage(form, filesData, function(resp) {
        // Initialize the Amazon Cognito credentials provider.
        AWS.config.region = resp.cognitoRegion;
        AWS.config.sslEnabled = true;
[...]

PR #788, opened 2015-11-11
Adds 'Key' to progress map emitted by ManagedUpload on httpUploadProg…

chrisradek:
…ress event

Resolves pr #679 and issue #785

/cc @jeskew


jeskew:
:shipit:

Issue #771, opened 2015-11-02
ManagedUploads and streams in the browser


kentor:
Is it possible to pass something that implements the Readable Stream interface as the Body of a ManagedUpload in the browser environment?

I'm trying to do some client side encryption on very large files. I was thinking of creating a `Readable Stream` from a `File`, then `pipe`ing it to some encryption function, and then passing that to `ManagedUpload`.

However, I believe the current sdk expects the body object to implement `slice` in the browser environment, otherwise it throws `Unsupported body payload`.

chrisradek:
@kentor 
There isn't currently a way to use streams with `ManagedUpload` in the browser.  What are you using to include support for streams in the browser?

It is possible to enable server side encryption so that objects are encrypted when stored in S3. More information on server-side encryption that S3 supports can be found here: http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html
However, I do realize server side encryption may not suit your needs.

kentor:
@chrisradek 
I was thinking of using the [filereader-stream](https://github.com/maxogden/filereader-stream) library to get a readable stream from a `File`. Then I'm using [crypto-browserify](https://github.com/crypto-browserify/crypto-browserify) to encrypt the stream.

Btw I did a quick and dirty modification of the `send` function in `lib/s3/managed_upload.js` to take a stream in the browser and it does appear to work.

I changed this
```js
    if (self.sliceFn) {
      self.fillQueue = self.fillBuffer;
    } else if (AWS.util.isNode()) {
      var Stream = AWS.util.nodeRequire('stream').Stream;
      if (self.body instanceof Stream) {
        ...
      }
    }
```

to this 
```js
    if (self.sliceFn) {
      self.fillQueue = self.fillBuffer;
    } else if (typeof self.body._read === 'function') {
      ...
    }
```

My app code is something like this:
```js
const crypto = require('crypto-browserify');
const fileReaderStream = require('filereader-stream');

...

const encrypt = crypto.createCipheriv(
  'aes-256-cbc',
  '3zTvzr3p67VC61jmV54rIYu1545x4TlY',
  '60iP0h6vJoEaJoEa'
);
const stream = fileReaderStream(file).pipe(encrypt);
const managedUpload = s3.upload({
  Body: stream,
  Key: `test/${file.name}`,
});
managedUpload.send();
```

chrisradek:
@kentor 
[...]

PR #795, opened 2015-11-13
Managed Upload: Allow readable streams in any environment

kentor:
This patch will allow any object that implements the ReadableStream
interface to be passed to the Body of a Managed Upload.

For example, one may create streams for the browser by using the
`stream` module with browserify.

addresses https://github.com/aws/aws-sdk-js/issues/771


jkudish:
I am trying to implement a stream uploader inside of an Electron application, and this is one of the barrier's that I've hit. This PR would help move things forward for me and anyone else who tries to implement something similar in an Electron application.

Issue #778, opened 2015-11-05
Multipart upload uploads a file with size of 5MB and not original file size


lonormaly:
Using Multipart upload, the uploaded file is eventually only 5MB and not the whole size.

I use leavePartsOnError so resuming is enabled, if it helps in figuring it out.
The frequency is higher than it was in previous version of the SDK (again, not sure if it's related) and stands on about 40% of downloads (!!!).
I use Cognito authentication.

![image](https://cloud.githubusercontent.com/assets/3742919/10966299/576de980-83bb-11e5-9eee-46fa6096e06d.png)

chrisradek:
@lonormaly 
Can you give some details on what is causing the initial multipart upload to abort? 

One thing that might be useful is to turn on the logger for S3 while you're troubleshooting (you can set the `logger` parameter to `console` when initializing S3 in the SDK).

Using the logger, you can verify that all the `uploadPart` operations for a given upload have the same `UploadId` and that you're seeing the expected number of `uploadPart` operations.  Also verify that the `completeMultipartUpload` command shows the number of parts you'd expect.

Let me know if this helps and what you find!

lonormaly:
The http progress announces "total" of 5MB and part 1 instead of total of original file size.
It happened to me especially when disconnecting from internet and reconnecting again for some reason.

kentor:
Yes in playing around the resuming feature, I can reproduce this using Buffers

script (I'm using node 4.2)

```js
const AWS = require('aws-sdk');

const credentials = new AWS.Credentials({
  accessKeyId: ...,
  secretAccessKey: ...,
});

const s3 = new AWS.S3({
  credentials,
  params: { Bucket: ... },
  region: ...,
});

const MB = 1024 * 1024;
var body = new Buffer(20 * MB);
body.fill('1', 0, 5 * MB);
body.fill('2', 5 * MB, 10 * MB);
body.fill('3', 10 * MB, 15 * MB);
body.fill('4', 15 * MB);

const upload = new AWS.S3.ManagedUpload({
  leavePartsOnError: true,
  params: { Body: body, Key: 'test' },
  service: s3,
});

upload.on('httpUploadProgress', function(progress) {
  console.log(progress);
});

upload.send = upload.send.bind(upload, (err, data) => {
  if (err) {
    console.log('error:', err);
  } else {
    console.log('success:', data);
  }
});

upload.send();

setTimeout(function() {
  console.log('abort!');
  upload.abort();
[...]

Issue #777, opened 2015-11-05
Managed Upload: Resume does not work when body is a stream


kentor:
Resume (calling `abort()` then `send()` again) works fine when the body is a File or Buffer, but nothing happens when calling `send()` again when the body is a stream. Sample code (node 4):

``` js
const AWS = require('aws-sdk');
const streamifier = require('streamifier');

const credentials = new AWS.Credentials({
  accessKeyId: ...,
  secretAccessKey: ...,
});

const s3 = new AWS.S3({
  credentials,
  params: { Bucket: ...},
  region: ...,
});

const MB = 1024 * 1024;
var body = new Buffer(20 * MB);
body.fill('1', 0, 5 * MB);
body.fill('2', 5 * MB, 10 * MB);
body.fill('3', 10 * MB, 15 * MB);
body.fill('4', 15 * MB);
body = streamifier.createReadStream(body);

const upload = new AWS.S3.ManagedUpload({
  leavePartsOnError: true,
  params: { Body: body, Key: 'test/test' },
  service: s3,
});

upload.on('httpUploadProgress', function(progress) {
  console.log(progress);
});

upload.send(function(err, data) {
  console.log(err, data);
});

setTimeout(function() {
  upload.abort();
  console.log('abort');
  setTimeout(function() {
    upload.send(function(err, data) {
      console.log(err, data);
    });
  }, 1000)
}, 1000);
```

chrisradek:
@kentor 
This is currently a limitation with resuming a managed upload. This limitation exists because once data is read from a stream, it can't be read again, so trying to automatically resume could result in some lost data.

That said, I would like to take a deeper look and see what options there are for us to support resuming on aborted streams.  For now I'm going to mark this as a feature request, and work on updating documentation to specify that resuming when using streams isn't currently supported.

kentor:
I see. I thought it would just keep what its read from the stream into a buffer until that part has uploaded successfully, or else read from the buffer again if the part was aborted and resumed.

ajredniwja:
Closing this issue, since the feature was implemented.

PR #796, opened 2015-11-13
Managed Upload: Support resuming streams

kentor:
Previously we would discard the read chunks of a stream on success or
failure of a part upload. This patch will make it discard the chunk
only on success so that when `leavePartsOnError` is set to `true`, these
parts can be reuploaded with the saved chunks still in memory.

There are also changes with how parts are queued when waiting for the
`createMultipartUpload` request. They're queued using the `self.parts`
object instead of via event handlers. This makes retry possible if the
`createMultipartUpload` request fails.

We also have to make sure not to discard the `partBuffers` from a
previous `send` when calling `send` again. Otherwise there will be lost
data.

This patch also fixes an inaccuracy with `totalUploadedBytes` when
uploads are aborted and resumed. Since we're retrying parts that have
failed, `totalUploadedBytes` should be reset to the last known
successfully uploaded bytes. We track that in the
`totalSuccessfullyUploadedBytes` property.

Also I am not sure why we call `stream.resume()` in the cleanup step.
This will drain the stream and will make it unreadable when the upload
is resumed.
addresses https://github.com/aws/aws-sdk-js/issues/777

This patch also probably fixes https://github.com/aws/aws-sdk-js/issues/778

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -155,6 +155,8 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
     self.failed = false;
     self.callback = callback || function(err) { if (err) throw err; };
 
+    self.resumeUploads();
+
     var runFill = true;
     if (self.sliceFn) {
       self.fillQueue = self.fillBuffer;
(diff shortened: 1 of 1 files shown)


lonormaly:
Amazing work! +1

PR #797, opened 2015-11-13
Add/error request ids

chrisradek:
This pr will update errors that come back from service calls to include request ids for easier access.  Currently, `requestIds` are available on the `response` object that is bound to the callback passed into service operation calls.  However, in some high-level abstractions, like `s3.upload`, the response isn't bound to the callback and if an error is returned by the service, there isn't a way to access the `requestId`.

This also exposes the `x-amz-id-2` header value provided by S3 as `extendedRequestId` for S3 errors.

/cc @jeskew

### DIFF
--- a/lib/services/ec2.js
+++ b/lib/services/ec2.js
@@ -57,5 +57,6 @@ AWS.util.update(AWS.EC2.prototype, {
         message: null
       });
     }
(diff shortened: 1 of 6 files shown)


jeskew:
:shipit:

PR #802, opened 2015-11-16
Allow doc client set type detection to understand 0 and ''

### LINKED ISSUE #801: DocumentClient put failure with { NS: ['0'] }
There's a bug in the conversion of a number set with a single `0` in it into a valid DynamoDB request. This test case demonstrates the issue (on a table with `id` as the HASH key):

```js
var AWS = require('aws-sdk');
var dynamo = new AWS.DynamoDB({ region: 'us-east-1' });
dynamo.putItem({
  TableName: 'rclark-test',
  Item: {
    id: { S: 'first' },
    remaining: { NS: [ '0' ] }
  }
}, function(err, data) {
  if (err) return console.error('Failed first %s', err);

  var docClient = new AWS.DynamoDB.DocumentClient({ service: dynamo });

  var req = docClient.put({
    TableName: 'rclark-test',
    Item: {
      id: 'second',
      remaining: docClient.createSet([0])
    }
  }, function(err) {
    console.log(req.params);
    if (err) console.error('Failed second %s', err);
  });
});
```

On the second request, `req.params` are incorrect:
```json
{ 
  "TableName": "rclark-test",
  "Item": { 
    "id": { "S": "second" }, 
    "remaining": {} 
  } 
}
```

... and the request fails validation.

Associated PR at #802

rclark:
Fixes #801

### DIFF
--- a/lib/dynamodb/set.js
+++ b/lib/dynamodb/set.js
@@ -20,19 +20,17 @@ var DynamoDBSet = util.inherit({
   detectType: function() {
     var self = this;
     var value = self.values[0];
-    if (value === '' || value === 0 || value) {
-      if (typeOf(value) === 'String') {
-        self.type = 'String';
-      } else if (typeOf(value) === 'Number') {
-        self.type = 'Number';
-      } else if (typeOf(value) === 'Binary') {
-        self.type = 'Binary';
-      } else {
-        throw util.error(new Error(), {
-          code: 'InvalidSetType',
-          message: 'Sets can contain string, number, or binary values'
-        });
-      }
+    if (typeOf(value) === 'String') {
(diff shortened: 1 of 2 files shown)


AdityaManohar:
This edge case should also be unit tested.

```coffeescript
    it 'supports sets with falsy values', ->
      expect(docClient.createSet([0]).type).to.equal('Number')
      expect(docClient.createSet(['']).type).to.equal('String')
```

chrisradek:
@rclark 
These final changes look good. Thank you for providing this pull request. Again, your contribution is greatly appreciated!

rclark:
Glad that his is helpful. Any insight into when it might land in a patch version?

Issue #800, opened 2015-11-15
Reading KMS encrypted S3 objects


tracend:
Recently [AWS SES](aws.amazon.com/ses) started accepting inbound email, saved in an S3 bucket. There's an option to encrypt the data, which is great, otherwise everything is exposed as plain text in your bucket. 

The problems start with only accepting a KMS key for encryption. That means it needs to be managed from the AWS side. 

The official documentation says on the subject: 

> Your mail is encrypted by Amazon SES using the Amazon S3 encryption client before the mail is submitted to Amazon S3 for storage. It is not encrypted using Amazon S3 server-side encryption. This means that you must use the Amazon S3 encryption client to decrypt the email after retrieving it from Amazon S3, as the service has no access to use your AWS KMS keys for decryption. This encryption client is available with the AWS Java SDK and AWS Java Ruby only. 

http://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-action-s3.html

I tracked down the Ruby code here:
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb#L36
http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Encryption/Client.html#client-instance_method

How can we bring that functionality to the JS lib, so we can decrypt and read our messages using Node.js?

Please let me know if I'm missing something. 

Cheers

chrisradek:
@tracend 
Thanks for the feedback! This has been tagged as a feature request. We tend to prioritize which feature requests get worked on based on community feedback and by looking at what other SDKs are doing. We also accept pull requests from the community.

tracend:
Thanks @chrisradek, 

Was considering contributing a pull request but thought I should ask first in case I missed an existing method of performing said task or there were special conventions in dealing with KMS I needed to be aware of. 

I spent some time on this yesterday and was able to decrypt the messages, although they came encrypted as ```AES/GCM/NoPadding```, not as stated in the Ruby lib ```AES/CBC/PKCS5Padding```. I'm not sure if this is variable for SES encrypted messages...

I'll form a proper solution and use the structure of the Ruby lib as a guide to submit a pull request. 

Thanks again.

rsweetland:
We would very much like to see this as well. It seems fitting given that the examples from AWS use Node.js with Lambda to receive inbound email.

[...]

PR #809, opened 2015-11-23
S3 Encrypted Client

tracend:
This request is a direct follow-up to the issue #800 , raised a few days ago.

As previously mentioned, I used the Ruby lib as a basis to reconstruct the logic for the JavaScript lib. Certain liberties were taken, some related with the JavaScript syntax and some with the existing conventions of the AWS SDK; the linting script was responsible for a few tweaks as well.

Please note, the new code relates with the ability to automatically encrypt/decrypt S3 objects from the client-side. It is not currently addressing SSE (Server-side Encryption), which I'm not sure if it's under the `AWS.S3` domain. 

At the current state my main objective, to read KMS encrypted SES messages is fulfilled. This can be easily done like this:

```
var params = {
  Bucket: '{{BUCKET_NAME}}',
  Key: '{{BUCKET_ADDRESS}}',
};

var store = new AWS.S3.Encryption.Client({
  accessKeyId: '...',
  secretAccessKey: '...',
  region: '...', 
  sslEnabled: true,
  signatureVersion: 'v4',
});

var downloader = store.getObject(params, function(err, response ){
  // output...
  res.end( response.Body );
});
```

I'm also satisfied that the logic is laid out so it can grow. Additional abilities are in place, but not fully tested:
### putObject
- Encrypt and upload content using a `KMSKeyId` and `EncryptionKey`.
- Encrypt and upload content using only a `KMSKeyId`, with the ``EncryptionKey` automatically generated.
- Encrypt content using a custom `kms_client`, passed with the options of  `AWS.S3.Encryption.Client`.
### getObject
- Decrypt and download content using a `KMSKeyId` and `EncryptionKey`.
- Decrypt and download content using only a `KMSKeyId`, with the ``EncryptionKey` resourced from the file headers (as _'x-amz-key-v2'_).
- Decrypt content using a custom `kms_client`, passed with the options of  `AWS.S3.Encryption.Client`.

There is also scaffolding for allowing an instruction file, which will contain the encryption context of an encrypted file, but that functionality is at present disabled. 

Some of these features I didn't want to complete by introducing my own conventions. For example, arbitrary including the `x-amz-key-v2` key in the putObject response (to allow the user to save that info).

In addition, I tried to incorporate the unit tests from the Ruby lib, although they are failing, mostly because the stub requests for the KMS and S3 services aren't setup properly. I'm sure this can be easily amended by someone more familiar with the inner workings of the API/SDK. 

I understand that further adjustments may be required before this code can be integrated to the main lib. That's why I'm committing it as a separate branch. Frankly, I feel it is a significant component that shouldn't be addressed by just one contributor anyway. 

On a positive note, everything is contained under the `AWS.S3.Encryption` namespace and should not interfere with any of the existing functionality. I can use my branch to decrypt SES messages and there's an open invitation for anyone that wants to work on the other features. 

Let me know if you have any questions.

### DIFF
--- a/lib/s3/encryption/materials.js
+++ b/lib/s3/encryption/materials.js
@@ -1,0 +1,73 @@
+var AWS = require('../../core');
+
+AWS.S3.Encryption.Materials = AWS.util.inherit({
+
+  constructor: function(options) {
+    this.initialize(options);
+  },
+  /*
+   * @option options [required, OpenSSL::PKey::RSA, String] :key
+   *   The master key to use for encrypting/decrypting all objects.
+   *
+   * @option options [String<JSON>] :description ('{}')
+   *   The encryption materials description. This is must be
+   *   a JSON document string.
+   */
+  initialize: function( options  ) {
+    options = options || {};
+    this.key = this.validate_key(options['key']);
+    this.description = this.validate_desc(options['description']);
+  },
+
+  // @return [OpenSSL::PKey::RSA, String]
+  //attr_reader: key,
+
+  // @return [String<JSON>]
+  //attr_reader: description,
+
+  /**
+   * @api private
+   */
+  validate_key: function( key ) {
+    var type = (typeof key);
+    switch ( type ) {
+      case 'OpenSSL::PKey::RSA':
+        return key;
+      break;
+      case 'string':
+        if ( [32, 24, 16].indexOf( key.length ) > -1 ) {
+          return key;
+        } else {
+          var msg = 'invalid key, symmetric key required to be 16, 24, or ';
+          //msg << "32 bytes in length, saw length 31"
+          //raise ArgumentError, msg
+          AWS.util.error(new Error(), {
+            message: msg
+          });
+        }
+      break;
+      default:
+        var msg = 'invalid encryption key, expected an OpenSSL::PKey::RSA key';
+        //msg << "(for asymmetric encryption) or a String (for symmetric "
+        //msg << "encryption)."
+        //raise ArgumentError, msg
+        AWS.util.error(new Error(), {
+          message: msg
+        });
+    }
+  },
+
+  validate_desc: function(description) {
+    var description = JSON.parse(description);
+
+    //rescue Json::ParseError
+    var msg = 'expected description to be a valid JSON document string';
+    //raise ArgumentError, msg
+    AWS.util.error(new Error(), {
+      message: msg
+    });
+  }
+
+});
+
+module.exports = AWS.S3.Encryption.Materials;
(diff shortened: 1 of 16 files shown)


leore:
Hey Guys, Any update on this?

Is this related to http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html

I see that the javascript library does not yet support Client Side Encryption

cc @robinjoseph08

jeskew:
Hi @leore,

This PR isn't ready to be merged in at the moment, both because the last commit had some failing tests and because it doesn't seem possible to use it in a browser. I see that files over one megabyte are decrypted into a temp file, and I'm not sure what the performance characteristics of doing that through filesystem-browserify would be.

leore:
What is the best way to do client side encryption through node?

mallocator:
**+1**  Very much need this to decrypt data uploaded from a java client using symmetric client side encryption in my nodejs server.

Only option I see right now is trying to reverse engineer whatever was done in Java.

lib/s3/encryption/materials.js
```diff
@@ -0,0 +1,73 @@
...
+          return key;
+        } else {
+          var msg = 'invalid key, symmetric key required to be 16, 24, or ';
+          //msg << "32 bytes in length, saw length 31"
+          //raise ArgumentError, msg
+          AWS.util.error(new Error(), {
+            message: msg
+          });
+        }
+      break;
```
mallocator: would be great to support buffers as keys

albertovasquez:
+1 This is a critical feature needed.  If Amazon wanted to allow easy adoption they should have a technical solution, either blog on how to accomplish decrypting or include in sdk.

I'm using this PR and it is working perfectly.  Thanks for making this available.

tracend:
Thanks @andylibrian for pointing out that the decryption may fail sometimes on the ```finalize``` method. 

According to the documentation: 

> Once the cipher.final() method has been called, the Cipher object can no longer be used to encrypt data. Attempts to call cipher.final() more than once will result in an error being thrown.

As suggested, I've wrapped it with a try catch to silent the error (temporarily). If anyone has a more proper solution, feel free to contribute.

guerrerocarlos:
Alternatively, if you have 'being able to do ranged GET' among your requirements, [stream-cipher](https://github.com/guerrerocarlos/stream-cipher/) can be used for client side encription (node.js).

chrisradek:
Thanks for getting the conversation started. We have been investigating a separate encryption client where this functionality would better live.

tracend:
Oh, I think I did more than just start the conversation, if you were willing to give me credit ;)

tracend:
The code contributed as a pull request in 2015 is now available as a separate module:

https://github.com/makesites/aws-s3-encryption-client

PR #824, opened 2015-12-02
Adds ability to specifiy a custom user agent to be appended to user a…

chrisradek:
…gent headers.

Added `customUserAgent` to the AWS.config object so that it can be set globally or at the service level.

/cc @jeskew


jeskew:
:shipit:

PR #830, opened 2015-12-06
fixed 'tableName =' to 'tableName: ' in document_client

nalindak:
there were quite a few instances where it used 'tableName =' inside params object


chrisradek:
@nalindak 
Thanks for the doc fix!

PR #839, opened 2015-12-14
No longer converts unauthenticated API calls to GET operations. Speci…

chrisradek:
…fically affects STS operations.

This change is to handle the scenario where a token or SAML assertion is too large to be handled properly in a query string.

/cc @jeskew


jeskew:
:shipit:

Issue #405, opened 2014-11-08
DynamoDB JS CRC32CheckFailed on iPhone


bpevangelista:
I have a DynamoDB query that always fails with "CRC32CheckFailed" if I don't add "Limit" to the query on iPhone only. If I limit it to a small value, e.g. 10, it always work if I limit it to 20 or more it always fails.

On Chrome/Firefox on Mac it always work without limit. If I change the AWS-JS code to disable the crc32 check things work on iPhone without a limit and the data I get back looks good/valid to me.

```javascript
var keyConditions = { 
    'hash': { ComparisonOperator: 'EQ', AttributeValueList: [ {S: "something"} ] } 
};

var queryParams = {TableName: "test5", KeyConditions: keyConditions,
    //Limit: 10, // If enabled make things work on iPhone
};
````

LiuJoyceC:
Closing due to inactivity. If the workaround provided above did not work, please let us know.

kalevet:
I'm having the same issue with limitation for iPhone. the workaround doesn't seem to work...
Any ideas?

LiuJoyceC:
@kalevet 
Are you saying that you get a `CRC32CheckFailed` error from DynamoDB even when you construct your DynamoDB client with the `dynamoDbCrc32` config set to false? `var dynamodb = new AWS.DynamoDB({dynamoDbCrc32: false});` When you later check the `config` property on the new instance, do you see what the value is stored at `dynamodb.config.dynamoDbCrc32`, and is it correctly set to false?

PR #840, opened 2015-12-14
Adds retry logic to dynamodb crc32 checksum errors

chrisradek:
This moves the CRC32 check to occur prior to parsing a JSON response body. Additionally, the SDK will retry a request when a crc32 checksum mismatch is encountered.

/cc @jeskew

### DIFF
--- a/lib/services/dynamodb.js
+++ b/lib/services/dynamodb.js
@@ -7,7 +7,9 @@ AWS.util.update(AWS.DynamoDB.prototype, {
    */
   setupRequestListeners: function setupRequestListeners(request) {
     if (request.service.config.dynamoDbCrc32) {
+      request.removeListener('extractData', AWS.EventListeners.Json.EXTRACT_DATA);
       request.addListener('extractData', this.checkCrc32);
+      request.addListener('extractData', AWS.EventListeners.Json.EXTRACT_DATA);
     }
   },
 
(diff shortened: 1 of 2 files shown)


AdityaManohar:
This could cause a large number of retries on browsers when the payload is greater than a certain size and `dynamoDbCrc32` is set to `false`. 

This is because of the way DynamoDB calculates CRC32 checksums. See #405

chrisradek:
@AdityaManohar 
Thanks for that information. This didn't come up when testing in the browser due to the header containing the crc32 checksums not being exposed. I'd suspect that users would disable the crc32 check if they were accepting gzipped content, but this could cause breaking changes for users that currently look for that crc32 error.

I'll update the the pr to still throw an error if the crc32 check fails, but leave retry logic up to the user. This will allow the user to better handle instances where the data integrity check of a response fails and causes other errors to supersede this one, without negatively impacting performance for existing users.

chrisradek:
@AdityaManohar 
I spoke too soon. The SDK already attempts to retry on crc32 failures. What this change really does then is bump the crc32 check to occur before the data is extracted from the response. This handles scenarios where the crc32 check detects data integrity issues that may cause the data extraction to throw an error.

jeskew:
Given that this change is just reordering the crc32 check and the JSON unmarshalling, I don't think there's a danger of regressing #405. :ship:

PR #841, opened 2015-12-15
Add a CloudFront signer

jeskew:
/cc @chrisradek

### DIFF
--- a/lib/cloudfront/signer.js
+++ b/lib/cloudfront/signer.js
@@ -1,0 +1,182 @@
+var crypto = require('crypto'),
+    url = require('url'),
+    AWS = require('../core'),
+    base64Encode = AWS.util.base64.encode,
+    inherit = AWS.util.inherit;
+
+var queryEncode = function (string) {
+    var replacements = {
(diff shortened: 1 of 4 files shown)


lib/cloudfront/signer.js
```diff
@@ -0,0 +1,182 @@
...
+                if (signatureHash.hasOwnProperty(key)) {
+                    parsedUrl.query[key] = signatureHash[key];
+                }
+            }
+
+            var signedUrl = url.format(parsedUrl);
+            if (determineScheme(options.url) === 'rtmp') {
+                cb(null, getRtmpUrl(signedUrl));
+            } else {
+                cb(null, signedUrl);
```
chrisradek: Here and in getSignedCookie, we should probably check that a callback was supplied before calling it, and return the signedUrl/cookieHash immediately if one wasn't provided.  This is how s3.getSignedUrl works, for example.

jeskew:
@chrisradek I updated the signer to support synchronous calls.

chrisradek:
Merging!

PR #845, opened 2015-12-15
Allowing use of bare objects to be passed in to utils.each

### LINKED ISSUE #844: object has no method 'hasOwnProperty' when sparse key value object sent to any service as configuration
Version: 2.2.22
Environment: Node 4.2.1
Dependencies: None
Reproduction: http://requirebin.com/?gist=84d37dcd704f1823bddf

Explanation: When passing options to the a service constructor as a bare key/value map (using ``` Object.create(null)```), the utils.each function will throw an error on calling object.hasOwnProperty

owencraig:
Resolves #844. Allows consumers to pass bare key/value pair objects into service constructors (or elsewhere in the codebase) and for the services to pull configs or the default version numbers

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -533,7 +533,7 @@ var util = {
 
   each: function each(object, iterFunction) {
     for (var key in object) {
-      if (object.hasOwnProperty(key)) {
+      if (Object.prototype.hasOwnProperty.call(object, key)) {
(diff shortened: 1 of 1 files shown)


chrisradek:
@owencraig 
Thank you for your contribution! This change looks solid, merging!

PR #857, opened 2015-12-29
Multifile upload

islam-taha:
Also need retrieving key in info to help in multi-file upload process and retrieving key in the callback of the upload function in order to keep track of files after finishing uploading.


chrisradek:
@islam-taha 
Thanks for the pull request. The file you modified is a distributable version of the SDK that is built for each release. Instead, please make the change to the source file:
https://github.com/aws/aws-sdk-js/blob/master/lib/s3/managed_upload.js#L572

islam-taha:
I have reverted old commits and added the new one, thanks sir.

chrisradek:
@islam-taha 
Accepted! Thanks for your contribution!

PR #858, opened 2015-12-31
remove unnecessary IF

sabrinaluo:
`this.errors.length` won't be changed. 
If there is any reason why here needs two `IF`, please let me know. thank you. :)

### DIFF
--- a/lib/param_validator.js
+++ b/lib/param_validator.js
@@ -10,12 +10,10 @@ AWS.ParamValidator = AWS.util.inherit({
 
     if (this.errors.length > 1) {
       var msg = this.errors.join('\n* ');
-      if (this.errors.length > 1) {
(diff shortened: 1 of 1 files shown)


lib/param_validator.js
```diff
@@ -10,12 +10,10 @@ AWS.ParamValidator = AWS.util.inherit({
 
     if (this.errors.length > 1) {
       var msg = this.errors.join('\n* ');
```
jeskew: This could be combined with the next line.

jeskew:
Thanks, @sabrinaluo! That does indeed look like a vestigial `if`.

chrisradek:
@sabrinaluo 
Thanks for the contribution! Merging it in.

PR #865, opened 2016-01-07
Adding the ability to perform strict validation.

mtdowling:
This commit updates the JavaScript SDK to allow it to perform strict
validation client-side using the enum, min, max, and pattern traits.
Validating based on these traits are (and should be) an opt-in for
customers as overly strict client-side validation can cause unnecessary
inconvenience when a service team relaxes any validation constraints.
However, being able to perform strict validation would enable tooling
that utilizes the JavaScript SDK to have more confidence on whether or
not inputs and outputs match a service's model at a point in time.

Note that this commit does not switch the JS SDK over to using the
regular models without minification.

/cc @chrisradek @jeskew

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -40,6 +40,12 @@ require('./credentials/credential_provider_chain');
  *   @return [Boolean] whether input parameters should be validated against
  *     the operation description before sending the request. Defaults to true.
  *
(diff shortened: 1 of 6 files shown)


chrisradek:
:shipit:

PR #869, opened 2016-01-11
Add embed parameter to resources method. Closes #764

### LINKED ISSUE #764: Embed option for getResources on API Gateway
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html#getRestApis-property

The docs specify the existence of an `embed` option but no more details about how to use it. I have used it directly via the rest api, but can't seem to get it to work in the SDK. When adding `embed: true` or `embed: "methods"` the to the `params`object the SDK throws an invalid params error:

```
{ [UnexpectedParameter: Unexpected key 'embed' found in params]
  message: 'Unexpected key \'embed\' found in params',
  code: 'UnexpectedParameter',
  time: Mon Oct 26 2015 23:24:25 GMT-0400 (EDT) }
```

southpolesteve:


chrisradek:
@southpolesteve 
Thank you for your pull request. We typically don't like to add customizations to the api json files directly, because they would be overwritten each time that service was updated. That said, we've identified what the correct representation of `embed` should be and can ensure it is included with future service updates. I'll work on testing/including that model shortly.

chrisradek:
This has finally been added with today's release `v2.43.0`.

PR #872, opened 2016-01-12
Remove mention of the old .done()/.fail() methods

nhinds:
These were removed in #22 but still exist in the [public documentation](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-making-requests.html)


chrisradek:
@nhinds 
Thanks for the contribution! Merging these changes now, they'll be available in the next release of the SDK.

Issue #780, opened 2015-11-06
getSignedUrl creates invalid signature when CacheControl parameter is present


Prinzhorn:
Here's a demo repo https://github.com/Prinzhorn/aws-s3-signature-cache-control-issue

Relevant code:

```js
var uuid = require('uuid');
var AWS = require('aws-sdk');

AWS.config.update({
    region: 'eu-central-1'
});

var s3 = new AWS.S3();

var params = {
    Bucket: process.env.BUCKET,
    Expires: 3600,
    Key: uuid.v4(),
    ContentType: 'image/jpeg',

    //Comment this line out and the signature is valid.
    CacheControl: 'max-age=' + 10000
};

s3.getSignedUrl('putObject', params, function(err, url) {
    if(err){
        return console.error(err);
    }

    console.log(url);
});
```

Basically if I remove the `CacheControl` from the params object, it works. Otherwise the upload with the generated URL fails because of a wrong signature. Maybe I'm doing something stupid there?

Prinzhorn:
This unwritten law kicked in again, where you get closer to the issue once you ask for help. It seems like `ContentType` is ignored, albeit the signature is valid. So how do I sign a POST/PUT request with all parameters except the file itself without having to duplicate the parameters again when making the request? I mean `getSignedUrl` doesn't seem like the right choice for `putObject`, is it? I guess I have to supply all the parameters separate from the URL and only need the signature itself.

chrisradek:
@Prinzhorn 
Thanks for the example code. My initial suspicion is that the Cache-Control header needs to be provided when making the request because it is included as a signed header in the presigned URL (X-Amz-SignedHeaders). I'm looking into if this is truly necessary.

chrisradek:
@Prinzhorn 
It is indeed necessary to provide the `Cache-Control` header in your request if it is included as a parameter to `getSignedUrl`. It is possible to only set the `Cache-Control` header when making your request, and leave it out of your `getSignedUrl` call.  If it is included in `getSignedUrl` however, it is essentially enforcing that `Cache-Control` must be sent with the same value in the request.

Let me know if this explanation helps, or if you'd like to discuss further.

Prinzhorn:
[...]

Issue #675, opened 2015-08-04
ContentMD5 for s3.getSignedUrl


thattommyhall:
I can't get a signed request to work with ContentMD5.

If I do 

```
var s3_params = {
  Bucket: bucketName,
  ContentType: 'image/jpeg',
  ContentMD5: "f0oxEyT+4qR3E/70ID0PfA==",
  ACL: 'public-read',
  Key: 'TESTDIRECT.jpeg',
  Body: fs.readFileSync('/home/thattommyhall/TEST.jpg')
};


s3.putObject(s3_params, function(err, data){
  if(err){
    debug('put', 'error', err);
  }
  else {
    // debug('put', data);
  }
});
```

It works, but 

```
var s3_params_sign = {
  Bucket: bucketName,
  ContentType: 'image/jpeg',
  ContentMD5: "f0oxEyT+4qR3E/70ID0PfA==",
  ACL: 'public-read',
  Key: 'TESTsigned.jpeg',
  Expires: 6000,
};

s3.getSignedUrl('putObject', s3_params_sign, function(err, url){
  if(err){
    debug('error', err);
  }
  debug(url);
});

```

will not work if I have ContentMD5.

I have tried adding Content-MD5 headers to the put (as the content type was not set until I added headers).

I made a project at https://github.com/thattommyhall/aws-sdk-bug, you should be able to run 

```
$ node example.js | python upload.py 
```
(but I have also tried it with `curl --upload-file` )

AdityaManohar:
@thattommyhall I see [here that you are creating the signed url](https://github.com/thattommyhall/aws-sdk-bug/blob/master/example.js#L36) *without* the `ContentMD5` parameter set.

You will have to create that signed url with the `ContentMD5` parameter *as well as* set the `Content-MD5` header when PUTting to the signed URL.

For example:
```javascript
// NOTE: This is a private API and may change in the future
var md5 = AWS.util.crypto.md5('hello' ,'base64');

var params = {
  Bucket: 'bucket', 
  Key: 'key', 
  ContentMD5: md5, 
  ContentType: 'text/plain'
}
var url = s3.getSignedUrl('putObject', params);

// Using jQuery in the browser
$.ajax({
  url: url, 
  method: 'PUT', 
  data: 'hello', 
  contentType: 'text/plain', 
  headers: {
    'Content-MD5': md5
  },
  success: callback
});

```
The above code works fine when PUTting to a pre-signed URL with Content-MD5 specified.

[...]

PR #874, opened 2016-01-15
Fixes S3 getSignedUrl for v4 signers when using CacheControl in param…

LiuJoyceC:
…eters.

Reference #780 

CC: @chrisradek @jeskew

### DIFF
--- a/lib/signers/v4.js
+++ b/lib/signers/v4.js
@@ -63,6 +63,9 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
     if (this.request.headers['Content-MD5']) {
       qs['Content-MD5'] = this.request.headers['Content-MD5'];
     }
+    if (this.request.headers['Cache-Control']) {
+      qs['Cache-Control'] = this.request.headers['Cache-Control'];
+    }
 
     // need to pull in any other X-Amz-* headers
     AWS.util.each.call(this, this.request.headers, function(key, value) {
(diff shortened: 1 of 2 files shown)


lib/signers/v4.js
```diff
@@ -63,6 +63,9 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
     if (this.request.headers['Content-MD5']) {
       qs['Content-MD5'] = this.request.headers['Content-MD5'];
     }
+    if (this.request.headers['Cache-Control']) {
```
chrisradek: So looking at the commit log, it looks like this is the 2nd time we've seen an issue like this. The first time was with Content-MD5 (#675).
This fixes the issue with #780, but there could still be issues in the future. For example, if you take the `s3.getSignedUrl` example in #780, and add the parameter `ContentLanguage`, we'll have the same issue.

Instead, it might be worth taking a look at removing line 74 below. According to the S3 v4 signature docs:
http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
`For the purpose of calculating an authorization signature, only the host and any x-amz-* headers are required; however, in order to prevent data tampering, you should consider including all the headers in the signature calculation.`
So it appears we should be able to support signing headers other than `x-amz` headers.

I'm not sure why the `x-amz` check was added below, so we may need to dig into what other ramifications removing that may have.

chrisradek:
:shipit:

PR #884, opened 2016-01-21
adding support to run within a web worker context

cesarpachon:
if the library is loaded inside a webworker, the window object is undefined. instead, self can be used as the global root object.

### DIFF
--- a/lib/browser.js
+++ b/lib/browser.js
@@ -18,3 +18,4 @@ require('./http/xhr');
 
 if (typeof window !== 'undefined') window.AWS = AWS;
 if (typeof module !== 'undefined') module.exports = AWS;
+if (typeof self !== 'undefined') self.AWS = AWS;


lib/browser.js
```diff
@@ -18,3 +18,4 @@ require('./http/xhr');
 
 if (typeof window !== 'undefined') window.AWS = AWS;
 if (typeof module !== 'undefined') module.exports = AWS;
+if (typeof self !== 'undefined')module.exports = AWS;
```
chrisradek: @cesarpachon 
It looks like the tests are failing on the linter because there isn't a space before `module`. Do you mind updating that?

cesarpachon: done!

cesarpachon:
hi @chrisradek wonder if this PR can be merged.. thks!

lib/browser.js
```diff
@@ -18,3 +18,4 @@ require('./http/xhr');
 
 if (typeof window !== 'undefined') window.AWS = AWS;
 if (typeof module !== 'undefined') module.exports = AWS;
+if (typeof self !== 'undefined') module.exports = AWS;
```
chrisradek: @cesarpachon 
Did you mean to set `self.AWS = AWS;` instead of `module.exports = AWS;`?

cesarpachon: @chrisradek you are right, I blame myself. I fixed and squashed my commits. thank you.

cesarpachon:
hi! some chance to merge this? :)

chrisradek:
@cesarpachon 
Sorry about the delay, I'll check this out today. One thing I'm curious about is if xml-based services will work within web workers. My assumption is they won't unless a DOMParser is shimmed in, but this may be worth merging in anyway as a starting point.

chrisradek:
@cesarpachon 
I'm going to merge this pull request in. There are still some issues where XML-based services don't work out of the box within web workers due to the absence of the DOMParser in web workers. However, there's still value in running the SDK in a web worker even as is, so I don't want to prevent users from doing so if they'd like.

Thanks for the pr!

PR #888, opened 2016-01-28
Allow calling createBucket with only bound params. Resolves #799.

### LINKED ISSUE #799: params in object 'new AWS.S3()' doesn't work for regions other than default region
the code before only works for default region : us-east-1, not for other regions if you set with ' AWS.config.region = 'xx-xxxxxxx-x''.

var s3 = new AWS.S3({params: {Bucket: 'myBucket', Key: 'myKey'}});
s3.createBucket(function(err) {
  if (err) { console.log("Error:", err); }
  else {
    s3.upload({Body: 'Hello!'}, function() {
      console.log("Successfully uploaded data to myBucket/myKey");
    });
  }
});

jeskew:
`AWS.S3.createBucket` was not handling cases where users passed in only a callback.

/cc @chrisradek @LiuJoyceC

### DIFF
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -38,9 +38,8 @@ AWS.util.update(AWS.S3.prototype, {
     request.addListener('beforePresign', this.prepareSignedUrl);
   },
 
-  /*
+  /**
    * @api private
-   *
    */
   validateScheme: function(req) {
     var params = req.params,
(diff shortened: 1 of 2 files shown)


lib/services/s3.js
```diff
@@ -411,7 +409,10 @@ AWS.util.update(AWS.S3.prototype, {
     // This chunk of code will set the location constraint param based
     // on the region (when possible), but it will not override a passed-in
     // location constraint.
-    if (!params) params = {};
+    if (typeof params === 'function') {
+      callback = params;
+      params = {};
+    }
```
chrisradek: It might be useful to add one more check after the if block to set params equal to itself or and empty object to protect us in the unlikely event someone passes in a value like `null`. Imagine this would be a rare edge case but we do this already for other operations.

chrisradek:
:shipit:

PR #894, opened 2016-02-03
Fixes error response for truncated XML response in Node, for issue #886.

### LINKED ISSUE #886: ListObjects failure on truncated response
I have an application that paginates through S3 ListObjects responses to find all the keys in a particular prefix. I noticed that sometimes it stopped early, without throwing any errors.

If I build a mock S3 server that begins providing ListObjects response XML, then abruptly closes the connection, the sdk silently accepts the response, and provides what looks like the end of the listing.

An http server like this one:

```js
var server = http.createServer(function (req, res) {
  res.writeHead(200, { 'Content-Length': 500 });
  res.write('<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>bucket</Name><Prefix>list-truncated</Prefix><Marker>next-one</Marker><MaxKeys>1000</MaxKeys><IsTruncated>true</IsTruncated><Contents><Key>first-one</Key><LastModified>2015-09-24T21:09:21.000Z</LastModified><ETag>&quot;9a194fc78eaede');
  return req.socket.destroy();
});
```

If I monitor the `extractData` event, I find that the response was accepted as "valid":

```js
s3.listObjects(params, function(err, data) {
  console.log(err); // null
  console.log(data); // { Contents: [], CommonPrefixes: [] }
  console.log(data.IsTruncated); // undefined
  console.log(data.Marker); // undefined
  console.log(data.Contents.slice(-1)[0]); // undefined
}).on('extractData', function(resp) {
  console.log(resp.error); // null
  console.log(resp.data); // { Contents: [], CommonPrefixes: [] }
});
```

It seems to me that this body ought to fail XML parsing... or fail since the `body.length` !== `headers.ContentLength`... or something other than present what's likely to be detected as a legitimate end to the listing.

LiuJoyceC:
S3.listObjects was not passing an error to the callback when the XML response was truncated. This fix corrects this issue in Node, and makes the error retryable. Resolves #886 

/cc @chrisradek

### DIFF
--- a/lib/xml/browser_parser.js
+++ b/lib/xml/browser_parser.js
@@ -14,24 +14,43 @@ DomXmlParser.prototype.parse = function(xml, shape) {
         result = parser.parseFromString(xml, 'text/xml');
       } catch (syntaxError) {
         throw util.error(new Error('Parse error in document'),
-          {originalError: syntaxError});
+          {
+            originalError: syntaxError,
+            code: 'XMLParserError',
+            retryable: true
+          });
       }
 
       if (result.documentElement === null) {
-        throw new Error('Cannot parse empty document.');
+        throw util.error(new Error('Cannot parse empty document.'),
+          {
+            code: 'XMLParserError',
+            retryable: true
+          });
       }
 
       var isError = result.getElementsByTagName('parsererror')[0];
       if (isError && (isError.parentNode === result ||
-          isError.parentNode.nodeName === 'body')) {
-        throw new Error(isError.getElementsByTagName('div')[0].textContent);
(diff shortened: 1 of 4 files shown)


chrisradek:
This will also close #655.

This is ok to ship, but I wouldn't close #886 until this case is resolved in the browsers we support as well.

chrisradek:
Just confirming, this works in all the major browsers we support, right?

If so, :shipit:

LiuJoyceC:
Yes, this has been tested in all major browsers we support. Thanks!

PR #895, opened 2016-02-04
test & support node 4.2 (latest LTS) and 5.x (Stable)

revington:
The change at `dist-tools/browser-builder.js` was required because of npm@3 flat structure.


chrisradek:
@revington 
Thanks, looks good to me!

PR #898, opened 2016-02-05
Fixes issue when specifying S3 SSE-C-Key when creating presigned URLs

### LINKED ISSUE #896: using pre-signed url for PUT returns SignatureDoesNotMatch using SSE-C
If I use `ServerSideEncryption` this entire workflow is perfect. I can upload a file using the pre-signed url and offer a link to my end users without having to send any header information (in the PUT or GET).  

However, switching over to use SSECustomerKey causes the same workflow to error out. 

It generates the proper signed url for a PUT request, but when I attempt to upload it fails with a SignatureDoesNotMatch error.

I have seen in some other tickets that you have to send the same headers across when using a signedurl. However, I'm not sending those headers across when I use ServerSideEencryption and it works fine. 

I was hoping and assumed that because the signed url includes the header information in the query string that it would just work. Which it does for sse. The moment I switch it to sse-c it errors out.

If I do have to send the headers for sse-c.. how am I supposed to offer a link for people to click on our site? And why does it work for sse, but not sse-c?

chrisradek:
Fixes #896 

The issue was the `x-amz-server-side-encryption-customer-key-MD5` header getting hoisted to the query string twice when creating a presigned url. This was due to the presigner lower-casing all `x-amz-` headers when building the query string. This change only lower-cases `x-amz-meta-` headers since the lowercase logic was added to support S3 object metadata.

/cc @LiuJoyceC

### DIFF
--- a/lib/signers/presign.js
+++ b/lib/signers/presign.js
@@ -49,7 +49,7 @@ function signedUrlSigner(request) {
 
   AWS.util.each(request.httpRequest.headers, function (key, value) {
     if (key === expiresHeader) key = 'Expires';
-    if (key.indexOf('x-amz-') === 0) {
+    if (key.indexOf('x-amz-meta-') === 0) {
       key = key.toLowerCase();
     }
     queryParams[key] = value;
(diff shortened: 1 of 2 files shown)


LiuJoyceC:
:shipit:

PR #899, opened 2016-02-10
Add/cust retry delay

chrisradek:
Added 2 new options for customizing retry delays

Retry delays will default to using exponential backoff with a base of 30 ms (current behavior).
1) User can specify a `base` value for the exponential backoff to use, in ms.
2) User can specify their own backoff function that accepts the current retry count, and returns a retry delay in ms.

These options do not currently work with DynamoDB because we implemented custom retry logic for this service already. In the future we could allow this customization for DynamoDB as well.

/cc @LiuJoyceC @jeskew


LiuJoyceC:
:shipit:

LiuJoyceC:
:ship:

PR #914, opened 2016-02-29
reduce partBuffers copy if read as single buffer from the source stream

guymguym:


guymguym:
Hi
This gives a significant boost of upload performance (~300 MB/sec increase in my case) when the source stream is providing buffers of the same size of the part size, since it avoids the memory copy that Buffer.concat() does even in case of a single buffer.
Would be great if you want to merge.
Cheers

lib/s3/managed_upload.js
```diff
@@ -370,7 +370,8 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
     }
 
     if (self.partBufferLength >= self.partSize) {
-      var pbuf = Buffer.concat(self.partBuffers);
+      var pbuf = self.partBuffers.length === 1 ?
+        self.partBuffers[0] : Buffer.concat(self.partBuffers);
```
chrisradek: Can you also apply this to line 389  right under `if (self.isDoneChunking && !self.isDoneSending) {` as well? 
Otherwise, looks good for us to merge in!

guymguym:
@chrisradek thanks for the comment. updated.

chrisradek:
@guymguym 
Thanks for the contribution! Merging this PR, it'll be available via npm in the next release.

guymguym:
Great! :)

PR #923, opened 2016-03-03
Upgrades waitFor method on all supported services to use Waiters v2

### LINKED ISSUE #828: ECS Waiters
When will the ecs waiters be added? Like `tasksRunning`, etc.

LiuJoyceC:
This pull request will upgrade the waiters to use the v2 models. This will add support for waiters in the following services that were previously not supported in Waiters v1: ElastiCache, IAM, OpsWorks, CloudFormation, Kinesis, ECS, EMR, and ELB. In addition, waiters now support more states in the following services: EC2 and Redshift.

Fixes #828 

/cc @chrisradek


chrisradek:
Since we're pulling in `jmespath` now, we should be able to get rid of `jamespath` from the `AWS.util` namespace.

lib/resource_waiter.js
```diff
@@ -32,76 +61,72 @@ AWS.ResourceWaiter = inherit({
...
-          resp.error = null;
-          resp.data = {};
-          resp.request.removeAllListeners('extractData');
-        }
-      });
+  listeners: new AWS.SequentialExecutor().addNamedListeners(function(add) {
+    add('RETRY_CHECK', 'retry', function(resp) {
+      var waiter = resp.request._waiter;
+      if (resp.error && resp.error.code === 'ResourceNotReady') {
+        resp.error.retryDelay = (waiter.config.delay || 0) * 1000;
```
chrisradek: It might make sense to use a different default for the delay than 0, maybe 1. This is an edge-case because I think all our waiters have a delay specified, but we would potentially see throttling errors if a delay of 0 was used.

chenliu0831:
:+1:

chrisradek:
Ship it! This is great in gaining parity with the other SDKs.

PR #926, opened 2016-03-09
Profile assume role

dump247:
The implementation matches what is in botocore and my [PR for the Java SDK](https://github.com/aws/aws-sdk-java/pull/641).

I had to remove the `get` call from the constructor and the provider from the Config credentials load to prevent multiple role assumptions from happening.

### DIFF
--- a/lib/aws.js
+++ b/lib/aws.js
@@ -34,7 +34,7 @@ AWS.util.update(AWS.Config.prototype.keys, {
     new AWS.CredentialProviderChain([
       function () { return new AWS.EnvironmentCredentials('AWS'); },
       function () { return new AWS.EnvironmentCredentials('AMAZON'); },
-      function () { return new AWS.SharedIniFileCredentials(); }
+      function () { return new AWS.SharedIniFileCredentials({ disableAssumeRole: true }); }
     ]).resolve(function(err, creds) {
       if (!err) credentials = creds;
     });
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -40,6 +40,9 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
    *   the name of the profile to load.
    * @option options filename [String] ('~/.aws/credentials') the filename
    *   to use when loading credentials.
+   * @option options disableAssumeRole [Boolean] (false) True to disable
+   *   support for profiles that assume an IAM role. If true, and an assume
+   *   role profile is selected, an error is raised.
    */
   constructor: function SharedIniFileCredentials(options) {
     AWS.Credentials.call(this);
@@ -48,7 +51,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
 
     this.filename = options.filename;
     this.profile = options.profile || process.env.AWS_PROFILE || 'default';
-    this.get(function() {});
+    this.disableAssumeRole = !!options.disableAssumeRole;
   },
 
   /**
@@ -67,11 +70,21 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
     try {
       if (!this.filename) this.loadDefaultFilename();
       var creds = AWS.util.ini.parse(AWS.util.readFileSync(this.filename));
-      if (typeof creds[this.profile] === 'object') {
-        this.accessKeyId = creds[this.profile]['aws_access_key_id'];
-        this.secretAccessKey = creds[this.profile]['aws_secret_access_key'];
-        this.sessionToken = creds[this.profile]['aws_session_token'];
+      var profile = creds[this.profile];
+
+      if (typeof profile !== 'object') {
+        throw new Error('Profile ' + this.profile + ' not found in ' +
+                        this.filename);
       }
+
+      if (profile['role_arn']) {
+        this.loadRoleProfile(creds, profile, callback);
+        return;
+      }
+
+      this.accessKeyId = profile['aws_access_key_id'];
+      this.secretAccessKey = profile['aws_secret_access_key'];
+      this.sessionToken = profile['aws_session_token'];
 
       if (!this.accessKeyId || !this.secretAccessKey) {
         throw new Error('Credentials not set in ' + this.filename +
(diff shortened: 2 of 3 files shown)


temujin9:
@chrisradek @LiuJoyceC @jeskew Can we get some feedback on this? Lack of this code is blocking migration of Node code to our new AWS environments.

lib/aws.js
```diff
@@ -33,8 +33,7 @@ AWS.util.update(AWS.Config.prototype.keys, {
     var credentials = null;
     new AWS.CredentialProviderChain([
       function () { return new AWS.EnvironmentCredentials('AWS'); },
-      function () { return new AWS.EnvironmentCredentials('AMAZON'); },
-      function () { return new AWS.SharedIniFileCredentials(); }
```
chrisradek: I don't think we can safely remove the SharedIniFileCredentials from being invoked when the SDK is loaded.

While it's true that the defaultProviders will be called sequentially when a request is made and no valid/current credentials were loaded, there may be users that are performing logic on the credentials prior to making any requests. Removing this has the potential to break them.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -87,6 +95,70 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
   /**
    * @api private
    */
+  loadRoleProfile: function loadRoleProfile(creds, callback) {
+    var self = this;
+    var roleArn = creds[this.profile]['role_arn'];
```
chrisradek: It might be a good idea to store `creds[this.profile]` in a variable since we're calling it so frequently within the same function.

chrisradek:
@temujin9 
Overall, the PR looks like an interesting and useful feature.

One issue is that this implementation currently looks for the role to use in the ~/.aws/credentials file. Boto3 actually documents that this information is stored in the ~/.aws/config file. We don't currently support reading from the ~/.aws/config file in the JavaScript SDK, but I'll bring this up with the team to see what the other SDKs think about where to store this configuration.

lib/aws.js
dump247: You could leave it in place, but it results in duplicate role assumptions. The reason being that the "resolve" here expects to complete synchronously and the role assumption completes asynchronously. So this method ends up returning null for assume-role profiles and the role has to be assumed again later.

dump247:
Boto3 (via botocore) most definitely loads from _~/.aws/credentials_. It may also support loading from _~/.aws/config_, but I haven't tested that and don't have the ability to test it at the moment.

chrisradek:
@dump247 
I discussed this with the team, and we agree this makes sense to include with the ~/.aws/credentials file.

I'll finish reviewing this asap!

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -87,6 +95,70 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
...
+    var sourceProfileName = creds[this.profile]['source_profile'];
+
+    if (!sourceProfileName) {
+      throw new Error('source_profile is not set in ' + this.filename +
+                      ' using profile ' + this.profile);
+    }
+
+    var sourceProfile = creds[sourceProfileName];
+
+    if (typeof creds[this.profile] !== 'object') {
```
chrisradek: This check would be better if it was moved before we grabbed properties from the object. The error message would need to be changed a bit so it doesn't use `sourceProfileName`

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -87,6 +95,70 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
...
+    var roleParams = {
+      RoleArn: roleArn,
+      RoleSessionName: roleSessionName || 'aws-sdk-js-' + Date.now()
+    };
+
+    if (externalId) {
+      roleParams.ExternalId = externalId;
+    }
+
+    console.log('assuming role:', roleParams);
```
chrisradek: Please remove any `console.log` calls from the final code.

lib/credentials/shared_ini_file_credentials.js
dump247: Updated

lib/credentials/shared_ini_file_credentials.js
dump247: Oops! That was my debugging.

lib/credentials/shared_ini_file_credentials.js
dump247: This was a bug. The condition should have been checking `sourceProfile`. It has been fixed.

lib/aws.js
dump247: @chrisradek Is there anything you want me to change here? The `credentials` method doesn't work for assume role profiles (it will always return null in that case). A couple of things I can think of, if just removing the provider is unacceptable:

1. Add a configuration option to the provider to disable assume role. It doesn't work in this case anyway. This seems bad to me, but would result in the same behavior as before my change.
2. Separate the normal roles and the assume roles into different providers. I don't think this is a good idea since it requires parsing the creds file twice and separates the parsing code. Botocore does this, but only because the code was copy+pasted from the cli.
3. Somehow make this method complete asynchronously. I don't know if this is feasible or not. I haven't looked into it.

For options 2 & 3, there is the risk that you get unnecessary assume role API requests if the user doesn't use the default credentials configuration. Those unnecessary API requests could slow down SDK initialization a bit.

chrisradek: I don't think there's a way to implement option 3 in a way that solves the issue without requiring users to change their flow.

If we added a configuration option to the provider to disable assume role, that could work. Something like `disableAssumeRole` that would still throw an error if the `roleArn` was found for the profile. I'd suggest defaulting it to false and setting it to true when it is called here.

The argument could be made for another credential provider to handle this, but I also think that having the `SharedIniFileCredentials` object handle these credentials from the ini file makes sense.

The impact is pretty small, only potentially causing an extra `STS` request when the SDK is required for the first time.

dump247: Done

dump247:
@chrisradek Anything else?

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -87,6 +100,75 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
   /**
    * @api private
    */
+  loadRoleProfile: function loadRoleProfile(creds, callback) {
```
chrisradek: We could probably just pass the `roleProfile` instead of `creds` so we don't have to grab the `roleProfile` later. I don't think we'll have to worry about having the whole creds file in the future for this method.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -48,11 +51,11 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
 
     this.filename = options.filename;
     this.profile = options.profile || process.env.AWS_PROFILE || 'default';
-    this.get(function() {});
```
chrisradek: I thought removing `this.get` from the constructor was a safe bet, but there is a scenario that would break existing users with this change. Specifically, if a user is generating presigned url's synchronously, and are not running other operations (a valid use-case), their code would start to fail because the credentials won't have been loaded.

Unfortunately I think it's not safe to remove the extra `get` call currently, but we could in a major version bump in the future.

dump247: I made the change, but the problem here is that the standard provider chain will result in multiple role assumptions. The reason being that the constructor initiates a role assumption and then any api calls will start another when they get credentials (since the initial one has not completed). Is there any way to address that issue? Another constructor argument to disable the initialization `get`?

lib/credentials/shared_ini_file_credentials.js
dump247: Done. The method still requires the `creds` since the `source_profile` has to be retrieved.

chrisradek: Ah, I missed that `creds` was being used elsewhere, but thanks for making the change.

lib/credentials/shared_ini_file_credentials.js
chrisradek: Short of creating a separate credential provider, I don't think there's a way we can safely change this. That would however mean reading the same file twice, but also eliminates the constructor parameter to disable assume role.

dump247: I don't think I really understand the issue with removing the `get` in the constructor. Are you concerned about someone directly instantiating SharedIniFileCredentials and not calling `get`? It should get called when the provider chain is resolved right? In what case is `get` not called before retrieving the creds?

chrisradek: Nevermind, after further testing, turns out this is not an issue since we didn't remove the `SharedIniFileCredentials` from the initial `CredentialProviderChain`.

chrisradek:
Thanks @dump247 for being so responsive to suggestions! I'll run the regression tests one last time. I don't think any further changes are needed.

dump247:
@chrisradek I removed the `get` call from the constructor. I can certainly appreciate the level scrutiny required here.

dump247:
@chrisradek Anything else?

chrisradek:
@dump247 
I added some additional unit tests to #963. It includes all your final changes in this pull request so the tests and changes will be merged together.

chrisradek:
Changes merged in with #963.
Thanks @dump247  for the pull request!

PR #927, opened 2016-03-10
Add support for AWS_DEFAULT_PROFILE environment variable

dump247:
The other SDKs use AWS_DEFAULT_PROFILE and not AWS_PROFILE.


chrisradek:
@dump247 
Thanks for the pull request! At this point some SDKs support AWS_DEFAULT_PROFILE, and some don't. The CLI does, for example, but there AWS_DEFAULT_PROFILE takes precedence over AWS_PROFILE. Unfortunately doing the same here would break some users, but then we have another inconsistent behavior where in one tool, AWS_DEFAULT_PROFILE overrides AWS_PROFILE, and in other tools, the opposite is true.

I think we need to have a conversation with all the SDKs to decide a standard approach.

dump247:
OK

Issue #811, opened 2015-11-23
Multipart Managed Upload encoded Location


fyockm:
When using the Managed Uploader on a large file (>5M), the `Location` returned in the data of the callback is incorrect, with `%2F` instead of a `/`. Using the Managed Uploaded with a file smaller than 5M does not yield the same result- the Location is set as expected.

Here's some code to illustrate:
```
var key = filename;

s3.upload({
    Bucket: 'bucket',
    Key: req.user.id + '/' + key,
    Body: req
}, function(err, data) {
    if (err) {
        return res.status(err.statusCode).end(err.message);
    }

    console.log('data:', data);

    // Fix the S3 Multipart Upload bug
    data.Location = data.Location.replace('%2F', '/');

    res.set(data);
    res.end();
});
```

For the large file, we get the following output. Notice the `%2F` in the `Location`.
```
data: { Location: 'https://s3.amazonaws.com/bucket/55159b8b109ea8d1653eb5b1%2F1largeImage.jpg',
  Bucket: 'bucket',
  Key: '55159b8b109ea8d1653eb5b1/1largeImage.jpg',
  ETag: '"96c30c200150682f407619e5046505d8-2"' }
```

The same code run for a smaller image yields this output:
```
data: { Location: 'https://s3.amazonaws.com/bucket/55159b8b109ea8d1653eb5b1/1testImage.jpg' }
```

chrisradek:
@fyockm 
What are you doing with the `Location`? Both URLs should be valid and return the same object from S3. This wouldn't be the case if you were using another SDK function like `getObject`, but in that scenario you'd likely use the `Key` instead.

fyockm:
We are using `Location` to store the full URL for the object, so that it's accessible later by whatever method is required for the given client. That might be directly via an html `<img>` tag, or through `getObject`. In case of the latter, we have to parse out the Bucket and Key, which becomes more difficult with inconsistent URLs.

As you can see from my code, I can easily fix the encoding problem with the `Location` for a multipart file. 

1. I thought this could trip up some other dev in the future and wanted it to be documented
2. I believe the `Location` url should be consistent, regardless of filesize

To my understanding, the whole point of the `s3.upload` function is to make it easier on devs by hiding the underlying `getObject` vs `createMultipartUpload` craziness. Unexpected results like this only add to the frustration.

chrisradek:
@fyockm 
Would it help if the `data` returned by `s3.upload` always included `Bucket` and `Key`?

[...]

PR #944, opened 2016-03-25
Added 'Key' and 'Bucket' to data returned from s3.upload when single …

### LINKED ISSUE #918: bucket.upload() passes "key" vs "Key" to callback upon successful upload of simple vs multipart
minor issue, but calling `.upload()` returns slightly different results depending on if the file was large enough to trigger a multipart chunked upload.  in the **multipart** case, the result object has a key of _Key_ (not upper-K), whereas the simpler upload (smaller files) it is only _key_ (lower-k).  it can be worked around by, for example, using `result.Key || result.key`, but thought it might be worth posting.

LiuJoyceC:
…part upload is used. This is to be consistent with the data that is returned from s3.upload when multipart upload is used. Resolves #918
/cc @chrisradek


chrisradek:
:shipit:

LiuJoyceC:
This also resolves #811

PR #945, opened 2016-03-28
Copy-less s3 upload for nodejs - progressStream copied the data in memory for no good reason

guymguym:
For Stream Body - changed the progress to tap into the stream without copying using the 'data' event in addition to piping.
For non-Stream Body -  simply emit the progress event immediately since the data is already loaded in memory.

This reduces CPU consumption, and increases the throughput pushed by the client.
Tested to push more than 2000 MB/sec from a single node client process compared to less than 1000 MB/sec before.


guymguym:
Hey @chrisradek, would love to get your feedback on this fix. I know it's in a very common path so let me know if you see anything i could do to write it safer for existing code and make it easier to merge. Thanks!

chrisradek:
@guymguym 
Thanks for the PR! I'll take a look at it today and run our integration tests on it. Will especially be testing against older environments of nodejs that we're still supporting (0.8.x and 0.10.x)

chrisradek:
@guymguym 
Just wanted to give you an update. Some of our regression tests failed when I pulled down your changes, but I haven't had a chance to dig into why yet. I'll work on that some point this week.

Sorry for the delay!

guymguym:
Hey @chrisradek Sorry about the hassle.
Are these tests public that I can also run?
Is it simply to run the project tests with 0.8 and 0.10 (I have nvm so no problem)?
I'm pretty sure it will easy to resolve.
Thanks

guymguym:
I ran the project's `npm install && npm test`on all the nvm versions I have below, and it worked for all besides 0.8 where is failed during install with errors that seem unrelated to the change -

```
$ nvm ls
->       v0.8.28
       v0.10.44
       v4.4.2
       v5.10.0

$ npm install
npm http GET https://registry.npmjs.org/sax/1.1.5
npm http GET https://registry.npmjs.org/xml2js/0.4.15
npm http GET https://registry.npmjs.org/repl.history
npm http GET https://registry.npmjs.org/xmlbuilder/2.6.2
npm http GET https://registry.npmjs.org/semver
npm http GET https://registry.npmjs.org/coffee-script/1.6.3
npm http GET https://registry.npmjs.org/eslint
npm http GET https://registry.npmjs.org/cucumber
npm http GET https://registry.npmjs.org/coffeeify
npm http GET https://registry.npmjs.org/browserify
npm http GET https://registry.npmjs.org/mocha
npm http GET https://registry.npmjs.org/uglify-js
npm http GET https://registry.npmjs.org/chai
npm http GET https://registry.npmjs.org/istanbul
npm http GET https://registry.npmjs.org/coveralls
npm http GET https://registry.npmjs.org/insert-module-globals
npm http 200 https://registry.npmjs.org/semver
npm http 200 https://registry.npmjs.org/sax/1.1.5
npm http GET https://registry.npmjs.org/sax/-/sax-1.1.5.tgz
npm http 200 https://registry.npmjs.org/repl.history
npm http 200 https://registry.npmjs.org/xml2js/0.4.15
npm http GET https://registry.npmjs.org/xml2js/-/xml2js-0.4.15.tgz
npm http 200 https://registry.npmjs.org/xmlbuilder/2.6.2
npm http GET https://registry.npmjs.org/xmlbuilder/-/xmlbuilder-2.6.2.tgz
npm http 200 https://registry.npmjs.org/eslint
npm http 200 https://registry.npmjs.org/sax/-/sax-1.1.5.tgz
npm http 200 https://registry.npmjs.org/cucumber
npm http 200 https://registry.npmjs.org/mocha
npm ERR! registry error parsing json
npm ERR! SyntaxError: Unexpected token 
npm ERR! ���o�6��A{�~��m��

.....

npm ERR! ƅ���1��FīJ�i#���"����d��l��`?\��;6Qy&%&{���r�݂Ph7�������w�R�����7�f�&����Y�b-�5�P�
���w����BH�e>��xAW���:��$��ݼ�m5�oK8���.+�n����]-������v��]�k��e���]����Kn�0��bt_-���bS!����n����-��jX�GC~[�y^3_���.n�n
9z29D�U|~�ir��R�T�k��Ea��jW������Z	�8��:�<�T��e:B���yM���vl�T�L��9��.��y��MH{�#M1�"ãUr;[�AF��]�?�R�>��G�&���p�1T�%�Aj�V
          �^3K�eߒ�;p](\g!H���{\`�A�A�i����l��$x+v�
                                                  �i���;�s~�}�^���!2	�B��=@��5v�NP�$�H�u�xk#�z�@�9r$�ͮ8�$�Ӱ�@ei��#0���A;�����I����6�#
                                                                                                                                       ��#�A���;E����1G�C�>,��ިͽ�˃JSYP��jy�[z�2n��]����~�uc�A����"�3U���.��T-��ݯ/l%�K����@n�b�*A�K&�p����L������L d۩j���de���/�n�!�����w��9M��[�1�L�
                                                                                                �>�[옪v�+���r�_}D�[lA�{��T�RN|�T+Y�����j�����4<
                                                                                                                                               ���S���゛���e<�So8��|�FI��6��e����QP�9AS�����Sz{��~v��<$L�a�����������F���mK�㲤��ȧׁ#.��j_��mǧ��4�6�J
�������N M�������>��!
npm ERR!     at Object.parse (native)
npm ERR!     at RegClient.<anonymous> (~/.nvm/v0.8.28/lib/node_modules/npm/node_modules/npm-registry-client/lib/request.js:238:23)
npm ERR!     at Request.init.self.callback (~/.nvm/v0.8.28/lib/node_modules/npm/node_modules/request/index.js:148:22)
npm ERR!     at Request.EventEmitter.emit (events.js:99:17)
npm ERR!     at Request.onResponse (~/.nvm/v0.8.28/lib/node_modules/npm/node_modules/request/index.js:876:14)
npm ERR!     at Request.EventEmitter.emit (events.js:126:20)
npm ERR!     at IncomingMessage.Request.onResponse.buffer (~/.nvm/v0.8.28/lib/node_modules/npm/node_modules/request/index.js:827:12)
npm ERR!     at IncomingMessage.EventEmitter.emit (events.js:126:20)
npm ERR!     at IncomingMessage._emitEnd (http.js:367:10)
npm ERR!     at HTTPParser.parserOnMessageComplete [as onMessageComplete] (http.js:149:23)
npm ERR! If you need help, you may report this log at:
npm ERR!     <http://github.com/isaacs/npm/issues>
npm ERR! or email it to:
npm ERR!     <npm-@googlegroups.com>

npm ERR! System Darwin 15.4.0
npm ERR! command "~/.nvm/v0.8.28/bin/node" "~/.nvm/v0.8.28/bin/npm" "install"
npm ERR! cwd ~/aws-sdk-js
npm ERR! node -v v0.8.28
npm ERR! npm -v 1.2.30
npm ERR! type unexpected_token
npm http 200 https://registry.npmjs.org/coffee-script/1.6.3
npm http GET https://registry.npmjs.org/coffee-script/-/coffee-script-1.6.3.tgz
npm http 200 https://registry.npmjs.org/chai
npm http 200 https://registry.npmjs.org/coffeeify
npm http 200 https://registry.npmjs.org/xml2js/-/xml2js-0.4.15.tgz
npm http 200 https://registry.npmjs.org/xmlbuilder/-/xmlbuilder-2.6.2.tgz
npm http 200 https://registry.npmjs.org/coveralls
npm http 200 https://registry.npmjs.org/coffee-script/-/coffee-script-1.6.3.tgz
npm http 200 https://registry.npmjs.org/uglify-js
npm http 200 https://registry.npmjs.org/istanbul
npm http 200 https://registry.npmjs.org/browserify
npm http 200 https://registry.npmjs.org/insert-module-globals
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     ~/aws-sdk-js/npm-debug.log
npm ERR! not ok code 0
```

chrisradek:
@guymguym 
Our regression tests are cucumber tests, specifically the `progress` tests are failing:
https://github.com/aws/aws-sdk-js/blob/master/features/s3/objects.feature#L164
https://github.com/aws/aws-sdk-js/blob/master/features/s3/step_definitions/objects.js#L160

Sorry I haven't dug deeper into whether this is an actual issue or not yet, but it is on the list!

guymguym:
@chrisradek no worries, at least the progress test that you pointed out is quite easy to figure out.

This test asserts that once a putObject of 2 MB buffer is used, there will be 2 or more progress events, where the first should have a loaded count of at least 10.

Before my change every buffer was converted to a stream, which had a highWaterMark of 16KB which means that the event was issued many times, certainly more than once.

In my change, I simply changed to avoid such a conversion from memory buffer to stream that was hurting performance, and simply write the entire buffer at one call (`stream.end(body)`) and emitted a single progress event that the entire data is loaded.

It is quite simple to imitate the asserted behavior, but the question is if it makes sense to assert that a buffer should provide multiple progress events? 

Would be great to know what you think.

Thanks!

guymguym:
I tried a fix that will emit a progress event once the output 'drain' event is received.
This should fix that case at least.
But I couldn't get to configure the cucumber tests for my env... any instructions for that?

guymguym:
BTW the travis failures seem unrelated to my changes - not sure what's that jam:
```
$ cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
/home/travis/build/aws/aws-sdk-js/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

guymguym:
@chrisradek 
The build issue was transient and not related.
I think I fixed the issue with the cucumber test that you pointed out - can you help to rerun?
Thanks!

chrisradek:
@guymguym 
Yes, I will rerun it, thank you!

guymguym:
@chrisradek 
I want to help push this, but not sure how to run the cucumber tests myself...
I tried naively using a `configuration` file same as `configuration.sample` but get this error:
```
@acm
Feature: 
  I want to use AWS Certificate Manager
```

chrisradek:
@guymguym 
You can modify the configuration.sample file (https://github.com/aws/aws-sdk-js/blob/master/configuration.sample) to include just the region and s3 properties.
`Note: Please update the S3 bucket used.`
You also need to npm install cucumber at version 0.5.3. (I think some newer versions don't work with our suite).

Then, to run the tests, you type:
`cucumber.js -t @s3` 
to run just the s3 tests. You will not want to run all the tests, because these will make actual requests and possibly create resources that you could be charged for.

Thanks for pinging me here. I should have more time this week to review it. 

/cc @LiuJoyceC  as well.

guymguym:
thanks, all passed:
```
39 scenarios (39 passed)
242 steps (242 passed)
```

olivierlesnicki:
@guymguym well needed! good job on this one

guymguym:
Hey @chrisradek 

Do you see any gaps?

I had a minor dilema about the way buffer/string data emits progress events:
* For best performance it is sent with `stream.end(body)` since the http request stream can take the entire memory buffer and process it as quickly as it can.
* However to give progress events as required by the tests I did something fuzzy - I emitted the progress event twice with `loadedBytes = totalBytes` - once just before pushing the buffer to the stream, and another time once the stream drains. This does pass the tests, but might seem a bit weird to anyone observing this from the client perspective.
* If you prefer to sacrifice the simplicity (and possibly some performance) for the sake of finer progress events then I can suggest the code change below that will send in bulks. Notice that it will behave exactly the same as before for body of length smaller than bulk size.

Personally I would probably prefer to keep it simple, but I can certainly understand if you think the finer events are better for clients, and it depends on how common it is to send very large request body.

Would be great to know how you think about it.
Thanks,
Guy


```
var MAX_BULK_SIZE = 8 * 1024 * 1024;
var remaining_body = body.slice();
send_body_in_bulks();

function send_body_in_bulks() {
  if (!remaining_body.length) {
    loadedBytes = totalBytes;
    stream.emit('sendProgress', {
      loaded: loadedBytes,
      total: totalBytes
    });
    return;
  } 

  // slice and send a bulk
  var bulk = remaining_body.slice(0, MAX_BULK_SIZE);
  remaining_body = remaining_body.slice(bulk.length);
  var completed = stream.write(bulk);

  // report progress
  loadedBytes += bulk.length;
  stream.emit('sendProgress', {
    loaded: loadedBytes,
    total: totalBytes
  });

  // if the write completed, we can immediately write more,
  // otherwise wait for stream to drain it's buffered data
  if (completed) {
      setImmediate(send_body_in_bulks);
  } else {
      stream.once('drain', send_body_in_bulks);
  }
}
```

lib/http/node.js
```diff
@@ -76,15 +75,40 @@ AWS.NodeHttpClient = AWS.util.inherit({
...
-      if (!(body instanceof Stream)) body = AWS.util.buffer.toStream(body);
-      body.pipe(this.progressStream(stream, httpRequest));
-    }
+    var totalBytes = parseInt(httpRequest.headers['Content-Length'], 10);
+    var loadedBytes = 0;
 
     if (body instanceof Stream) {
+      // for progress support of streaming content
+      // tap the data event of the stream in addition to piping
+      body.on('data', function(chunk) {
```
chrisradek: I think tapping into the `data` event on a stream could cause some unintended side-effects. In node.js 0.10.x, binding a `data` listener on a stream will cause the stream to emit 'data' events as fast as it can, ignoring the back-pressure that's automatically in place when using `pipe`. If the writable stream is slow, then this could cause loss of data.

I don't think this is an issue in versions of node.js >= 0.12.x, and some simple testing seems to confirm that. However, we need to work with node.js 0.10.x as well.

The current method creates a new `writable` stream that also gets piped into in order to emit the 'sendProgress' events. I know it'll require refactoring your logic but that path seems safer across all our supported versions of node.

lib/http/node.js
```diff
@@ -76,15 +75,40 @@ AWS.NodeHttpClient = AWS.util.inherit({
...
+      // so for performance it's best to send it as a whole by calling stream.end(body),
+      // but since callers expect some progression, we emit one progress event before sending,
+      // and another once the stream drains.
+      // the use of totalBytes is important over body.length for strings where length is char length and not byte length.
+      loadedBytes = totalBytes;
+      stream.emit('sendProgress', {
+        loaded: loadedBytes,
+        total: totalBytes
+      });
+      stream.once('drain', function() {
```
chrisradek: I don't think it's important to emit `sendProgress` twice. Our docs simply state that the `httpUploadProgress` event will be triggered when data is uploaded. Emitting once `drain` is called is probably more accurate, since that indicates that the data has actually been uploaded.

I suspect that buffers/strings were converted to a stream so that the user had more fine-grained info on how much data was sent. However, I doubt it's worth the performance trade-off.

guymguym: Great so I can fix the cucumber test to expect > 0 events instead of > 1?

lib/http/node.js
guymguym: Very good point. Thanks for taking the time to review in depth.

In one of my previous tests I saw that the pipe to the writable progressStream caused buffers to be copied. But now I'm not sure about it completely, as it might have mixed with the buffer path in the code in my runs. 

Anyhow, I think to change the path of stream-body progress to use Transform stream to emit the events in the familiar pattern `body.pipe(progressStream).pipe(stream)`. Transform stream will push the same buffer so will surely avoid copies. sounds right?

chrisradek: Using a transform stream to emit the events sounds like a great idea!

lib/http/node.js
chrisradek: Yes, I think that's a more accurate test.

guymguym:
@chrisradek 
See the fixes according to our discussion.
I tested with `npm test` and `cucumber` with both node v4.4.3 and v0.10.

lib/http/node.js
```diff
@@ -76,17 +76,28 @@ AWS.NodeHttpClient = AWS.util.inherit({
...
-      if (!(body instanceof Stream)) body = AWS.util.buffer.toStream(body);
-      body.pipe(this.progressStream(stream, httpRequest));
-    }
+    var totalBytes = parseInt(httpRequest.headers['Content-Length'], 10);
 
     if (body instanceof Stream) {
-      body.pipe(stream);
+      // For progress support of streaming content -
+      // pipe the data through a transform stream to emit 'sendProgress' events
+      body.pipe(this.progressStream(stream, totalBytes)).pipe(stream);
```
chrisradek: Just one last change! Since we still have users running on node 0.8, we should only pipe into the progress stream if `typeof TransformStream !== undefined`.

guymguym: Oh sure, so there is no way to provide stream progress for 0.8?

guymguym: BTW I remembered correctly that using multiple writable streams behaves weird in nodejs v4 - I just submitted this issue https://github.com/nodejs/node/issues/6491. 
It was actually only fixed on nodejs v5.11 but I guess they will want to pull back to v4 for LTS.
Anyway that's a good reason to use Transform instead.

guymguym:
BTW I remembered correctly that using multiple writable streams behaves weird in nodejs v4 - I just submitted this issue nodejs/node#6491. 
It was actually only fixed on nodejs v5.11 but I guess they will want to pull back to v4 for LTS.
Anyway that's a good reason to use Transform instead.

chrisradek:
@guymguym 
Looks good! Thanks for all the work you put into this. Merging this in now, and it will be available via npm with the next release!

PR #946, opened 2016-03-28
Correctly recognize S3 errors with 200 status code for completeMultiPartUpload, copyObject, and uploadPartCopy

LiuJoyceC:
Previously, errors returned from S3 when the response had a status code of 200 were only being recognized in completeMultiPartUpload, but now they will be recognized for copyObject and uploadPartCopy.


chrisradek:
:shipit:

PR #953, opened 2016-04-05
Updates AWS.APIGateway.getSdk to return body as a buffer

### LINKED ISSUE #815: APIGateway getSdk data.body is not a buffer
Hi

I am using the `apigateway.getSdk` method to download the zip file of my SDK. This way I can integrate that with my gulpfile to make sure I always have the latest version.

The documentation can be found [here](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html#getSdk-property).

It states that the callback returns a `data` object with the property `body`, which is a `String`, `Buffer`, etc. So I tried that call and write the `data.body` directly as zip file to my filesystem.

```js
apigateway.getSdk(params, function (err, data) {
    if(err) {
        // handle error
        return;
    }

    fs.writeFileSync('apigateway.zip', data.body);
});
```

It didn't work as turned out, `data.body` is not a `Buffer`, but a `string`. So I tried to convert it into a buffer with `new Buffer(data.body)` and tried every possible encoding argument there is. Nothing worked for me.

But, what did work was the following.

```js
var request = apigateway.getSdk(params, function (err, data) {
    if(err) {
        // handle error
        return;
    }

    fs.writeFileSync('apigateway.zip', request.response.httpResponse.body);
});
```

So it turns out, the property in the `httpResponse` is a `Buffer` which can be written directly as zip file.

Is this an issue in the library or should I convert the `data.body` property body with a different function?

Kind regards,
Sam

chrisradek:
instead of a utf-8 string

fixes #815 
/cc @LiuJoyceC

### DIFF
--- a/lib/services/apigateway.js
+++ b/lib/services/apigateway.js
@@ -16,6 +16,19 @@ AWS.util.update(AWS.APIGateway.prototype, {
    */
   setupRequestListeners: function setupRequestListeners(request) {
     request.addListener('build', this.setAcceptHeader);
+    if (request.operation === 'getSdk') {
+      request.addListener('extractData', this.useRawPayload);
+    }
+  },
+
+  useRawPayload: function useRawPayload(resp) {
+    var req = resp.request;
+    var operation = req.operation;
+    var rules = req.service.api.operations[operation].output || {};
+    if (rules.payload) {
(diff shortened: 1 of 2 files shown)


LiuJoyceC:
:sheep: :it:

PR #964, opened 2016-04-14
Correction and clarification for CognitoIdentityCredentials documentation

### LINKED ISSUE #912: Improve documentation of AWS.CognitoIdentityCredentials(params)
The documentation of the [AWS.CognitoIdentityCredentials(params)](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CognitoIdentityCredentials.html#constructor-property) constructor can be improved:

- In the example code, the location of the inline comments is inconsistent. Sometimes, they are at the end of a line, sometimes on a new line making it confusing to which key they correspond.
- The `LoginId` key in the example code is not documented in the `Parameters` section.
- It is not stated which parameter keys are mandatory and which are optional.
- There are three different documentations of the `params (Object)` that have different keys, none of which maps to the example code. There is some overlap between them (the `Logins` key appear twice, but is only explained in detail once), but it is unclear why there are three different documentations and why one would choose one set of params instead of another one.

LiuJoyceC:
This adds clarification to the `CognitoIdentityCredentials` documentation, and adds a correction to state that either `IdentityId` or `IdentityPoolId` (which uses `AWS.CognitoIdentity.getId()` to obtain `IdentityId`) is required, regardless if `AWS.CognitoIdentity.getCredentialsForIdentity()` or `AWS.STS.assumeRoleWithWebIdentity()` is used to obtain credentials.
Resolves #912 
/cc @chrisradek

### DIFF
--- a/lib/credentials/cognito_identity_credentials.js
+++ b/lib/credentials/cognito_identity_credentials.js
@@ -69,28 +69,29 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
    * Creates a new credentials object.
    * @example Creating a new credentials object
    *   AWS.config.credentials = new AWS.CognitoIdentityCredentials({
-   *     
+   *
    *     // either IdentityPoolId or IdentityId is required
    *     IdentityPoolId: 'us-east-1:1699ebc0-7900-4099-b910-2df94f52a030',
    *     IdentityId: 'us-east-1:128d0a74-c82f-4553-916d-90053e4a8b0f'
-   *     
+   *
    *     // optional, only necessary when the identity pool is not configured
    *     // to use IAM roles in the Amazon Cognito Console
    *     RoleArn: 'arn:aws:iam::1234567890:role/MYAPP-CognitoIdentity',
-   *     
+   *
    *     // optional tokens, used for authenticated login
    *     Logins: {
    *       'graph.facebook.com': 'FBTOKEN',
    *       'www.amazon.com': 'AMAZONTOKEN',
    *       'accounts.google.com': 'GOOGLETOKEN'
    *     },
-   *     
+   *
    *     // optional name, defaults to web-identity
    *     RoleSessionName: 'web',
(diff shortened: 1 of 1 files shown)


lib/credentials/cognito_identity_credentials.js
```diff
@@ -66,27 +67,36 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
...
    *     IdentityPoolId: 'us-east-1:1699ebc0-7900-4099-b910-2df94f52a030',
+   *     IdentityId: 'us-east-1:128d0a74-c82f-4553-916d-90053e4a8b0f'
+   *
    *     // optional, only necessary when the identity pool is not configured
    *     // to use IAM roles in the Amazon Cognito Console
    *     RoleArn: 'arn:aws:iam::1234567890:role/MYAPP-CognitoIdentity',
-   *     Logins: { // optional tokens, used for authenticated login
+   *
+   *     // optional tokens, used for authenticated login
+   *     Logins: {
```
chrisradek: It might be good to expand on this list, since the parameters are being removed from the docs.

lib/credentials/cognito_identity_credentials.js
```diff
@@ -66,27 +67,36 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
...
+   *     RoleSessionName: 'web',
+   *
    *     // optional, only necessary when application runs in a browser
-   *     // and multiple users are signed in at once
+   *     // and multiple users are signed in at once, used for caching
    *     LoginId: 'example@gmail.com'
+   *
    *   });
-   * @see AWS.STS.assumeRoleWithWebIdentity
+   * @see AWS.CognitoIdentity.getId
```
chrisradek: It might also be good to make it more obvious when to use the parameters from which operation.

chrisradek:
:shipit:

PR #970, opened 2016-04-19
Adds accelerate as a subresource to s3 v2 signer

LiuJoyceC:
For new accelerate operations, accelerate is added as a subresource for v2 signing.

/cc: @chrisradek


chrisradek:
:shipit:

PR #976, opened 2016-04-25
Uses POST instead of GET for CloudSearch Domain search operation

### LINKED ISSUE #614: Handle CloudSearch Query Limit
When making a request to a CloudSearchDomain with the SDK, if the query is too long, the SDK returns an `UnknownError` with code 413.  To overcome the error, we had to patch the `request` object to convert to a *POST*.

```javascript
function awsRequestToPostIfHuge(request) {
  request.on('build', function() {
    var http = request.httpRequest
    // convert queries to POST to avoid length restrictions
    var path = http.path.split('?')
    http.method = 'POST'
    http.path = path[0]
    http.body = path[1]
    http.headers['Content-Length'] = http.body.length
    http.headers['Content-Type'] = 'application/x-www-form-urlencoded'
  })
}
// Usage
var request = searchDomain.search(params)
awsRequestToPostIfHuge(request)
request.send(function callback() { /* do stuff */ })
```

It seems like this should be handle transparently by the SDK.  You could use *POST* for all requests or optionally if the `request.httpRequest.path.length` is too large.

LiuJoyceC:
Converts search requests to CloudSearch Domain to use POST instead of GET to lift restriction on query length. Resolves #614

### DIFF
--- a/test/services/cloudsearchdomain.spec.coffee
+++ b/test/services/cloudsearchdomain.spec.coffee
@@ -49,8 +49,16 @@ describe 'AWS.CloudSearchDomain', ->
       params = { query: 'foo' }
       req = build('search', params)
       expect(req.headers).not.to.have.property('Authorization')
-    
+
     it 'signs request if credentials are provided', ->
       params = { query: 'foo' }
       req = build('search', params)
       expect(req.headers).to.have.property('Authorization')
+
+    it 'converts the GET request to POST for search operation', ->
+      params = { query: 'food' }
+      req = build('search', params)
+      expect(req.method).to.equal('POST')
+      expect(req.path.indexOf('?')).to.equal(-1)
+      expect(typeof req.body).to.equal('string')
+      expect(req.headers['Content-Length']).to.equal(req.body.length)
--- a/test/services/cloudsearchdomain.spec.coffee
+++ b/test/services/cloudsearchdomain.spec.coffee
@@ -56,9 +56,16 @@ describe 'AWS.CloudSearchDomain', ->
       expect(req.headers).to.have.property('Authorization')
 
     it 'converts the GET request to POST for search operation', ->
-      params = { query: 'food' }
+      params = { query: 'foo' }
       req = build('search', params)
       expect(req.method).to.equal('POST')
       expect(req.path.indexOf('?')).to.equal(-1)
       expect(typeof req.body).to.equal('string')
       expect(req.headers['Content-Length']).to.equal(req.body.length)
+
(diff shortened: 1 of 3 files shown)


test/signers/v4.spec.coffee
```diff
@@ -117,7 +117,10 @@ describe 'AWS.Signers.V4', ->
 
   describe 'canonicalString', ->
     it 'sorts the search string', ->
-      req = new AWS.CloudSearchDomain({endpoint: 'host.domain.com'}).search({query: 'foo', cursor: 'initial', queryOptions: '{}'}).build()
+      req = new AWS.CloudSearchDomain({endpoint: 'host.domain.com'})
+        .search({query: 'foo', cursor: 'initial', queryOptions: '{}'})
+        .removeListener('build', AWS.CloudSearchDomain.prototype.convertGetToPost)
```
chrisradek: Why is this listener being removed in the test?

test/services/cloudsearchdomain.spec.coffee
```diff
@@ -49,8 +49,16 @@ describe 'AWS.CloudSearchDomain', ->
...
       req = build('search', params)
       expect(req.headers).not.to.have.property('Authorization')
-    
+
     it 'signs request if credentials are provided', ->
       params = { query: 'foo' }
       req = build('search', params)
       expect(req.headers).to.have.property('Authorization')
+
+    it 'converts the GET request to POST for search operation', ->
```
chrisradek: We should probably have the opposite test case as well, where we ensure an operation that is not `search` is still using a `GET` http method.

test/signers/v4.spec.coffee
LiuJoyceC: This test is testing whether the query string is alphabetically sorted by its field names. The listener that I'm removing here previously did not exist, as it was added as part of this PR, so I'm not removing something that used to be there. Since the listener I added in this PR changes the GET request to a POST, there is no longer a query string, and the test cannot check if it's alphabetically sorted. Therefore, I removed the listener so that just for the purpose of this test, the request is still a GET and has a query string. The other two options I have if I don't remove this listener is to switch the test to use a different operation (I could use `suggest` instead of `search`) or to remove this test altogether.

test/services/cloudsearchdomain.spec.coffee
LiuJoyceC: I will add a test for `suggest`, as this is the only operation that uses a `GET`.

chrisradek:
:shipit:

PR #983, opened 2016-05-03
Fixes DynamoDB response data for CRC32CheckFailed error

### LINKED ISSUE #981: CRC32 check failed, marked as retryable but actually request will not be repeated
```
const dynamodb = new AWS.DynamoDB({
  accessKeyId: '...',
  secretAccessKey: '...',
  region: '...'
})

dynamodb.listTables({}, function(err, data) {
  if (err) {
    console.log(err)
  } else {
    console.log(data)
  }
})
```
Will produce this error:
```
{ 
  [CRC32CheckFailed: CRC32 integrity check failed]
  message: 'CRC32 integrity check failed',
  code: 'CRC32CheckFailed',
  retryable: true,
  time: Sun May 01 2016 23:20:52 GMT+0300 (EEST),
  statusCode: 200 
}
```
In the first place why `CRC32CheckFailed` happens at all? Second, as far as i know `maxRetries` defaults to `10`, when `CRC32CheckFailed` happens it will return `statusCode` equal to `200` with `retryable: true`, but i don't see any repeated requests. Shouldn't request be repeated?

Third, most important, even `if(err)` resolve to `true`, second parameter, `data` is also defined and actually it will return good results. How is that possible? That should never happen if `err` exists

LiuJoyceC:
Fixes bug in which response data is not null when DynamoDB returns a CRC32CheckFailed error. Response data will now return null when error occurs. Fixes #981 
/cc @chrisradek

### DIFF
--- a/lib/services/dynamodb.js
+++ b/lib/services/dynamodb.js
@@ -18,6 +18,7 @@ AWS.util.update(AWS.DynamoDB.prototype, {
    */
   checkCrc32: function checkCrc32(resp) {
     if (!resp.httpResponse.streaming && !resp.request.service.crc32IsValid(resp)) {
+      resp.data = null;
       resp.error = AWS.util.error(new Error(), {
         code: 'CRC32CheckFailed',
         message: 'CRC32 integrity check failed',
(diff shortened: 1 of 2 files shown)


chrisradek:
:shipit:

PR #985, opened 2016-05-04
Fixes multiple bugs in scripts/console

LiuJoyceC:
Fixes a bug in which .send() is called twice on a request whenever a callback is supplied to a service call, causing unintended retry behavior. Also fixes a bug in which manually calling .send() on a Request instance would not properly write to the data, error, request, and response variables in the REPL context and also would not show the error stack trace if an unsuccessful response was received. Also adds an optional flag 'autoSend', which can be set to false to turn off the default behavior of sending requests when callbacks are not supplied, allowing inspection of the Request instance before it gets sent.

/cc @chrisradek

### DIFF
--- a/scripts/console
+++ b/scripts/console
@@ -14,26 +14,59 @@ function customEval(cmd, context, filename, callback) {
       callback(err, null);
       return;
     }
+
+    function consoleDataExtraction(resp) {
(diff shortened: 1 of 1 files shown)


chrisradek:
Ship it!

PR #988, opened 2016-05-11
Adds a new S3 configuration options: s3DisableBodySigning

chrisradek:
s3DisableBodySigning turns off signing of the body for S3 operations when Sigv4 and https are being used. The Content-MD5 will still be sent. Defaults to true.
This should greatly help performance when uploading to Sigv4 regions from within the browser.

Related to #942

Fixes S3 SigV4 presigned URLs when supplying a body.

/cc @LiuJoyceC

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -77,6 +77,10 @@ require('./credentials/credential_provider_chain');
  *   @return [Boolean] whether the provided endpoint addresses an individual
  *     bucket (false if it addresses the root API endpoint).
  *
+ * @!attribute s3DisableBodySigning
+ *   @return [Boolean] whether to disable S3 body signing when using signature version `v4`.
+ *     Defaults to `true`.
+ *
  * @!attribute useAccelerateEndpoint
  *   @note This configuration option is only compatible with S3 while accessing
  *     dns-compatible buckets.
@@ -203,6 +207,9 @@ AWS.Config = AWS.util.inherit({
    *   addresses an individual bucket (false if it addresses the root API
    *   endpoint). Note that setting this configuration option requires an
    *   `endpoint` to be provided explicitly to the service constructor.
+   * @option options s3DisableBodySigning [Boolean] whether S3 body signing
+   *   should be disabled when using signature version `v4`. Defaults to `true`.
+   *
    * @option options retryDelayOptions [map] A set of options to configure
    *   the retry delay on retryable errors. Currently supported options are:
    *
@@ -457,6 +464,7 @@ AWS.Config = AWS.util.inherit({
     sslEnabled: true,
     s3ForcePathStyle: false,
     s3BucketEndpoint: false,
+    s3DisableBodySigning: true,
     computeChecksums: true,
     convertResponseTypes: true,
     correctClockSkew: false,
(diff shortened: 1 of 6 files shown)


lib/config.js
```diff
@@ -457,6 +464,7 @@ AWS.Config = AWS.util.inherit({
     sslEnabled: true,
     s3ForcePathStyle: false,
     s3BucketEndpoint: false,
+    s3DisableBodySigning: true,
```
LiuJoyceC: It might be good to default `s3DisableBodySigning` to true only when `sslEnabled` is true. If `sslEnabled` is explicitly set to false, then `s3DisableBodySigning` should default to false (and the documentation comments above should note this). Even though `shouldDisableBodySigning()` in the s3.js file checks that the protocol is https, this makes it clearer (and may make debugging easier). And if `{sslEnabled: false, s3DisableBodySigning: true}` is explicitly set, I would even add a validation listener to return an error when a request is being validated, to make it clearer that this combination is not allowed.

chrisradek: `sslEnabled` is also defaulted to `true`. The user may have `sslEnabled` set globally or at the client level, and depending on the order that properties get merged in (which is not guaranteed when using JavaScript objects), we could incorrectly set the 'default' `s3DisableBodySigning`. For example, if the user globally set `sslEnabled` to false, and set it to true on the s3 client, if the `s3DisableBodySigning` property was merged before `sslEnabled`, it would default to `false` incorrectly. I think this complicates the behavior unnecessarily, and could lead to trickier behavior to troubleshoot.

I can update the documentation to specify that this config is also only applied when using https. Besides `sslEnabled`, users can also specify their own endpoint that uses something other than `https`, so it'd be good to document this config for that scenario as well.

I am hesitant to throw an error if both settings are explicitly set. I think documentation is the better route to go, since we can reliably handle the case where `sslEnabled: false` and `s3DisableBodySigning: true`, and the behavior is identical to what we're currently doing.

lib/services/s3.js
```diff
@@ -436,13 +459,41 @@ AWS.util.update(AWS.S3.prototype, {
...
+   * @api private
+   * @param request
+   */
+  disableBodySigning: function disableBodySigning(request) {
+    // Only file streams are supported (content-md5 still calculated)
+    request.service.validateStreamBody(request.httpRequest.body || '');
+    var headers = request.httpRequest.headers;
+    // Add the header to anything that isn't a presigned url, unless that presigned url had a body defined
+    if (headers.hasOwnProperty('X-Amz-Content-Sha256') || !headers.hasOwnProperty('presigned-expires')) {
+      headers['X-Amz-Content-Sha256'] = 'UNSIGNED-PAYLOAD';
```
LiuJoyceC: Is there a reason to override the `X-Amz-Content-Sha256` header if it already exists? If the reason to disable body signing is to prevent expensive computations that could slow down the browser, then what is the benefit of overriding it when it has already been computed? The only place in the SDK that adds the `X-Amz-Content-Sha256` header is in the `AWS.EventListeners.Core.COMPUTE_SHA256` listener, which is already removed if `shouldDisableBodySigning()` returns true.

If a customer decides to compute the Sha256 on their own and adds it to the header, it would seem like surprising behavior that their own header gets overridden, especially if they did not explicitly set `s3DisableBodySigning` to false and did not realize that the SDK defaults that option to true.

chrisradek: I think you're probably right that I don't need to check for the header. I'll remove it and verify my test cases with presigned urls still work.

chrisradek:
Updated the PR based on comments.

LiuJoyceC:
🐑 🇮🇹

PR #989, opened 2016-05-11
Changes default signer when using the S3 client to signature version 4.

chrisradek:
Presigned urls will continue to default to signature version 2.
Fixes issue where specifying 'v2' as the signatureVersion resulted in errors when sending requests. SDK now silently translates 'v2' to 's3'.

This pull request can not be merged until #988 is merged, and region-redirection support is added for sigv4.

/cc @LiuJoyceC

### DIFF
--- a/lib/event_listeners.js
+++ b/lib/event_listeners.js
@@ -92,7 +92,7 @@ AWS.EventListeners = {
     addAsync('COMPUTE_SHA256', 'afterBuild', function COMPUTE_SHA256(req, done) {
       req.haltHandlersOnError();
       if (!req.service.api.signatureVersion) return done(); // none
-      if (req.service.getSignerClass(req.isPresigned()) === AWS.Signers.V4) {
+      if (req.service.getSignerClass(req) === AWS.Signers.V4) {
         var body = req.httpRequest.body || '';
         AWS.util.computeSha256(body, function(err, sha) {
           if (err) {
@@ -146,7 +146,7 @@ AWS.EventListeners = {
 
         try {
           var date = AWS.util.date.getDate();
-          var SignerClass = req.service.getSignerClass(req.isPresigned());
+          var SignerClass = req.service.getSignerClass(req);
           var signer = new SignerClass(req.httpRequest,
             req.service.api.signingName || req.service.api.endpointPrefix,
             req.service.config.signatureCache);
--- a/lib/event_listeners.js
+++ b/lib/event_listeners.js
@@ -92,7 +92,7 @@ AWS.EventListeners = {
     addAsync('COMPUTE_SHA256', 'afterBuild', function COMPUTE_SHA256(req, done) {
       req.haltHandlersOnError();
       if (!req.service.api.signatureVersion) return done(); // none
-      if (req.service.getSignerClass(req) === AWS.Signers.V4) {
+      if (req.service.getSignerClass(req.isPresigned()) === AWS.Signers.V4) {
         var body = req.httpRequest.body || '';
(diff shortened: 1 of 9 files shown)


LiuJoyceC:
Is it necessary to make so many changes to core functions that aren't specific to S3? What if instead, if a user doesn't explicitly specify `'v4'`, `'s3'`, or `'v2'` for the `signatureVersion`, then we store the string `'default'` (or some other string) in the `config.signatureVersion` of the S3 client. Then we can add the `getSignerClass` method to S3 that will return the appropriate signer class. After searching through the SDK, the only part of the code that actually uses the value stored in `config.signatureVersion` is inside the `getSignerClass`, so as long as you have a custom `getSignerClass` for S3 that overrides the generic service `getSignerClass`, then having an arbitrary string value as the `signatureVersion` shouldn't be a problem. I'll add more notes on specific lines of the PR.

lib/event_listeners.js
```diff
@@ -92,7 +92,7 @@ AWS.EventListeners = {
     addAsync('COMPUTE_SHA256', 'afterBuild', function COMPUTE_SHA256(req, done) {
       req.haltHandlersOnError();
       if (!req.service.api.signatureVersion) return done(); // none
-      if (req.service.getSignerClass(req) === AWS.Signers.V4) {
+      if (req.service.getSignerClass(req.isPresigned()) === AWS.Signers.V4) {
```
LiuJoyceC: Keeping the argument as `req` instead of a boolean indicating if it's presigned would keep this file more generic to all services, since determining signer class based on if the request is presigned is S3 specific. Although the generic service `getSignerClass` currently does not use any arguments, it's possible that more custom `getSignerClass` methods could be written for individual services. The `req.isPresigned()` could easily be moved into S3's custom `getSignerClass`.

chrisradek: That's a fair point, I'll make that change.

chrisradek:
We talked about the `signatureVersion` changes in person, so to quickly reiterate, one thing that is difficult to do today is keep track of which options a user specified, vs which options were defaulted based on config or some heuristic. Due to the way the service client constructors are implemented (they are really closer to factory methods that return a different object based on the apiVersion specified), we can't simply override a service client's constructor. Unfortunately, by the time the object is instantiated for us, the configuration will already be altered.

I added the `originalConfig` as a way to keep track of user-defined config. This information could also be valuable for other service clients, and it doesn't inject service-specific logic in the generic service constructor either. I took care to 'hide' it so that it is less likely to be altered by a user.

chrisradek:
Updated the PR based on feedback around the `getSignerClass` method.

lib/signers/presign.js
```diff
@@ -12,7 +12,7 @@ var expiresHeader = 'presigned-expires';
 function signedUrlBuilder(request) {
   var expires = request.httpRequest.headers[expiresHeader];
   var isPresigner = true;
```
LiuJoyceC: If this isn't being used anymore, it should be removed.

lib/signers/presign.js
```diff
@@ -11,11 +11,13 @@ var expiresHeader = 'presigned-expires';
  */
 function signedUrlBuilder(request) {
   var expires = request.httpRequest.headers[expiresHeader];
+  var isPresigner = true;
```
LiuJoyceC: Since the `isPresigner` variable is no longer used, it can be removed. (Same as a comment I made earlier, but I realized the other comment was made on an outdated diff)

chrisradek:
Replaced by #1023

PR #991, opened 2016-05-12
Fixes a bug where httpUploadProgress isn't emitted for files less than 16KB

winsome:
By default, nodejs streams buffer 16KB of data before a drain event will
be triggered. For small enough files, this means that the drain event
never occurs, the sendProgress event is never triggered, and thus
downstream httpUploadProgress never occurs. This change opts to use the
finish event which indicates that stream.end() has been called and all
data has been flushed.


chrisradek:
@winsome 
Thanks for the PR! Sorry this wasn't caught before. I think `drain` was used instead of `finish` initially to support node.js 0.8.x, but we never supported the `httpUploadProgress` event in version 0.8.x of node.js anyway. Merging in the code, it'll be available via NPM with the next release!

winsome:
@chrisradek Glad to help! I'll keep a look out for the next release

PR #992, opened 2016-05-12
S3 region redirect

### LINKED ISSUE #835: Behavior difference in DNS compliant and "simple" bucket names
I have 2 buckets in region eu-west-1, e.g. 'simplebucket' and 'dnsbucket.sc5.io'
I have the following code:

  var AWS = require('AWS') 
  var s3client = new AWS.S3()
  s3client.listObjects({ BucketName: ... }, function(err,data) { ... });

I have not initialized the region, i.e. seems to default to us-east-1. However, for some reason the code above succeeds for 'simplebucket' but fails for 'dnsbucket.sc5.io' with an error 'PermanentRedirect: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.'

Adding 
  AWS.config.region = 'eu-west-1' 
before defining s3client fixed the issue. 

Figuring the issue was not trivial since 'simplebucket' succeeded (i.e. assumed that the config was correct). From a developer's perspective, the functionality should be consistent, i.e. fail or succeed in both cases. In case of failure, the error could be more informative (e.g. "Wrong region for bucket" or similar).

LiuJoyceC:
Redirects to correct S3 bucket region if configured region does not match the bucket region. Should work in all cases except for requests in the browser for path-style buckets. Also caches bucket region so that future requests are corrected to the right region on the first try. Resolves #835 
/cc @chrisradek

### DIFF
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -340,21 +340,46 @@ AWS.util.update(AWS.S3.prototype, {
 
     var code = resp.httpResponse.statusCode;
     var body = resp.httpResponse.body || '';
+    var region = resp.httpResponse.headers['x-amz-bucket-region'] || null;
+    var bucket = resp.request.params.Bucket || null;
+    var bucketRegionCache = this.service.bucketRegionCache;
+    if (region && bucket && region !== bucketRegionCache[bucket]) {
+      bucketRegionCache[bucket] = region;
+    }
     if (codes[code] && body.length === 0) {
+      if (bucket && !region) {
+        region = bucketRegionCache[bucket] || null;
+      }
       resp.error = AWS.util.error(new Error(), {
-        code: codes[resp.httpResponse.statusCode],
-        message: null
+        code: codes[code],
+        message: null,
+        region: region
       });
     } else {
       var data = new AWS.XML.Parser().parse(body.toString());
+      if (data.Region && !region) {
+        region = data.Region;
+        if (bucket && region !== bucketRegionCache[bucket]) {
+          bucketRegionCache[bucket] = region;
+        }
+      } else if (bucket && !region && !data.Region) {
+        region = bucketRegionCache[bucket] || null;
+      }
       resp.error = AWS.util.error(new Error(), {
         code: data.Code || code,
         message: data.Message || null,
-        region: data.Region || null
+        region: region
       });
     }
     resp.request.service.extractRequestIds(resp);
   },
+
+  /**
+   * Cache for bucket region.
+   *
+   * @api private
+   */
+   bucketRegionCache: {},
 
   /**
    * Extracts S3 specific request ids from the http response.
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -495,7 +495,7 @@ AWS.util.update(AWS.S3.prototype, {
       return done();
     }
 
-    var regionReq = req.service.listObjects({Bucket: bucket});
+    var regionReq = req.service.listObjects({Bucket: bucket, MaxKeys: 0});
     regionReq._requestRegionForBucket = bucket;
     regionReq.send(function() {
       var region = req.service.bucketRegionCache[bucket] || null;
@@ -535,7 +535,7 @@ AWS.util.update(AWS.S3.prototype, {
       }
       done();
     } else if (request.httpRequest.virtualHostedBucket) {
-      var getRegionReq = service.listObjects({Bucket: bucket});
+      var getRegionReq = service.listObjects({Bucket: bucket, MaxKeys: 0});
       service.updateReqBucketRegion(getRegionReq, 'us-east-1');
       getRegionReq._requestRegionForBucket = bucket;
 
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -351,6 +351,9 @@ AWS.util.update(AWS.S3.prototype, {
     if (typeof region === 'string' && region.length) {
       httpRequest.region = region;
     }
+    if (!httpRequest.endpoint.host.match(/s3(?!-accelerate).*\.amazonaws\.com$/)) {
+      return;
+    }
     var service = request.service;
     var s3Config = service.config;
     var s3BucketEndpoint = s3Config.s3BucketEndpoint;
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -364,8 +364,7 @@ AWS.util.update(AWS.S3.prototype, {
     delete newConfig.endpoint;
     newConfig.region = httpRequest.region;
 
-    var endpoint = (new AWS.S3(newConfig)).endpoint;
-    httpRequest.endpoint = new AWS.Endpoint(endpoint);
+    httpRequest.endpoint = (new AWS.S3(newConfig)).endpoint;
     service.populateURI(request);
     s3Config.s3BucketEndpoint = s3BucketEndpoint;
     httpRequest.headers.Host = httpRequest.endpoint.host;
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -12,6 +12,16 @@ var operationsWith200StatusCodeError = {
   'uploadPartCopy': true
 };
 
+/**
+ * @api private
+ */
+ var regionRedirectErrorCodes = [
+  'AuthorizationHeaderMalformed', // non-head operations on virtual-hosted bucket endpoints
+  'BadRequest', // head operations on virtual-hosted bucket endpoints
+  'PermanentRedirect', // non-head operations on path-style bucket endpoints
+  301 // head operations on path-style bucket endpoints
+ ];
+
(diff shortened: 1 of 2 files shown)


lib/services/s3.js
```diff
@@ -338,22 +428,172 @@ AWS.util.update(AWS.S3.prototype, {
...
+   * @api private
+   */
+   bucketRegionCache: {},
+
+  /**
+   * Clears bucket region cache.
+   *
+   * @api private
+   */
+   clearBucketRegionCache: function(buckets) {
```
chrisradek: If `buckets` isn't passed into this method, can't we just set `this.bucketRegionCache` to an empty object? Or if you want to be thorough, `delete this.bucketRegionCache` first? Then you wouldn't need to loop through 2 lists to clear the cache completely.

LiuJoyceC: The SDK doesn't actually ever call `clearBucketRegionCache` without passing in bucket, except for in the unit tests where almost all of the tests store at most 1 bucket, so it shouldn't cause any performance issues in the SDK. The reason I decided to delete properties rather than replace the cache with a whole new object was that if customers wanted to call the method themselves, and they had assigned the `bucketRegionCache` to a variable (so they could avoid typing long chains such as `response.request.service.bucketRegionCache` multiple times), then `clearBucketRegionCache` would still behave the way they expect, and the `bucketRegionCache` they stored in their variable would be cleared and still valid. If I just replaced `s3.bucketRegionCache` with a new object, not only would that not clear the cache that they have stored in their variable, any modifications they make to the cache stored in their variable would unexpectedly no longer be accessible to the SDK. So the decision was mainly rooted in a more intuitive customer experience, but I don't see it making too big a difference either way, since the method is marked as private and I don't think there are many scenarios in which users would want to clear the entire cache instead of just a select number of buckets.

LiuJoyceC:
Per our discussion, the updated PR will no longer change the endpoint if you specify a custom endpoint, since customers could be using proxies or 3rd party S3 clones. However, if the user sets the endpoint to a real direct S3 endpoint, such as `{endpoint: 'https://s3-us-west-2.amazonaws.com'}`, then the endpoint will still get updated, since that can't be a proxy or S3 clone anyway. If we don't want to update the endpoint in that scenario, we would have to know the original configuration specified by the customer, so the other PR (defaulting to v4) would need to be merged in first.

We discussed making `getBucketLocation` requests instead of `listObjects` for retrieving the bucket location. However, after further investigation, I realized that this would not work unless the customer configures their IAM policy to allow the `getBucketLocation` operation, and would be a breaking change if current customers haven't already done that. The reason `listObjects` works even if the IAM policy doesn't allow that operation is that it always returns the region in the header, even if you receive an `AccessDenied` error. In almost all of the scenarios in which the SDK calls `listObjects` to get the region, the response will return an error that will not be retried because the region will have been extracted and cached. Only in one scenario in the browser, the `listObjects` may return a successful response, so in the updated PR, I specified a `MaxKeys` constraint of 0 so that no object keys will be returned.

lib/services/s3.js
```diff
@@ -322,7 +393,29 @@ AWS.util.update(AWS.S3.prototype, {
         resp.data.LocationConstraint = '';
       }
     }
-    resp.request.service.extractRequestIds(resp);
+    var bucket = req.params.Bucket || null;
+    if (req.operation === 'deleteBucket' && typeof bucket === 'string' && !resp.error) {
```
chrisradek: Solid!

lib/services/s3.js
```diff
@@ -306,6 +341,42 @@ AWS.util.update(AWS.S3.prototype, {
...
+    newConfig.region = httpRequest.region;
+
+    var endpoint = (new AWS.S3(newConfig)).endpoint;
+    httpRequest.endpoint = new AWS.Endpoint(endpoint);
+    service.populateURI(request);
+    s3Config.s3BucketEndpoint = s3BucketEndpoint;
+    httpRequest.headers.Host = httpRequest.endpoint.host;
+
+    if (request._asm.currentState === 'validate') {
+      request.removeListener('build', service.populateURI);
```
chrisradek: Are you removing the `populateURI` because you want to avoid duplicating some work?

lib/services/s3.js
```diff
@@ -306,6 +341,42 @@ AWS.util.update(AWS.S3.prototype, {
...
+    var s3BucketEndpoint = s3Config.s3BucketEndpoint;
+    if (s3BucketEndpoint) {
+      delete s3Config.s3BucketEndpoint;
+    }
+    var newConfig = AWS.util.copy(s3Config);
+    delete newConfig.endpoint;
+    newConfig.region = httpRequest.region;
+
+    var endpoint = (new AWS.S3(newConfig)).endpoint;
+    httpRequest.endpoint = new AWS.Endpoint(endpoint);
```
chrisradek: Is it necessary to create a new Endpoint object when you have one from the new S3 object already?

lib/services/s3.js
```diff
@@ -338,22 +431,172 @@ AWS.util.update(AWS.S3.prototype, {
...
+    if (!error || !bucket || error.region || req.operation === 'listObjects' ||
+        (AWS.util.isNode() && req.operation === 'headBucket') ||
+        (error.statusCode === 400 && req.operation !== 'headObject') ||
+        regionRedirectErrorCodes.indexOf(error.code) === -1) {
+      return done();
+    }
+
+    var regionReq = req.service.listObjects({Bucket: bucket, MaxKeys: 0});
+    regionReq._requestRegionForBucket = bucket;
+    regionReq.send(function() {
```
chrisradek: Is this region available within the callback because the `regionReq` request updates the same cache that the original request uses?

lib/services/s3.js
```diff
@@ -338,22 +431,172 @@ AWS.util.update(AWS.S3.prototype, {
...
+   */
+   reqRegionForNetworkingError: function reqRegionForNetworkingError(resp, done) {
+    if (!AWS.util.isBrowser()) {
+      return done();
+    }
+    var error = resp.error;
+    var request = resp.request;
+    var bucket = request.params.Bucket;
+    if (!error || error.code !== 'NetworkingError' || !bucket ||
+        request.httpRequest.region === 'us-east-1') {
```
chrisradek: Do we ever see the `NetworkingError` when the region is set to `us-east-1`? I'm curious why we're specifically checking this region.

lib/services/s3.js
LiuJoyceC: Not only would it duplicate work, it would break the endpoint for virtual-hosted buckets because the bucket name gets populated twice, e.g. `mybucketname.mybucketname.s3.amazonaws.com`. This only is a problem for requests that have not been built yet (requests still on the `validate` step), since requests that are being retried go straight to the `sign` step and do not go through the `build` step anymore. This is also why `populateURI` needs to be called inside this function, because requests being retried need to update their endpoint but won't go through the `build` step.

lib/services/s3.js
LiuJoyceC: Yes. The cache is on `S3.prototype`, so all S3 instances share the same cache.

lib/services/s3.js
LiuJoyceC: Because `NetworkingError` can be caused by other reasons, it can occur when the region is set to `us-east-1`, but it won't be because of incorrect region. A `BadRequest` or `AuthorizationHeaderMalformed` would be received in that case instead, and would be handled by the `extractError` listener. If the region is set to `us-east-1`, this listener wouldn't be doing anything helpful for it anyway, since it is just either changing the region to `us-east-1`, or trying a region request in `us-east-1`. This exclusion filter is needed to prevent this listener from recursively sending off an infinite number of requests that cause the next request to be sent. For example, if there is something wrong with the network and we're receiving a `NetworkingError` because of it, then this listener would keep getting triggered because each request it sends off would receive another `NetworkingError`. So instead of allowing that, this listener will check that the region is not `us-east-1`, because otherwise the `NetworkingError` is not due to incorrect region.

lib/services/s3.js
LiuJoyceC: The original `httpRequest.endpoint` was constructed by passing the service object endpoint into the `AWS.Endpoint` constructor. This essentially copies the non-hidden properties of the service object endpoint to a new object, which won't contain any of the hidden properties of the original. If the only purpose of constructing a new endpoint is to create a new copy so it's independent from the service object endpoint, then it won't be necessary to do it here, but I'll have to test to make sure the hidden properties don't make a difference.

LiuJoyceC:
I updated the `updateReqBucketRegion` method so that the endpoint on the new S3 client does not get copied (through the `AWS.Endpoint` constructor), since it's unnecessary. I've tested it to make sure the hidden properties do not affect the request in any way.

chrisradek:
Ship it!

PR #1000, opened 2016-05-23
Adds support for signing file streams that specify a range.

### LINKED ISSUE #998: Cannot stream file to s3 using signature version v4
I am using node 0.10.30.    I would like to stream a large file to S3 in multipart fashion.    
This works when using signatureVersion V2.    If I change to signatureVersion v4, the operation fails with the error:  The provided 'x-amz-content-sha256' header does not match what was computed when uploading the first part.

```
(function() {
  'use strict';

  // universal stuff
  var fs = require('fs'),

  // configurable options
    AWS = require('aws-sdk'),
    bucket = 'yourbucketname',
    key = 'testing-multipart/6mb.txt',
    // created with command: dd if=/dev/zero of=/tmp/6mb.txt bs=6291456 count=1
    fileName = '/tmp/6mb.txt';


  AWS.config.update(
    {
      accessKeyId: 'akeyid',
      secretAccessKey: 'asecretkeyid',
      region: 'us-east-1'
    });

  var s3 = new AWS.S3({
    maxRetries: 1,
    signatureVersion: 'v4',    //if you comment this out, everything works
    logger: process.stdout
  });


  if (require.main === module) {

    var createParams = {
      Bucket: bucket,
      Key: key,
      ServerSideEncryption: 'AES256'
    };

    var req = s3.createMultipartUpload(createParams);

    req.send(function (err, createResponse) {

      var uploadId = createResponse.UploadId,
        partResults = [],
        firstPart = {
          Bucket: bucket,
          Key: key,
          UploadId: uploadId,
          PartNumber: '1',
          Body: fs.createReadStream(fileName, {
            start: 0,
            end: 5242879
          }),
          ContentLength: 5242880
        };


      var reqPart1 = s3.uploadPart(firstPart);


      reqPart1.send(function (err, firstPartResult) {
        if (err) {
          console.log('failed to upload first part: ' + JSON.stringify(err, null, '  '));
          s3.abortMultipartUpload({
            Bucket: bucket,
            Key: key,
            UploadId: uploadId
          });
        } else {
          console.log('successfully uploaded the first part');
        }
      });
    });

  }
})();
```

I'm trying to determine if this is a SDK bug or a procedural piece that I am missing.

My understanding is the SDK should handle setting all the headers correctly.    Is that correct or am i missing a step when trying to move to sig version V4?

I also understand that there is an abstraction for managed uploads, but this does not workl on node 10 with streams of large files

Any thoughts on how best to proceed?

chrisradek:
Fixes #998 

/cc @LiuJoyceC

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -700,7 +700,14 @@ var util = {
       var fs = util.nodeRequire('fs');
       if (body instanceof Stream) {
         if (typeof body.path === 'string') { // assume file object
-          body = fs.createReadStream(body.path);
+          var settings = {};
+          if (typeof body.start === 'number') {
+            settings.start = body.start;
+          }
+          if (typeof body.end === 'number') {
+            settings.end = body.end;
+          }
+          body = fs.createReadStream(body.path, settings);
         } else { // TODO support other stream types
           return done(new Error('Non-file stream objects are ' +
                                 'not supported with SigV4'));
--- a/test/services/s3.spec.coffee
+++ b/test/services/s3.spec.coffee
@@ -1120,7 +1120,7 @@ describe 'AWS.S3', ->
         req = s3.putObject(Bucket: 'example', Key: 'key', Body: stream)
         req.send (err) ->
           expect(mock.calls[0].arguments).to.eql(['path/to/file'])
-          expect(mock.calls[1].arguments).to.eql(['path/to/file'])
+          expect(mock.calls[1].arguments).to.eql(['path/to/file', {}])
           expect(err).not.to.exist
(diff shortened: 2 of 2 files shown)


LiuJoyceC:
It would be good to add a test specifically for streams with a specified range. You could create a test similar to the one you added a minor modification for, but instead of putObject you test a multipart upload and check that the mock `createReadStream` was called with the correct `start` and `end` values in the second argument.

chrisradek:
I updated the tests. A `putObject` call is fine, as that should also work with a stream created from a range.

LiuJoyceC:
:shipit:

PR #1009, opened 2016-05-31
Updates readFileSync to more accurately determine if the SDK is in a …

### LINKED ISSUE #1006: lib util readFileSync does not work in Electron
The `readFileSync` function in [lib/util.js](https://github.com/aws/aws-sdk-js/blob/master/lib/util.js#L100)  does not work in environments like [Electron](http://electron.atom.io/).
In Electron, typeof window := "object" but I have access to node's fs module.
Maybe it would be easier to check if we are in a node.js context?

Btw: This behavior prevents me from using version 2.3.16 lib/credentials/shared_ini_file_credentials.js with Electron :)

chrisradek:
…browser environment.

Fixes #1006

/cc @LiuJoyceC

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -98,7 +98,7 @@ var util = {
   },
 
   readFileSync: function readFileSync(path) {
-    if (typeof window !== 'undefined') return null;
+    if (util.isBrowser()) return null;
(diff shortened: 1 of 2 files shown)


LiuJoyceC:
🚢 🇮🇹

PR #1023, opened 2016-06-14
Changes default signer when using the S3 client to signature version 4.

chrisradek:
Replaces #989 

/cc @LiuJoyceC 

This is exactly the same as #989, except that it fixes the merge conflicts and removes the unused variable from `presign.signedUrlBuilder` based on your last comment.

With this change I'll do a minor update to the SDK.


LiuJoyceC:
🚀 🌔

PR #1028, opened 2016-06-21
Fix for Cognito Credentials: ID becomes null issue

mdravida19:
Fixes for issues on aws/amazon-cognito-js and aws/amazon-cognito-identity-js

### DIFF
--- a/test/credentials.spec.coffee
+++ b/test/credentials.spec.coffee
@@ -741,20 +741,15 @@ describe 'AWS.CognitoIdentityCredentials', ->
     
     it 'should call clearCachedId if user is not authorized', ->
       clearCache = helpers.spyOn(creds,'clearCachedId')
-      helpers.mockResponses [
-        {data: null, error: {code: 'NotAuthorizedException'}}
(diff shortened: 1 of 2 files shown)


test/credentials.spec.coffee
```diff
@@ -859,7 +873,7 @@ describe 'AWS.CognitoIdentityCredentials', ->
       expect(creds.cacheId.calls.length).to.equal(0)
       expect(creds.getStorage('id')).not.to.exist
 
-    it 'clears cache if getId fails', ->
```
LiuJoyceC: Can you include a corresponding test for each of these scenarios that tests that the cache is cleared for unauthorized users? So one for each of getId, getOpenIdToken, and getCredentialsForIdentity.

neosekar:
This is not working, 
Cognito sync while offline access in javascript sdk, 
**AWS.config.credentials.get(function(err) {});** this method returns **Network failure error**

johnborges:
Shouldn't this use the cached IdentityId?

PR #1030, opened 2016-06-22
getRtmpUrl should not append hash if it's null

lytc:


chrisradek:
@lytc 
Looks good to me, thanks for the PR!

Issue #965, opened 2016-04-15
Error: Non-file stream objects are not supported with SigV4 when using s3.uploadPart()


olivierlesnicki:
I get the following error:
```
Non-file stream objects are not supported with SigV4
```
when passing a `FileStream` to `s3.uploadPart()`.

Oddly I don't get the error if I swap `s3.uploadPart()` for `s3.upload()` - this leads me to believe my object is a valid file stream.

olivierlesnicki:
```
  signatureVersion: 'v4',
  region: 'eu-west-1',
```

chrisradek:
@olivierlesnicki 
Can you share some more details on how you're doing a multi-part upload with a stream? Internally for the `upload` operation, the SDK will read the stream in 5 MB chunks, and make the `uploadPart` calls using those buffered chunks.

olivierlesnicki:
The stream is created by `Busboy` https://github.com/mscdex/busboy and I just pass it as is to s3

```js
router.post('/upload/:key/:uploadId/:partNumber', function(req, res, next) {
  var busboy = new Busboy({
    headers: req.headers
  });

  busboy.on('file', (fieldname, stream, filename, encoding, mimetype) => {
      let s3req = s3.uploadPart({
        Bucket: process.env.AWS_BUCKET,
        Key: req.params.key,
        UploadId: req.params.uploadId,
        PartNumber: req.params.partNumber,
        Body: stream,
      });

      s3req.send((err, data) => {
        if (err) {
          return next(err);
        }

        res.send(data);
      });

  });

  req.pipe(busboy);
});
```

Note that if I build the incoming busboy stream in memory before handing it to `s3.uploadPart()` it works fine:

The stream is created by `Busboy` https://github.com/mscdex/busboy and I just pass it as is to s3 - but that's just to slow and memory intensive.

```js
router.post('/upload/:key/:uploadId/:partNumber', function(req, res, next) {
  var busboy = new Busboy({
    headers: req.headers
  });

  busboy.on('file', (fieldname, stream, filename, encoding, mimetype) => {
    stream.fileRead = [];
    stream.on('data', (chunk) => {
      stream.fileRead.push(chunk);
    });

    stream.on('end', () => {
      let s3req = s3.uploadPart({
        Bucket: process.env.AWS_BUCKET,
        Key: req.params.key,
        UploadId: req.params.uploadId,
        PartNumber: req.params.partNumber,
        Body: stream.fileRead,
      });

[...]

PR #1034, opened 2016-06-23
Allows non-file streams to be used with S3 SigV4 when body signing is disabled (default).

chrisradek:
Users uploading non-file streams when using sigv2 by default may have been affected when version 2.4.0 of the SDK was updated to switch sigv4 as the default signer.

Related to #965

/cc @LiuJoyceC


LiuJoyceC:
⛵

PR #1041, opened 2016-06-30
Modifies SDK region request to use headBucket instead of listObjects in Node

LiuJoyceC:
Currently, when the SDK makes a request for bucket region for redirect purposes, it calls listObjects because in the browser, CORS restrictions may prevent region information from being obtained through headBucket because region information is only exposed in the response header. However, this restriction does not apply in Node, so this PR changes the region request to use headBucket instead of listObjects in the Node environment.

/cc: @chrisradek

### DIFF
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -536,7 +536,7 @@ AWS.util.update(AWS.S3.prototype, {
    *
    * @api private
    */
-   requestBucketRegion: function requestBucketRegion(resp, done) {
+  requestBucketRegion: function requestBucketRegion(resp, done) {
     var error = resp.error;
     var req = resp.request;
     var bucket = req.params.Bucket || null;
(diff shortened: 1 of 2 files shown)


lib/services/s3.js
```diff
@@ -547,15 +547,15 @@ AWS.util.update(AWS.S3.prototype, {
         regionRedirectErrorCodes.indexOf(error.code) === -1) {
       return done();
     }
-
-    var regionReq = req.service.listObjects({Bucket: bucket, MaxKeys: 0});
+    var reqOperation = AWS.util.isNode() ? 'headBucket' : 'listObjects';
+    var regionReq = req.service[reqOperation]({Bucket: bucket, MaxKeys: 0});
```
chrisradek: `MaxKeys` isn't a valid parameter for `headBucket`, so you'll get an `UnexpectedParameter` error here.

LiuJoyceC: Good catch. I'll modify that.

chrisradek:
:shipit:

PR #1045, opened 2016-07-06
Adds changelog to JS SDK. Adds script for automating update of change…

LiuJoyceC:
…log based on json files in .changes directory.

/cc @chrisradek


LiuJoyceC:
I couldn't add the .changes directory yet because I can't commit an empty directory, but we can add it when the first changelog json is added.

scripts/update-changelog
```diff
@@ -0,0 +1,80 @@
...
+	throw new Error('json not in correct format');
+}
+
+var entry = '\n\n## ' + version + '\n<ul>';
+
+changes.forEach(function(change) {
+	if (!change.type || !change.category || !change.description) {
+		throw new Error('json not in correct format');
+	}
+	entry += '\n\t<li>' + change.type + ': ' + change.category + ': ' +
```
chrisradek: Might want to use standard markdown instead of html:
https://guides.github.com/features/mastering-markdown/

scripts/update-changelog
```diff
@@ -0,0 +1,80 @@
...
+if (version !== undefined) {
+	if (version === 'major') {
+		version = (lmajor + 1) + '.0.0';
+	} else if (version === 'minor') {
+		version = lmajor + '.' + (lminor + 1) + '.0';
+	} else {
+		version = version.match(new RegExp('^' + versionRegStr + '$'));
+		if (version) {
+			// convert to num for comparison and for normalizing leading zeros
+			var major = +version[1];
```
chrisradek: It works, but `parseInt(version[x], 10)` is clearer and doesn't make use of a JS quirk to convert a string to a number. We'll never need to worry about floats with our versioning.

LiuJoyceC:
For now, service updates for releases will need to have JSON files manually created in the `next-release` directory along with the JSON files for PRs.

chrisradek:
Do you mind adding a README for these new scripts that explains what they do, or how they should be run? I know we don't expect these particular scripts to be run by users but it would still be helpful for future contributors.

LiuJoyceC:
New changes have been pushed. The README contains a fictional example of an entry in the changelog that we should swap out for a real example when we create our first real entry.

chrisradek:
:shipit:

Issue #671, opened 2015-07-30
Support KeepAlive for HTTP requests


parisholley:
Currently, neither HTTP (no agent configured ) or HTTPS agent within http/node.js is using an agent with keep alive configured. For a high number of calls (for example processing messages of a large queue), it is throttled tremendously. I'm hacking in a agent for testing, which has shown a tremendous improvement. Going to research a bit more and potentially issue a pull.

AdityaManohar:
@parisholley The SDK is supported across Node.js v0.8, v0.10, and v0.12. Keep-Alive supported landed only in Node v.0.12. That being said it is simple to configure a keep-alive agent with the SDK by using `httpOptions`

```javascript
var AWS = require('aws-sdk');
var http = require('http');
AWS.config.update({
  httpOptions: {
    agent: new http.Agent({keepAlive: true})
  }
})
```

Let me know if this helps.

parisholley:
Already processed my backlog, but I've implemented this and sure it solves my issue. Though I wonder if it is worth detecting the node version, and implementing by default, or updating documentation to note the performance gain by using an agent.

AdityaManohar:
@parisholley Given how simple it is to use a keep-alive agent, I think documenting this is a better approach for now. I'll make sure to add a note about using keep-alive agents to the configuring section of our developer guide. Thank you for bringing this to our attention!

Alonkad:
We also experienced this issue with Firehose.
We have a NodeJS server that accepts requests and writes to Firehose (the SDK uses HTTPS agent for that).
When the throughput reached a certain level (depends on the VM) we started getting timeouts (after the default 2 minutes).
Adding keep-alive solved the problem (tested with twice the load that originally failed).

We read the documentation - no mention of the above subject.
This is highly important for services like Firehose which are build for scale (users will experience this issue real quick)

Please consider adding keep-alive by default.

PR #1047, opened 2016-07-08
Use AWS.config.httpOptions as default httpOptions

pdesgarets:
Currently, AWS.config.httpOptions seems not to be used in handleRequest. Which makes the answer to  #671 incorrect.

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -13,7 +13,7 @@ AWS.NodeHttpClient = AWS.util.inherit({
     var cbAlreadyCalled = false;
     var endpoint = httpRequest.endpoint;
     var pathPrefix = '';
-    if (!httpOptions) httpOptions = {};
+    httpOptions = httpOptions ? AWS.util.merge(AWS.config.httpOptions, httpOptions) : AWS.util.copy(AWS.config.httpOptions);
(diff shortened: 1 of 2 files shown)


chrisradek:
@pdesgarets 
Thanks for your pull request!

This change, however, would change the way configuration works for our SDK.
We document that configuration data is immutable:
http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-configuring.html#Immutable_Configuration_Data

[Here](https://github.com/aws/aws-sdk-js/blob/master/lib/event_listeners.js#L239) you can see that we grab the `httpOptions` from the service client configuration.  If `httpOptions` is defined on `AWS.config` when the service client is instantiated, then those settings will be used unless they were also provided to the service client's constructor.

This change actually has potential to break existing code. If some users have created service clients, then updated the AWS.config.httpOptions (possibly adding a proxy, or other custom settings) and created additional service clients, the already existing ones would now use the global config, which is counter to what our documentation states.

With regards to the answer in #671, users would need to either update the `AWS.config.httpOptions` prior to creating their service clients, or pass the config directly to their service client constructors:
```javascript
var s3 = new AWS.S3({
  httpOptions: {/*...*/}
});
```

PR #1048, opened 2016-07-08
Documents that S3 requests with stream bodies cannot be retried.

### LINKED ISSUE #1043: RequestTimeout in s3 putObject when region is not specified
I'm not sure if it's by design (there's a point in docs that sdk doesn't auto-select region). It took me like several hours of attempts to make it work and I've found several issues with same problem without solution (#281, #454). So I've decide to let you know about it.

This will fail with timeout for me:

```
const s3Client = new aws.S3({
  accessKeyId: 'such-a-key',
  secretAccessKey: 'so-secret'
});
s3Client.putObject(//...
```

and this will work:
```
const s3Client = new aws.S3({
  accessKeyId: 'such-a-key',
  secretAccessKey: 'so-secret',
  region: 'eu-central-1'
});
s3Client.putObject(//...
```

LiuJoyceC:
Resolves #1043 
/cc: @chrisradek

### DIFF
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -774,7 +774,9 @@ AWS.util.update(AWS.S3.prototype, {
    * @overload upload(params = {}, [options], [callback])
    *   Uploads an arbitrarily sized buffer, blob, or stream, using intelligent
    *   concurrent handling of parts if the payload is large enough. You can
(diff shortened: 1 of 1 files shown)


chrisradek:
:shipit:

PR #1054, opened 2016-07-12
Fixes issue where 2 clients from the same service would share a signature cache.

### LINKED ISSUE #1020: Cloudwatch SDK cache's key signing across objects
The docs for the [cloudwatch sdk](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudWatch.html#constructor-property) constructor show a parameter for `signatureCache`. I understand why this would perform caching on the object that this constructor call creates; that makes sense.

However, it controls caching for other constructed SDK objects. *This is surprising behavior*. Object constructors should only affect the object they are creating; they should not set global flags that affect future objects using the same constructor. *Parameters to one constructor should not affect global application state*.

I have a lambda that needs to constuct multiple Cloudwatch instances, to access Cloudwatch alarms in multiple accounts. Calling this constructor twice, with different access keys, should not produces two Cloudwatch instances that use the same keys.

### LINKED ISSUE #1052: Credential caching
I'm using the AWS-SDK with several sets of credentials. I use this code to test if the credentials are correct and if they have the right permissions. I see a strange caching of the credentials when switching between the different credentials.

Here's an example code:
```
AWS     = require 'aws-sdk'

correctCredentials = '...' #these are correct credentials
wrongCredentials = '...' #these are wrong credentials

s3Correct = new AWS.S3
  credentials: correctCredentials

s3Wrong = new AWS.S3
  credentials: wrongCredentials

params =
  Bucket: 'mybucket'

s3Correct.listObjectsV2 params, (err, data)=>
  if err?
    console.log err
  unless err?
    console.log 'no error'

s3Wrong.listObjectsV2 params, (err, data)=>
  if err?
    console.log err
  unless err?
    console.log 'no error'
```

On version 2.3.19 of the SDK I get no error on the correct credentials and an error on the wrong credentials, while with version 2.4.6 I get **no errors on both**.

It seems that while I don't change the `awsAccessKeyId`, the SDK caches the `awsSecretAccessKey`.

chrisradek:
This issue would result in the same signed key to be used by 2 different clients if they were configured to use the same region and accessKeyId. (Very unlikely if possible).

This issue also resulted in the signature cache being cleared anytime the client making a request was alternated with another client of the same service type.

Fixes #1020
Fixes #1052

/cc @LiuJoyceC 
I added a unique client id to each client created to differentiate them when caching the signature. It's hidden for now as I didn't want it to be easy for users to find, but I am open to creating a method on Service to return the ID as well.

### DIFF
--- a/lib/signers/v4.js
+++ b/lib/signers/v4.js
@@ -9,6 +9,16 @@ var cachedSecret = {};
 /**
  * @api private
  */
+var cacheQueue = [];
+
+/**
+ * @api private
+ */
+var maxCacheEntries = 50;
+
+/**
+ * @api private
+ */
 var expiresHeader = 'presigned-expires';
 
 /**
@@ -100,6 +110,15 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
     var cacheIdentifier = this.serviceName + (this.getServiceClientId() ? '_' + this.getServiceClientId() : '');
     if (this.signatureCache) {
       var cache = cachedSecret[cacheIdentifier];
+      // If there isn't already a cache entry, we'll be adding one
+      if (!cache) {
+        cacheQueue.push(cacheIdentifier);
+        if (cacheQueue.length > maxCacheEntries) {
+          // remove the oldest entry (may not be last one used)
+          delete cachedSecret[cacheQueue.shift()];
+        }
+      }
+
     }
     var date = datetime.substr(0, 8);
 
--- a/lib/signers/v4.js
+++ b/lib/signers/v4.js
@@ -97,8 +97,9 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
 
   signature: function signature(credentials, datetime) {
     var cache = null;
+    var cacheIdentifier = this.serviceName + (this.getServiceClientId() ? '_' + this.getServiceClientId() : '');
     if (this.signatureCache) {
-      var cache = cachedSecret[this.serviceName];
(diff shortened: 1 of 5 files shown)


lib/signers/v4.js
```diff
@@ -117,13 +118,13 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
         return AWS.util.crypto.hmac(kCredentials, this.stringToSign(datetime), 'hex');
       }
 
-      cachedSecret[this.serviceName] = {
+      cachedSecret[cacheIdentifier] = {
         region: this.request.region, date: date,
         key: kCredentials, akid: credentials.accessKeyId
       };
```
LiuJoyceC: We should put a maximum size constraint on `cachedSecret`. Before, there was only one cache per service (and only the last signature key for that service is kept), so the cache was necessarily bounded by the number of services. There is no upper bound on the number of service clients generated. Some applications could be generating a new service client for every request. The service client objects get garbage-collected eventually, but the cache doesn't and could grow to millions of objects fairly quickly and hog a lot of memory. Perhaps we could have a `cacheIdentifierQueue` of some sort so that when it's length reaches a certain number, we can get the oldest `cacheIdentifier` and delete it from `cachedSecret`.

chrisradek: Great point! I like the idea of having a queue keep track of what should be removed from cache first.

chrisradek:
Code has been updated to only cache up to 50 service clients at a time.

LiuJoyceC:
🚢

PR #1057, opened 2016-07-13
Sets a default value for maxSockets of 50 when the https.globalAgent.maxSockets is set to Infinity.

chrisradek:
This is a performance tuning change. With node.js 0.10 and lower, the default maxSockets on the global agent was 5, but later versions of node.js sets this to Infinity.

/cc @LiuJoyceC


LiuJoyceC:
🚀

PR #1058, opened 2016-07-13
ECSCredentials

LiuJoyceC:
This adds a new ECSCredentials provider to source refreshable credentials from an ECS container.

/cc: @chrisradek


chrisradek:
:shipit:

PR #1059, opened 2016-07-14
Adds add-change CLI to allow easy creation of changelog entry JSON files.

chrisradek:
We can merge our README.md entries together, I included your README entries to make it easier to see what the final README will look like.

Tested the cli on mac os x terminal and windows command prompt.

/cc @LiuJoyceC

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,8 +77,22 @@ npm run-script unit
 
 See the implementation of the `test` script in `package.json` for more options.
 
+### Changelog
+
+We have moved to using a changelog to document changes between SDK versions instead of [release notes][releasenotes].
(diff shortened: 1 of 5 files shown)


LiuJoyceC:
Not a big deal, but for the sake of consistency (and therefore less of a need to refer back to documentation to remind oneself of which command to run for which script), it would be nice to add `#!/usr/bin/env node` at the beginning of of the `add-change` script so the user could run it with or without prepending `node` to their command, like the other scripts in the `script` folder. But otherwise, :shipit: !

chrisradek:
The reason I didn't add `#!/usr/bin/env node` was because that won't work in a windows environment. I could add a separate script for unix/windows, but that didn't seem easier than just calling the script with node.

LiuJoyceC:
🐑

PR #1067, opened 2016-07-20
Parse ini files with comments starting with #

Jonnymcc:
Related to https://github.com/aws/aws-sdk-ruby/issues/837.

Boto config says to use # to begin comments and some may use it as an alternative to ; in the shared credentials file. This change will support using both for comments.

### DIFF
--- a/.changes/next-release/feature-util-5a565afe.json
+++ b/.changes/next-release/feature-util-5a565afe.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "util",
+  "description": "Parse ini files containing comments using #"
+}
--- a/lib/util.js
+++ b/lib/util.js
@@ -189,7 +189,7 @@ var util = {
     parse: function string(ini) {
       var currentSection, map = {};
       util.arrayEach(ini.split(/\r?\n/), function(line) {
-        line = line.split(/(^|\s);/)[0]; // remove comments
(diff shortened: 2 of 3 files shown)


chrisradek:
@Jonnymcc 
Thanks for the PR! I'll just verify this works but don't see any obvious issues.

chrisradek:
@Jonnymcc 
Would you mind just adding a few more lines to our existing test?
https://github.com/aws/aws-sdk-js/blob/master/test/util.spec.coffee#L203

Something like:
```coffeescript
key3 = value4 # yet another comment
[emptysection]
#key1=value1
```
and
```coffeescript
expect(map.section1.key3).to.equal('value4')
```

Also, please run the `add-change` script to create a change-log entry:
https://github.com/aws/aws-sdk-js/blob/master/CONTRIBUTING.md
`node ./scrips/changelog/add-change.js`
This will create a JSON file that looks much like [this](https://github.com/aws/aws-sdk-js/blob/master/.changes/2.4.8.json#L2-L6) that will help us provide more detail in our changelogs.

If you could make those changes, I can merge this in right away. Otherwise I can make those changes in a new PR if you don't have time to do this.

Again, thanks for submitting the PR!

chrisradek:
@Jonnymcc 
Thanks for making the updates! Assuming the tests pass, we'll merge this in.

chrisradek:
Changes have been merged. Thank you for the contribution!

PR #1072, opened 2016-07-22
request.createReadStream emits error for mismatched content-length

LiuJoyceC:
Readable stream from request.createReadStream emits error when bytes received is fewer than content-length header

/cc: @chrisradek

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -585,15 +585,17 @@ AWS.Request = inherit({
         if (expectedLen !== undefined && !isNaN(expectedLen) && expectedLen >= 0) {
           shouldCheckContentLength = true;
           var receivedLen = 0;
+        }
 
-          var checkContentLength = function checkContentLength() {
-            if (receivedLen !== expectedLen) {
-              stream.emit('error', AWS.util.error(
-                new Error('Stream content length mismatch. Received ' +
-                  receivedLen + ' of ' + expectedLen + ' bytes.'),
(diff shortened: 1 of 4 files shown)


lib/request.js
```diff
@@ -576,19 +576,63 @@ AWS.Request = inherit({
...
+
+        var checkContentLengthAndEmit = function checkContentLengthAndEmit() {
+          if (shouldCheckContentLength && receivedLen !== expectedLen) {
+            stream.emit('error', AWS.util.error(
+              new Error('Stream content length mismatch. Received ' +
+                receivedLen + ' of ' + expectedLen + ' bytes.'),
+              { code: 'StreamContentLengthMismatch' }
+            ));
+          } else {
+            stream.emit('end')
```
chrisradek: At least when using node.js 0.10.x and higher (streams 2+), you'll want to actually call `stream.end()`. This will emit the `unpipe` and `end` events, and clean up some listeners on the source/destination streams that aren't being cleaned when we just emit `end`.

That won't work in node 0.8, but just emitting `end` might suffice there.

LiuJoyceC:
@chrisradek 
I've updated based on the feedback. Note that calling `stream.end()` will only emit an `'end'` event after the stream has been piped into a writeable stream. But this behavior is consistent with what was happening before this PR anyway, so this is the desired behavior.

chrisradek:
:shipit:

PR #1077, opened 2016-07-27
docs/fix: Explicit about bucket name uniqueness

valgaze:
It's easy to miss the line that s3 bucket names must be to be unique-- provided whitespace & IMPORTANT to call it out. 

The error is a bit cryptic & reads like a permissions error:
image

These docs need a unique S3 bucket name comment: https://aws.amazon.com/sdk-for-node-js/


LiuJoyceC:
Hi @valgaze 

Thanks for the PR. I will merge it in.

Issue #570, opened 2015-04-09
S3 upload fails when file size exceeds 5mb


ecdeveloper:
When I'm trying to upload a file larger than 5mb with `s3.upload()` it just hangs. After some time it throws the following error - `Part number must be an integer between 1 and 10000, inclusive`. After a quick research, I end up putting the following code into this line - https://github.com/aws/aws-sdk-js/blob/master/lib/s3/managed_upload.js#L410:
```
console.log('upload part', chunk.toString().length, partNumber);
```
When I try to upload a file larger than 5mb, I see the following output:
```
upload part 5044533 1
upload part 98150 2
upload part 0 3
upload part 0 4
...
```
It keeps trying to upload a 0-length chunk endlessly (until it hits the maxTotalParts limit).

I use the latest module version - 2.1.22. Node version - 0.10.30. OS - OSX 10.9.5

AdityaManohar:
@ecdeveloper 
Thanks for reporting! We are working on a fix and will have a patch release out for this regression shortly.

ecdeveloper:
Great. Thanks!

AdityaManohar:
@ecdeveloper 
This fix is out in master. I will cut a patch release out shortly.

AdityaManohar:
@ecdeveloper 
I just cut a release `v2.1.23`. Let me know if this works for you.

ecdeveloper:
@AdityaManohar Perfect! Works like a charm now. Thank you!

AdityaManohar:
@ecdeveloper Thanks for reporting! Glad you have it working.

PR #1079, opened 2016-07-28
Adds Promise support for S3.ManagedUpload

### LINKED ISSUE #982: S3.ManagedUpload send method doesn't expose promise()
I am using `version": "2.3.7`

### LINKED ISSUE #1076: Error : s3.upload.promise() is not a funciton
I'm trying to call the setPromisesAWS.config.setPromisesDependency config function but my config object doesn't seem to have the setPromisesDependency property.
Am I missing something?

    AWS.config.setPromisesDependency(bluebird)
    // Error : AWS.config.setPromisesDependency is not a function

I have version 2.2.48 installed via bower.
Any help would be appreciated.

### LINKED ISSUE #1121: EC2MetadataCredentials.refresh does not return a promise
[`EC2MetadataCredentials.refresh`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2MetadataCredentials.html#refresh-property) can not be converted to a promise, it returns `undefined` instead of something I could call `.promise()` on.

LiuJoyceC:
Adds Promise support for S3.ManagedUpload. `s3.upload(params).promise()` returns a Promise.

Updated original PR to now also support promises for AWS.Credentials and AWS.CredentialProviderChain, in addition to AWS.S3.ManagedUpload.

Resolves #1076 
Resolves #982 
Resolves #1121 

/cc: @chrisradek

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -503,14 +503,13 @@ AWS.Config = AWS.util.inherit({
     PromisesDependency = dep;
     var constructors = [AWS.Request, AWS.Credentials, AWS.CredentialProviderChain];
     if (AWS.S3 && AWS.S3.ManagedUpload) constructors.push(AWS.S3.ManagedUpload);
-    AWS.util.addPromises(AWS.Request, dep);
+    AWS.util.addPromises(constructors, dep);
   },
 
   /**
-   * Gets the promise dependency the SDK will use wherever Promises are returned.
+   * Gets the promise dependency set by `AWS.config.setPromisesDependency`.
    */
   getPromisesDependency: function getPromisesDependency() {
-    if (!PromisesDependency && typeof Promise === 'function') return Promise;
     return PromisesDependency;
   }
 });
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,6 +1,7 @@
 var AWS = require('./core');
 require('./credentials');
 require('./credentials/credential_provider_chain');
+var PromisesDependency;
 
 /**
  * The main configuration class used by all service objects to set
@@ -499,7 +500,18 @@ AWS.Config = AWS.util.inherit({
    * @param [Constructor] dep A reference to a Promise constructor
    */
   setPromisesDependency: function setPromisesDependency(dep) {
-    AWS.util.addPromisesToRequests(AWS.Request, dep);
+    PromisesDependency = dep;
+    var constructors = [AWS.Request, AWS.Credentials, AWS.CredentialProviderChain];
+    if (AWS.S3 && AWS.S3.ManagedUpload) constructors.push(AWS.S3.ManagedUpload);
+    AWS.util.addPromises(AWS.Request, dep);
+  },
+
+  /**
+   * Gets the promise dependency the SDK will use wherever Promises are returned.
+   */
+  getPromisesDependency: function getPromisesDependency() {
+    if (!PromisesDependency && typeof Promise === 'function') return Promise;
+    return PromisesDependency;
   }
 });
 
--- a/lib/credentials.js
+++ b/lib/credentials.js
@@ -148,3 +148,6 @@ AWS.Credentials = AWS.util.inherit({
     callback();
   }
 });
+
+AWS.util.addPromises(AWS.Credentials);
+AWS.util.addPromises(AWS.CredentialProviderChain);
--- a/lib/credentials.js
+++ b/lib/credentials.js
@@ -146,23 +146,23 @@ AWS.Credentials = AWS.util.inherit({
   refresh: function refresh(callback) {
     this.expired = false;
     callback();
-  },
-
-  /**
-   * @api private
-   */
-  addPromisesToClass: function addPromises(PromiseDependency) {
-    this.getPromise = AWS.util.promisifyMethod('get', PromiseDependency);
-    this.refreshPromise = AWS.util.promisifyMethod('refresh', PromiseDependency);
-  },
-
-  /**
-   * @api private
-   */
-  deletePromisesFromClass: function deletePromisesFromClass() {
-    delete this.getPromise;
-    delete this.refreshPromise;
   }
 });
+
+/**
+ * @api private
+ */
+AWS.Credentials.addPromisesToClass = function addPromisesToClass(PromiseDependency) {
+  this.prototype.getPromise = AWS.util.promisifyMethod('get', PromiseDependency);
+  this.prototype.refreshPromise = AWS.util.promisifyMethod('refresh', PromiseDependency);
+};
+
+/**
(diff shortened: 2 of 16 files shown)


lib/util.js
```diff
@@ -781,7 +781,8 @@ var util = {
         });
         self.runTo();
       });
-    }
+    };
+    constructor._PromiseDependency = PromiseDependency;
```
chrisradek: It seems messy to attach the Promise constructor to the Request constructor and then to reference that outside of the request class (in managed upload). 

This might be a good time to refactor the `util.setPromisesDependency` method to store the Promise constructor, and then add a `util.getPromisesDependency` method to retrieve it.

rightaway:
+1 I've tested locally and it works for me.

niftylettuce:
+1 please accept PR

lib/util.js
```diff
@@ -748,27 +748,74 @@ var util = {
...
+          });
         });
-        self.runTo();
-      });
+      };
+    };
+
+    for (var ind = 0; ind < constructors.length; ind++) {
+      var constructor = constructors[ind];
+      switch (constructor.name) {
```
chrisradek: It looks like `constructor.name` may not work in all the browsers we support and could have issues when minifiers are used:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#Browser_compatibility

Can you do an equality check against the constructor instead?

chrisradek: Actually, what do you think about giving each class that should return a promise a static method that accepts a Promise constructor, then the class can control how it should promisify itself?

The pros to that approach would be the logic for adding promises would be controlled by each class, rather than defined in a long if/switch statement within a utility function. If the method to promisify a class was named the same for all classes, you can just check if the constructor has that method then call it, instead of maintaining a list of enums to check a class against. You could still make the `promisifyMethod` a utility method if that reduces code duplication.

lib/config.js
```diff
@@ -499,7 +500,18 @@ AWS.Config = AWS.util.inherit({
    * @param [Constructor] dep A reference to a Promise constructor
    */
   setPromisesDependency: function setPromisesDependency(dep) {
-    AWS.util.addPromisesToRequests(AWS.Request, dep);
+    PromisesDependency = dep;
+    var constructors = [AWS.Request, AWS.Credentials, AWS.CredentialProviderChain];
+    if (AWS.S3 && AWS.S3.ManagedUpload) constructors.push(AWS.S3.ManagedUpload);
+    AWS.util.addPromises(AWS.Request, dep);
```
LiuJoyceC: Yes, thanks for catching that.
And for your question above about setting third-party promise dependencies when the environment supports native promises: all of the existing promisified methods will update to the set dependency because `setPromisesDependency` calls `AWS.util.addPromises` on the constructors (after I make the correction you mentioned above, of course). I've manually tested this. I can update one of the tests to be explicit about it.

lib/credentials.js
```diff
@@ -148,3 +148,6 @@ AWS.Credentials = AWS.util.inherit({
     callback();
   }
 });
+
+AWS.util.addPromises(AWS.Credentials);
+AWS.util.addPromises(AWS.CredentialProviderChain);
```
LiuJoyceC: Due to the order the modules are loaded, moving that line to the `credential_provider_chain.js` file would throw an error. However, I discovered this is due to the fact that I'm loading `config.js` from inside the `AWS.util.addPromises()` (which is needed if I want to get the promise dependency via `AWS.config.getPromisesDependency()`. I can refactor that so that I can move this line to the right file.

lib/util.js
LiuJoyceC: The reason I originally checked `constructor.name` is because I hadn't originally planned on requiring `core.js`, so I didn't have access to the actual constructors for an equality check before I added that require statement. I may be refactoring the require out anyway (as discussed above). I'll look into adding the static method for each class.

LiuJoyceC:
The PR has been refactored to move class-specific promisification to static methods on the classes themselves rather than in a switch-case statement. `util.addPromises` still exists (since there's still some common logic I'd rather not duplicate in the static methods of each class), but has been refactored to not require `config.js` anymore, which means I had to move the default promise dependency (the native Promise) determination to this method rather than `config.getPromisesDependency`. I prefer to keep only one source of truth for the default dependency, in case we ever want to change it (for example, if a future ECMAScript specification had a new backwards-incompatible `Promise2` or something like that). Not requiring `config.js` in the util method means the module loading order is no longer wonky, and the promisification of `CredentialProviderChain` can be done in the `credential_provider_chain.js` now.

lib/util.js
```diff
@@ -748,28 +748,44 @@ var util = {
...
-      delete constructor.prototype.promise;
-      return;
+    if (typeof PromiseDependency !== 'function') var deletePromises = true;
+    if (!Array.isArray(constructors)) constructors = [constructors];
+
+    for (var ind = 0; ind < constructors.length; ind++) {
+      var classPrototype = constructors[ind].prototype;
+      if (deletePromises) {
+        if (classPrototype.deletePromisesFromClass) {
+          classPrototype.deletePromisesFromClass();
```
chrisradek: This works, and is very close to what I was trying to describe before. Is there a reason you put the `deletePromisesFromClass` and `addPromisesToClass` on the prototype instead of on the constructor? This works, but the code within those methods is a little confusing, since the `this` context won't be an instance of the method as you'd typically expect, it would be the prototype itself.

LiuJoyceC: I can move these methods to the constructor.

lib/credentials.js
```diff
@@ -146,5 +146,23 @@ AWS.Credentials = AWS.util.inherit({
   refresh: function refresh(callback) {
     this.expired = false;
     callback();
+  },
+
+  /**
+   * @api private
+   */
+  addPromisesToClass: function addPromises(PromiseDependency) {
```
LiuJoyceC: Yup, I meant to change all of the function names to match their fields, but I missed that one.

lib/s3/managed_upload.js
```diff
@@ -215,7 +239,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
   validateBody: function validateBody() {
     var self = this;
     self.body = self.service.config.params.Body;
-    if (!self.body) throw new Error('params.Body is required');
+    if (self.body === undefined) throw new Error('params.Body is required');
```
chrisradek: Can you confirm this won't cause https://github.com/aws/aws-sdk-js/issues/570 to happen again? Looks like this change was made in the past, and caused a regression. Just want to make sure we don't cause the same regression again.

PR #1088, opened 2016-08-07
update verbiage in api docs

gswalden:
`GetObject` and `HeadObject` only read data, so changing the `Metadata` description to reflect that.

### DIFF
--- a/apis/s3-2006-03-01.normal.json
+++ b/apis/s3-2006-03-01.normal.json
@@ -2857,7 +2857,7 @@
         },
         "Metadata": {
           "shape": "Metadata",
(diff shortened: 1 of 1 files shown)


LiuJoyceC:
Hi @gswalden 

Thanks for the PR. Unfortunately, we cannot accept PRs that modify the `apis` directory, as those are maintained by service teams and are shared amongst all of our SDKs, not just the JavaScript SDK. If you would like, you can post on the AWS Forum for S3 to request the change: https://forums.aws.amazon.com/forum.jspa?forumID=24

I will close this PR as we cannot merge it. Thanks.

PR #1114, opened 2016-08-26
EC2MetadataCredentials and ECSCredentials retry

### LINKED ISSUE #692: `Missing credentials in config` happening intermittently
We've been having some difficulties in working that the SDK is intermittently unable to fetch credentials and this renders our application unauthorised. The ec2 where this is occurring has a particular IAM role and the SDK is therefore reaching out to the metadata endpoint (169.254...) to fetch it's keys. However, when it does so it occasionally appears to throw this type of error: 

So, for example this dynamoDB was logged by our application with an SDK error: 

```
{
    "error": {
        "message": "Missing credentials in config",
        "code": "CredentialsError",
        "errno": "ECONNREFUSED",
        "syscall": "connect",
        "time": "2015-07-15T21:55:06.083Z",
        "originalError": {
            "message": "Could not load credentials from EC2MetadataCredentials",
            "code": "CredentialsError",
            "errno": "ECONNREFUSED",
            "syscall": "connect",
            "time": "2015-07-15T21:55:06.083Z",
            "originalError": {
                "code": "ECONNREFUSED",
                "errno": "ECONNREFUSED",
                "syscall": "connect",
                "message": "connect ECONNREFUSED"
            }
        }
    },
    "level": "error",
    "message": "DynamoDB Query failed",
    "timestamp": "2015-07-15T21:55:06.087Z"
}
```

More recently, this S3 call had this similar error: 

```
...
    "originalError": {
      "message": "Could not load credentials from any providers",
      "code": "CredentialsError",
      "errno": "ECONNREFUSED",
      "syscall": "connect",
      "address": "169.254.169.254",
      "port": 80,
      "time": "2015-08-26T06:08:18.008Z",
      "originalError": {
        "code": "ECONNREFUSED",
        "errno": "ECONNREFUSED",
        "syscall": "connect",
        "address": "169.254.169.254",
        "port": 80,
        "message": "connect ECONNREFUSED 169.254.169.254:80"
      }
    }
...
```

We've experienced the problem with multiple applications intermittently, but as frequently as half a dozen times per day on a single ec2. We're using NodeJS aws-sdk version 2.1.46 in the example above and iojs 2.3.1 here, nodeJS 0.12.x elsewhere. We're in the `ap-southeast-2` region. 

While it would appear that the connection's being refused, I'd be surprised to see this endpoint actually go down. Is it possible we're doing something stupid with node to create this, or else possibly there be a genuine issue?

LiuJoyceC:
Adds retry to sourcing credentials for IAM Roles in the EC2 Metadata Service or in ECS. By default will retry up to 3 times, with exponential backoff. Also for ECSCredentials, this PR will allow one asynchronous request to serve multiple refresh calls, so that at most only one request will be in flight at any point. Resolves #692 

/cc: @chrisradek

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -100,7 +100,7 @@ require('./credentials/credential_provider_chain');
  *     Currently supported options are:
  *
  *     * **base** [Integer] &mdash; The base number of milliseconds to use in the
- *       exponential backoff for operation retries. Defaults to 100 ms.
+ *       exponential backoff for operation retries. Defaults to 30 ms.
  *     * **customBackoff ** [function] &mdash; A custom function that accepts a retry count
  *       and returns the amount of time to delay in milliseconds. The `base` option will be
  *       ignored if this option is supplied.
--- a/lib/config.js
+++ b/lib/config.js
@@ -100,7 +100,7 @@ require('./credentials/credential_provider_chain');
  *     Currently supported options are:
  *
  *     * **base** [Integer] &mdash; The base number of milliseconds to use in the
- *       exponential backoff for operation retries. Defaults to 30 ms.
+ *       exponential backoff for operation retries. Defaults to 100 ms.
  *     * **customBackoff ** [function] &mdash; A custom function that accepts a retry count
  *       and returns the amount of time to delay in milliseconds. The `base` option will be
  *       ignored if this option is supplied.
--- a/lib/credentials/ec2_metadata_credentials.js
+++ b/lib/credentials/ec2_metadata_credentials.js
@@ -18,7 +18,7 @@ require('../metadata_service');
  * AWS.config.credentials = new AWS.EC2MetadataCredentials({
  *   httpOptions: { timeout: 5000 }, // 5 second timeout
  *   maxRetries: 10, // retry 10 times
- *   retryDelayOptions: { base: 30 } // see AWS.Config for information
+ *   retryDelayOptions: { base: 200 } // see AWS.Config for information
  * });
  * ```
  *
--- a/lib/credentials/ec2_metadata_credentials.js
+++ b/lib/credentials/ec2_metadata_credentials.js
@@ -9,15 +9,20 @@ require('../metadata_service');
  * can connect, and credentials are available, these will be used with zero
  * configuration.
  *
- * This credentials class will timeout after 1 second of inactivity by default.
+ * This credentials class will by default timeout after 1 second of inactivity
+ * and retry 3 times.
  * If your requests to the EC2 metadata service are timing out, you can increase
- * the value by configuring them directly:
+ * these values by configuring them directly:
  *
  * ```javascript
  * AWS.config.credentials = new AWS.EC2MetadataCredentials({
- *   httpOptions: { timeout: 5000 } // 5 second timeout
+ *   httpOptions: { timeout: 5000 }, // 5 second timeout
+ *   maxRetries: 10, // retry 10 times
+ *   retryDelayOptions: { base: 30 } // see AWS.Config for information
  * });
  * ```
+ *
+ * @see AWS.Config.retryDelayOptions
  *
  * @!macro nobrowser
  */
@@ -26,6 +31,8 @@ AWS.EC2MetadataCredentials = AWS.util.inherit(AWS.Credentials, {
     AWS.Credentials.call(this);
 
     options = options ? AWS.util.copy(options) : {};
+    options = AWS.util.merge(
+      {maxRetries: this.defaultMaxRetries}, options);
     if (!options.httpOptions) options.httpOptions = {};
     options.httpOptions = AWS.util.merge(
       {timeout: this.defaultTimeout}, options.httpOptions);
@@ -38,6 +45,11 @@ AWS.EC2MetadataCredentials = AWS.util.inherit(AWS.Credentials, {
    * @api private
    */
   defaultTimeout: 1000,
+
+  /**
+   * @api private
+   */
+  defaultMaxRetries: 3,
 
   /**
    * Loads the credentials from the instance metadata service
--- a/lib/credentials/ecs_credentials.js
+++ b/lib/credentials/ecs_credentials.js
@@ -8,15 +8,20 @@ var AWS = require('../core');
  * in the container. If valid credentials are returned in the response, these
  * will be used with zero configuration.
  *
- * This credentials class will timeout after 1 second of inactivity by default.
+ * This credentials class will by default timeout after 1 second of inactivity
+ * and retry 3 times.
  * If your requests to the relative URI are timing out, you can increase
  * the value by configuring them directly:
  *
  * ```javascript
  * AWS.config.credentials = new AWS.ECSCredentials({
- *   httpOptions: { timeout: 5000 } // 5 second timeout
+ *   httpOptions: { timeout: 5000 }, // 5 second timeout
+ *   maxRetries: 10, // retry 10 times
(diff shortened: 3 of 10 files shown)


lib/config.js
```diff
@@ -100,7 +100,7 @@ require('./credentials/credential_provider_chain');
  *     Currently supported options are:
  *
  *     * **base** [Integer] &mdash; The base number of milliseconds to use in the
- *       exponential backoff for operation retries. Defaults to 100 ms.
+ *       exponential backoff for operation retries. Defaults to 30 ms.
```
chrisradek: The default used to be 30 ms but was changed to 100 ms with the configurable retry delays.

lib/credentials/ec2_metadata_credentials.js
```diff
@@ -9,23 +9,30 @@ require('../metadata_service');
...
  * If your requests to the EC2 metadata service are timing out, you can increase
- * the value by configuring them directly:
+ * these values by configuring them directly:
  *
  * ```javascript
  * AWS.config.credentials = new AWS.EC2MetadataCredentials({
- *   httpOptions: { timeout: 5000 } // 5 second timeout
+ *   httpOptions: { timeout: 5000 }, // 5 second timeout
+ *   maxRetries: 10, // retry 10 times
+ *   retryDelayOptions: { base: 30 } // see AWS.Config for information
```
chrisradek: I know it's an example, but we should probably show increasing the base retry delay above 100 (current default). That puts it more in line with the comment that says you can increase these values on line 14.

lib/credentials/ecs_credentials.js
```diff
@@ -69,23 +79,18 @@ AWS.ECSCredentials = AWS.util.inherit(AWS.Credentials, {
...
-        httpResponse.on('end', function() { callback(null, data); });
-      }, callback);
-    });
+    AWS.util.handleRequestWithTimeoutRetries(httpRequest, this, callback);
   },
 
   /**
+   * @api private
+   */
+  refreshQueue: [],
```
chrisradek: Why is the refresh queue only implemented for ECS credentials and not EC2 credentials? EC2 credentials has the same issue.

LiuJoyceC: EC2MetdataCredentials already had this feature, through the MetadataService. The `refreshQueue` on the ECSCredentials is analogous to the `loadCredentialsCallbacks` on the MetadataService. They're structured slightly differently but work the same way. (`refreshQueue` has to retain references to the instances of ECSCredentials because the data isn't given to the callback the way it is MetadataService).

lib/credentials/ecs_credentials.js
```diff
@@ -119,13 +139,15 @@ AWS.ECSCredentials = AWS.util.inherit(AWS.Credentials, {
...
-          var creds = JSON.parse(data);
-          if (self.credsFormatIsValid(creds)) {
-            self.expired = false;
-            self.accessKeyId = creds.AccessKeyId;
-            self.secretAccessKey = creds.SecretAccessKey;
-            self.sessionToken = creds.Token;
-            self.expireTime = new Date(creds.Expiration);
+          data = JSON.parse(data);
+          if (self.credsFormatIsValid(data)) {
+            var creds = {
```
chrisradek: I'm not 100% certain since I haven't tested this myself, but if these fields aren't assigned to `self`, will the next batch of requests cause the credentials to be refreshed again?

LiuJoyceC: Assigning these fields to `self` is the only way the credentials can be accessed, since the callbacks for `refresh` are error-only callbacks, and when the callback returns with no error, the credentials are expected to have been loaded on to the instance of the credential provider itself. This is how it worked before on the ECSCredentials and how it works with the other credential providers. When a service request is made, the SDK checks directly on the credential provider for these fields.

lib/credentials/ecs_credentials.js
chrisradek: Ah, so EC2 already supported this, excellent.

lib/credentials/ecs_credentials.js
LiuJoyceC: I realized I misunderstood your question. It is indeed necessary to assign these fields to `self`, I've just simply moved them up into the `callbacks` function, and I'm passing this `creds` object containing all of the needed fields into `callbacks`. On line 118 above (`if (!err) AWS.util.update(call.provider, creds);`) is where these fields are getting assigned to the credential provider instance. The reason I can no longer use `self` is because if there are multiple callbacks in the `refreshQueue` coming from multiple instances of the ECS credential provider, only the first `refresh()` call will actually call `this.request()`, so `self` will only refer to that one instance of the credential provider. The `call.provider` in the `refreshQueue` is a reference to the instance of the credential provider where the corresponding callback in the `refreshQueue` is from. Before, assigning the fields directly to `self` was not a problem because every single `refresh()` call would invoke `'this.request()` and the right instance of the credential provider would be referenced.

lib/util.js
```diff
@@ -793,6 +793,51 @@ var util = {
...
+   */
+  handleRequestWithTimeoutRetries: function handleRequestWithTimeoutRetries(httpRequest, options, cb) {
+    if (!options) options = {};
+    var http = AWS.HttpClient.getInstance();
+    var httpOptions = options.httpOptions || {};
+    var retryCount = 0;
+
+    var errCallback = function(err) {
+      var maxRetries = options.maxRetries || 0;
+      if (err && err.code === 'TimeoutError' && retryCount < maxRetries) {
```
chrisradek: Have you considered retrying for all the codes we retry when a service requests fails? `TimeoutError` is thrown by the SDK when a connection times out. I'm not sure if there are other errors the metadata service may also throw, but it shouldn't hurt to check.

LiuJoyceC: I've updated the PR to retry for 5xx status codes, as well as for 4xx status codes only if there is a retry-after header. I was able to reliably replicate the Metadata Service returning 429 (Too Many Requests) errors, which also returned a retry-after header which give a minimum number of seconds to wait before retrying.

lib/util.js
```diff
@@ -793,6 +793,64 @@ var util = {
...
+    var sendRequest = function() {
+      var data = '';
+      http.handleRequest(httpRequest, httpOptions, function(httpResponse) {
+        httpResponse.on('data', function(chunk) { data += chunk.toString(); });
+        httpResponse.on('end', function() {
+          var statusCode = httpResponse.statusCode;
+          if (statusCode < 300) {
+            cb(null, data);
+          } else {
+            var retryAfter = parseInt(httpResponse.headers['retry-after']) * 1000 || 0;
```
chrisradek: Minor thing, but might want to specify that the int is a decimal number when parsing `parseInt(str, 10)` for a little extra safety.

lib/util.js
```diff
@@ -793,6 +793,64 @@ var util = {
...
+            var err = util.error(new Error(),
+              { retryable: !!(statusCode >= 500 || (statusCode >= 400 && retryAfter)) }
```
chrisradek: We do retry for all 500+ status codes, but not 400+. Some of these we shouldn't retry, like `403` (not authorized) or `404` (not found). You're right that we should retry `429` (nice catch!), but this seems a little too inclusive, though the `retryAfter` check probably causes the correct behavior in most cases.

LiuJoyceC: I can restrict it to just 429

lib/util.js
LiuJoyceC: Good catch, I'll modify it.

chrisradek:
:shipit:

Issue #603, opened 2015-05-13
Webpack support


anthonator:
Has anybody gotten this library to work with webpack? The webpack team recently fixed a [ternary require issue](https://github.com/webpack/webpack/issues/627) which fixed a show stopper.

I was able to fix an issue with loading a JSON file with the `json-loader`. Now I'm getting:

```
WARNING in ./~/aws-sdk/lib/util.js
Critical dependencies:
40:30-45 the request of a dependency is an expression
43:11-53 the request of a dependency is an expression
 @ ./~/aws-sdk/lib/util.js 40:30-45 43:11-53

WARNING in ./~/aws-sdk/lib/api_loader.js
Critical dependencies:
13:15-59 the request of a dependency is an expression
104:12-46 the request of a dependency is an expression
108:21-58 the request of a dependency is an expression
114:18-52 the request of a dependency is an expression
```

Which is causing the individual services to not be included in the packaged bundle.

So doing `new AWS.S3()` returns a `is not a function error`.

It looks like you might be able to fix this with the [`ContextReplacementPlugin`](https://github.com/webpack/docs/wiki/list-of-plugins#contextreplacementplugin) but I'm not sure how to configure it properly for this scenario.

Does anybody have any examples or anything?

lsegal:
I think in this case a site like [StackOverflow](http://stackoverflow.com/) might be a better venue to ask this kind of question. The SDK unfortunately does not currently support Webpack, since we currently have our tooling working with browserify. It seems like this specific issue might be a lot more complicated to solve based on the error, but someone who knows more about the webpack tool may be of more assistance-- it might not be as complicated as it seems. If you do resolve this we'd be happy to hear how you were able to do it!

AdityaManohar:
I'm marking this as closed since this isn't an SDK issue. Feel free to open another issue if you have other questions.

jblack10101:
I'm having this same issue. Was a stack overflow ever created?

geminiyellow:
+1 i want webpack support too.

```
WARNING in ./~/aws-sdk/lib/util.js
Critical dependencies:
40:30-45 the request of a dependency is an expression
43:11-53 the request of a dependency is an expression
 @ ./~/aws-sdk/lib/util.js 40:30-45 43:11-53

WARNING in ./~/aws-sdk/lib ^\.\/.*$
Module not found: Error: Cannot resolve directory '.' in /Users/me/Documents/Sources/my-project/client/node_modules/aws-sdk/lib
[...]

Issue #696, opened 2015-08-28
browserify not working for node apps


doapp-ryanp:
Trying to browserify my node code for use in lambda to help combat the cold start issues (which also raises the ceiling on the hard 50MB zipped limit).

Is [browserfying](http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/browser-building.html) for node not supported?  Ultamately I'd like to just use the `--node` browserify option, but its not working.

Ex: (I wont be hard coding creds, this is just for ease of reproduce.)
```javascript
var AWS = require('aws-sdk');
AWS.config.update({
  accessKeyId: 'mykey',
  secretAccessKey: 'mysecret'
});
var s3 = new AWS.S3();
s3.listBuckets(function(err, data) {
  console.log(err, data);
});
```

If I run `browserify bundletests.js > bundled.js` then run `node bundle.js` I get `XMLHttpRequest is not defined` which is expected since I'm not running in the browser.

So I run `browserify bundletests.js --dg --no-builtins --no-commondir > bundled.js` to not use the builtins and I get 
```
module.js:340
    throw err;
          ^
Error: Cannot find module '_process'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at s (/Users/ryan/projects/JAWS/tests/bundle/testapp/myLambda/bundled.js:1:176)
```

Any ideas?

AdityaManohar:
@doapp-ryanp I'm still a little unclear on how you intend to use your browserify-ed code inside of AWS Lambda. Lambda requires the function to export a top level CommonJS module. 

That being said, browserify-ing the SDK for use in Node.js is currently not supported, because we rely on the `window` object to hang the `AWS` namespace on. There might be other potential issues like the one you are running into. 

Would you be able to explain a use case a little more? I'm curious about why uploading an archive of Node.js modules is not working well for you.

doapp-ryanp:
@AdityaManohar I have broswerfied code that is working just fine in lambda.  You can export top level CommonJS method.  The bummer is the only node module I have not gotten to work so far is aws-sdk.  

If your curious on how I'm doing this, take a look at the [JAWS code](https://github.com/jaws-stack/JAWS/blob/v1.0-deploy-lambda/lib/commands/deploy_lambda.js#L204) I wrote.  

[...]

PR #1117, opened 2016-08-30
[WIP] Webpack support

chrisradek:
/cc @LiuJoyceC 

This PR adds support for webpack out of the box, and maintains browserify support.
This will allow webpack to pull in the SDK with all default services, assuming no custom webpack config. There is a way to specify services with custom config but it is still a bit messy, I want to clean that up.

Tested with a personalized starter app as well as a create-react-app project.

Still a WIP. Need to test in IE, add documentation, and update tooling.

Address #603


lib/browser.js
```diff
@@ -16,6 +24,15 @@ AWS.XML.Parser = require('./xml/browser_parser');
 // Load the XHR HttpClient
 require('./http/xhr');
 
+if (typeof process === 'undefined') {
+  process = {
+    browser: true
+  };
```
simonbuchan: Should be `this.process = ...`?

simonbuchan:
Nice!

Initial Qs:
- Can `dist-tools/service-loader` reuse the existing `dist-tools`? Seems we want to avoid duplication, and if I'm using those, I expect others are too, so they can't be removed.
- Could `/lib/browser_services` be something like `/svc/all` and require all `/svc/{id}`?

simonbuchan:
To be a bit clearer on the last point, to do something like [what core.js do to support partial use](https://github.com/zloirock/core-js#commonjs), so you can use only `import S3 from 'aws-sdk/svc/s3';` and get *just* S3 in your build:

```
// aws-sdk/svc/s3.js
// generated by aws-sdk/dist-tools/service-loader.js
var AWS = require('../lib/aws'); // will be '../lib/browser.js' in browserify, webpack, etc...?

AWS.apiLoader.services['s3'] = {};
AWS.S3 = AWS.Service.defineService('s3', [ '2006-03-01' ]);
require('./services/s3');
AWS.apiLoader.services['s3']['2006-03-01'] = require('../apis/s3-2006-03-01.min');
AWS.apiLoader.services['s3']['2006-03-01'].paginators = require('../apis/s3-2006-03-01.paginators').pagination;
AWS.apiLoader.services['s3']['2006-03-01'].waiters = require('../apis/s3-2006-03-01.waiters2').waiters;

module.exports = AWS.S3;
```

It... looks like it should work?

chrisradek:
@simonbuchan 
Thanks for posting!
Yes, I should be able to merge `service-loader` into `service-collector`.

I think we can do something similar to your example. I'll take a look into that today.

lib/browser.js
```diff
@@ -1,3 +1,11 @@
+var util = require('./util');
```
AdityaManohar: @chrisradek Do you think it might be worthwhile adding a buffer polyfill as well?

chrisradek: Absolutely. Webpack actually does polyfill buffer, but I like taking control of which buffer polyfill to use back to us (makes updating when bugs are fixed easier), and to make supporting other tools easier.

dist-tools/service-loader.js
```diff
@@ -0,0 +1,169 @@
...
+    svc = new ClassName({apiVersion: version, endpoint: 'localhost'});
+    api = apis.load(service, svc.api.apiVersion);
+  } catch (e) {
+    return null;
+  }
+
+  var serviceFileName = metadata[service].prefix || service;
+  var lines = [];
+  var line = util.format(
+    'AWS.apiLoader.services[\'%s\'][\'%s\'] = %s;\n',
```
AdityaManohar: Curious why we can't just rely on browserify pulling in deps?

chrisradek: I'm working on rewriting this script to better support custom builds, so a lot will probably change here.

I'm not quite sure what you mean. Are you asking why we can't rely on browserify to pull in the service clients, or all the node.js module deps as well? Part of my motivation was so that webpack users wouldn't need browserify as well to customize their build.

chrisradek:
I added some changes to the PR.

With these changes, users can require individual services and webpack/browserify will only pull those in.

In the user's code, they would do something like this to require S3:
```javascript
var S3 = require('aws-sdk/browser/s3');
```
Access to the `AWS` namespace (useful for setting config across multiple service clients) is available like so:
```javascript
var AWS = require('aws-sdk/browser/global');
```

`aws-sdk/browser/all` can be required to get all services.

Of course, calling `var AWS = require('aws-sdk')` will continue to work as it does today, returning the `AWS` namespace with the default service clients on it.

Today I'll be working on some documentation and some more testing. The PR shows a lot of file changes but this is largely due to generating the individual services so that they can be required individually.

AdityaManohar:
Can this be changed to just

```javascript
var S3 = require('aws-sdk/s3');
var AWS = require('aws-sdk/global');
```
Why do we need the browser prefix? Browserifying for Node seems to be a legitimate use case. See #696

chrisradek:
@AdityaManohar 
The way I have things currently set up, the stubs in `aws-sdk/browser/*` will also build the browser version of `AWS`, which means it'll be using JS crypto and buffer libraries, as well as xhr instead of https. I did this so that requiring a single service gives you everything you need (including the core of the SDK).

I chose the `browser` prefix in case we wanted to do the same for node.js. I don't believe dynamically choosing which `AWS` to load will work either, since tools will just look at what's required and pull that in, regardless of if it's actually used or not.

I'll have to look into that issue some more to see if what's here would work for that use case as well.

AdityaManohar:
@chrisradek 
Let me illustrate the use case I'm talking about:

```js
// main.js in my npm module


// In the past I would do this
// var AWS = require('aws-sdk');

// This is the new proposal
var S3 = require('aws-sdk/browser/s3');

var headS3Object = function() {
  var s3 = new AWS.S3({region: 'us-west-2'});
  s3.headObject({/*params*/}, callback);
}

module.exports = headS3Object;
```
Using browserify:

```sh
$> browserify main.js -o bundle.js
```
This is use case that currently works with browserify, except for the selective service module inclusions.

Now I could also chose to `main.js` as a node module. 

From what you are saying, my understanding is that I cannot just take `main.js` as use it as a Node module because I'm specifically importing browser versions of the the service modules. Is that understanding correct?

chrisradek:
@AdityaManohar 
That's correct. Essentially, the browser entry point would be loaded when requiring the service.

I might be able to create a node.js copy of the service stubs, and then in the package.json browser field map the node.js stubs to the browser ones. Then, if the user is using a tool that looks at the browser field (like webpack and browserify), the browser stubs would be used by default.

I'm open to any ideas on how else to accomplish this (besides having separate packages), and I can at least go down the above route to see how feasible that is.

dist-tools/create-all-services.js
```diff
@@ -0,0 +1,161 @@
+var fs = require('fs');
+var path = require('path');
+
+var defaultServices = 'acm,apigateway,applicationautoscaling,autoscaling,cloudformation,cloudfront,cloudhsm,cloudtrail,cloudwatch,cloudwatchlogs,cloudwatchevents,codecommit,codedeploy,codepipeline,cognitoidentity,cognitoidentityserviceprovider,cognitosync,configservice,devicefarm,directconnect,dynamodb,dynamodbstreams,ec2,ecr,ecs,elasticache,elasticbeanstalk,elastictranscoder,elb,emr,firehose,gamelift,inspector,iotdata,kinesis,kms,lambda,marketplacecommerceanalytics,mobileanalytics,machinelearning,opsworks,rds,redshift,route53,route53domains,s3,ses,sns,sqs,ssm,storagegateway,sts,waf';
```
LiuJoyceC: Instead of maintaining a separate list (I know this is what we currently do anyway), we would have fewer places to worry about making manual changes for service updates if we move this information into metadata.json, e.g.
```
{
    "acm": {
        "name": "ACM",
        "defaultService": true
    },
    ...
}
```
That way we can also get rid of the identical identical defaultServices lists in service-collector.js as well.

chrisradek: Good point, I like that idea.

dist-tools/create-all-services.js
```diff
@@ -0,0 +1,161 @@
...
+  this._apiFileNames = null;
+}
+
+ClientCreator.prototype.getAllApiFilenames = function getAllApiFilenames() {
+    if (this._apiFileNames) {
+        return this._apiFileNames;
+    }
+    var apiFileNames = fs.readdirSync(this._apisFolderPath);
+    // filter out metadata
+    this._apiFileNames = apiFileNames.filter(function(name) {
```
LiuJoyceC: Just a minor comment, but it never made much sense to me to have `metadata.json` in the `apis/` directory where service models were. It was always the oddball in that directory, and always required extra logic (like this bit of code here) to exclude it from being counted as a service. It makes sense to me to move `metadata.json` to the root of `lib/`, and would allow us to get rid of this extra filter.

chrisradek: Sure, I think that's reasonable, but we'll need to be extra careful to update all of our tooling and make sure we update all references.

LiuJoyceC: Sure, and it's not important for it be part of this PR. I noticed this filter function, and this isn't the first time it has crossed my mind that perhaps `metadata.json` should be moved, so I just decided to mention it.

AdityaManohar:
>I'm open to any ideas on how else to accomplish this (besides having separate packages), and I can at least go down the above route to see how feasible that is.

I'll try to send you something soon. 

IMO, it is a better user experience to not have to make the distinction between browser bundles and Node bundles upfront. The existing tooling ecosystem (Webpack, Browserify) largely supports writing once and running anywhere, and it would be nice if the SDK could support this out of the box as well.

chrisradek:
Closing in favor of #1123

PR #1123, opened 2016-09-08
Webpack Support

chrisradek:
This PR allows the SDK to be used with webpack.
It also allows webpack and browserify to generate node bundles that include the SDK.

Services can also now be required individually. For example, you can require the S3 service in your code by doing the following:
```javascript
var S3 = require('aws-sdk/clients/s3');
```
To get access to the AWS namespace (useful when sharing config across multiple services), you can do the following:
```javascript
var AWS = require('aws-sdk/global');
```
Of course, you can continue to use the SDK as usual:
```javascript
var AWS = require('aws-sdk');
var S3 = AWS.S3;
```

Notes:
For webpack, you will still need to add a json loader to your configuration.


clients/acm.js
```diff
@@ -0,0 +1,14 @@
+require('../lib/node_loader');
+var AWS = require('../lib/core');
+var Service = require('../lib/service');
```
LiuJoyceC: `core.js` already requires `service.js`, and the `Service` constructor can already be accessed at `AWS.Service`. Why does it need to be required here? We also wouldn't need to export `Service` in `service.js`

clients/acm.js
```diff
@@ -0,0 +1,14 @@
...
+var apiLoader = require('../lib/api_loader');
```
LiuJoyceC: Similarly here, `node_loader.js` and `browser_loader.js` already require `api_loader.js`, and `apiLoader` can already be accessed at `AWS.apiLoader`. `apiLoader` would also not need to be exported from `api_loader.js`.

dist-tools/service-collector.js
```diff
@@ -83,6 +102,7 @@ function ServiceCollector(services) {
     if (!builtServices[service]) {
       builtServices[service] = {};
 
+      //lines = clientCreator.generateClientStub(metadata[service], version);
```
LiuJoyceC: Delete

dist-tools/client-creator.js
```diff
@@ -0,0 +1,223 @@
...
+    models.versions = versions;
+    return models;
+};
+
+ClientCreator.prototype.customizationsExist = function customizationsExist(serviceName) {
+    var customizationsFolder = this._serviceCustomizationsFolderPath;
+    return fs.existsSync(path.join(customizationsFolder, serviceName + '.js'));
+};
+
+ClientCreator.prototype.generateClientStub = function generateClientStub(serviceMetadata, specifiedVersion) {
```
LiuJoyceC: Since this function isn't used anymore, it can be deleted.

lib/api_loader.js
```diff
@@ -1,139 +1,9 @@
...
-}
-
-module.exports = {
-  serviceVersions: serviceVersions,
-  serviceName: serviceName,
-  serviceIdentifier: serviceIdentifier,
-  serviceFile: serviceFile,
-  load: load
+AWS.apiLoader = function(svc, version) {
+  return AWS.apiLoader.services[svc][version];
```
LiuJoyceC: Is it possible that the `AWS.apiLoader` may be called before the api is actually loaded and stored on `AWS.apiLoader.services`? If that happens, this would throw an error (eg "Cannot read property '11-30-2014' of undefined"). We could either throw a more helpful error (such as "S3 service not yet loaded. Require './clients/s3 first"), or better yet, require the file for them to load the service (since that's what the original `AWS.apiLoader` does for you anyway, and although I don't think this is a public method, adding that would allow us not to break this method).

LiuJoyceC:
:shipit: pending changes

chrisradek:
Changes made, merging!

simonbuchan:
Looks good! I hit an issue where using the non-client public APIs like default config or `CognitoIdentityCredentials` without pulling in all the default services has to import from `aws-sdk/lib/core`, so it may be a good idea to mention that somewhere:

```js
import { config, CognitoIdentityCredentials } from 'aws-sdk/lib/core';
```

simonbuchan:
Also the changes to `dist-tools/service-collector` broke my older hacky build that was munging and injecting the output of that into my entry, due to the missing `.json`, but probably not a big use case :)

simonbuchan:
Damn, just noticed that if you load `aws-sdk/lib/core` before any clients, then you need to first import `aws-sdk/lib/node_loader`, which is ugly.

chrisradek:
@simonbuchan 
You can get the global AWS object from `aws-sdk/global`. If you do that, you should be able to set config on the AWS object, and any services instantiated after that will use that configuration.

simonbuchan:
That is much better, although it still also adds it to the global which is
a little bad mannered

kellyjanderson:
@chrisradek As this change has been merged, can information about using with Webpack be added to the documentation?

https://github.com/aws/aws-sdk-js/blob/master/doc-src/guide/browser-intro.md

chrisradek:
@kellyjanderson 
The guide in this repo is out of date. A new guide has been published, with information about using webpack here:
http://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/webpack.html

The updated link to the new guide is in our readme:
https://github.com/aws/aws-sdk-js#usage-and-getting-started

The new guide is maintained separately from this repo, but we can remove the old guide to avoid confusion.

PR #1125, opened 2016-09-09
Lazy loads services when requiring the full SDK. Fixes #1124

### LINKED ISSUE #1124: Requiring the SDK now loads all services into memory
It looks like changes to the API loader in 2.6.0 now cause `require('aws-sdk')` to load all service definitions into memory, including definitions for previous (typically unused) service versions, which increases the memory footprint of an app by roughly 20+MB (on my machine):

```js
console.log('mem before', process.memoryUsage());
var AWS = require('aws-sdk');
console.log('mem after', process.memoryUsage());

// all services are loaded into API loader:
var l = require('aws-sdk/lib/api_loader');
var e = l.services.ec2;
if (e) {
  var numOps = Object.keys(e[Object.keys(e)[0]].operations).length;
  console.log('EC2 Operations', numOps);
} else {
  console.log('EC2 not yet loaded');
}
```

In `aws-sdk@2.5.x`:

```
mem before { rss: 19165184, heapTotal: 10481664, heapUsed: 4117368 }
mem after { rss: 29241344, heapTotal: 18870272, heapUsed: 9707864 }
EC2 not yet loaded
```

In `aws-sdk@2.6.0`:

```
mem before { rss: 19111936, heapTotal: 10481664, heapUsed: 4116632 }
mem after { rss: 42848256, heapTotal: 34598912, heapUsed: 16902912 }
EC2 Operations 210
```

Note the almost doubling of memory usage from previous releases.

Load time is also significantly increased:

```
$ time node -e 'a=require("aws-sdk");console.log(a.VERSION)'
2.5.6

real	0m0.208s
user	0m0.174s
sys	0m0.034s

$ time node -e 'a=require("aws-sdk");console.log(a.VERSION)'
2.6.0

real	0m0.300s
user	0m0.260s
sys	0m0.047s
```

This represents a startup time increase of over 50%.

chrisradek:
Service clients will now be lazy loaded when importing the SDK using: `require('aws-sdk')`
This fixes #1124 which noted that all services were loaded into memory when the SDK was imported, as a result of the changes to support webpack introduced in version `2.6.0` of the SDK.

Notes:
`doc-custom` was added to contain documentation for service customizations.
Service customizations now accept their Service constructor rather than grabbing it from the AWS namespace. Doing this allowed these customizations to be lazy-loaded as well, and doesn't require the service clients to exist on the AWS namespace anymore.

/cc @LiuJoyceC

### DIFF
--- a/.changes/next-release/bugfix-SDK-a6e681af.json
+++ b/.changes/next-release/bugfix-SDK-a6e681af.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "SDK",
+  "description": "Fixes an issue that caused all services to be loaded into memory when requiring the SDK. This issue was introduced in version `2.6.0` of the SDK, and address #1124."
+}
--- a/.npmignore
+++ b/.npmignore
@@ -8,6 +8,7 @@ configuration.sample
 coverage
 doc
 doc-src
+doc-custom
 eslint-rules
 Gemfile
 Gemfile.lock
--- a/.yardopts
+++ b/.yardopts
@@ -15,6 +15,7 @@
 --mixin-module-expr (AWS\.util\.)?mixin
 lib/core.js
 lib/**/*.js
+doc-custom/**/*.js
 -
 README.md
 UPGRADING.md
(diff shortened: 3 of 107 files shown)


chrisradek:
Closing in favor of #1126

PR #1130, opened 2016-09-14
Fixes an issue when https.globalAgent is set to false.

### LINKED ISSUE #1122: ec2 describeInstances silently fails when globalAgent is disabled
The following code silently fails with no explanation. This happens on every version I've tried including 2.5.5

```
require('https').globalAgent = false;

aws = require('aws-sdk');
ec2 = new aws.EC2({region: 'us-east-1'});
ec2.describeInstances({}, console.log);
```

chrisradek:
Fixes #1122

/cc: @LiuJoyceC


LiuJoyceC:
🚢

PR #1140, opened 2016-09-15
Fixes bugs in documentation tool

LiuJoyceC:
/cc: @chrisradek


LiuJoyceC:
The waiters are already pulling in the specific version (see line 259 of plugin.rb). The paginators.json files aren't used in our docs (instead, we rely on documentation from service teams about their service-specific paginator-related params (such as NextToken, MaxKeys, etc.))

chrisradek:
Gotcha, thanks for the info!
:shipit:

PR #1141, opened 2016-09-16
Fix intermittently failing unit test and allows running individual unit test files

LiuJoyceC:
Increases timeout on unit test that intermittently failed due to server timeout.Also adds an npm script to allow running only specified unit test files. To use it, simply run `npm run testfiles [names of unit tests files/directories separated by space]`. For example, running `npm run testfiles test/request.spec.coffee test/services` will run only the unit tests in the test/request.spec.coffee file and the toplevel files in the directory test/services. Works only with NPM version 2 or higher.

/cc: @chrisradek

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
       "buildertest": "mocha --compilers coffee:coffee-script -s 1000 -t 10000 dist-tools/test",
       "integration": "cucumber.js",
       "lint" : "eslint lib dist-tools/*.js",
-      "console": "./scripts/console"
+      "console": "./scripts/console",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -72,6 +72,7 @@
       "buildertest": "mocha --compilers coffee:coffee-script -s 1000 -t 10000 dist-tools/test",
       "integration": "cucumber.js",
       "lint" : "eslint lib dist-tools/*.js",
-      "console": "./scripts/console"
+      "console": "./scripts/console",
+      "testfiles" : "istanbul `[ $COVERAGE ] && echo 'cover _mocha' || echo 'test mocha'`"
```
chrisradek: I think you added an extra script into package.json.

chrisradek:
:shipit:

PR #1144, opened 2016-09-16
Fixes issue with service-bound global config

### LINKED ISSUE #1139: Possible breaking change about service configuration
This used to work until 2.5.6:

```js
AWS.config.update({dynamodb: {endpoint: '...'}});
```

With 2.6.2, this change was required:

```js
AWS.config.update({dynamodb: {endpoint: '...'}}, /* allowUnknownKeys */ true);
```

Was this intended? It doesn't figure on CHANGELOG.

chrisradek:
Fixes issue where passing service specific configuration into AWS.config.update would silently fail by default unless allowUnknownKeys was specified.

Previously, service clients were only built when the client (i.e. `AWS.S3`) was accessed. Now, service clients will be built on initial load, but the api models won't be loaded until the client is instantiated.

Fixes #1139 

//cc: @LiuJoyceC


LiuJoyceC:
I like these changes. In addition to fixing the bug, it also now lazy loads service API models even when you directly require an individual service, which is consistent with how it was done before version 2.6.0, and consistent with how models are loaded when all services are required. Even better, now it only loads the model for a specific version.

🚀

PR #1169, opened 2016-10-06
Propagate an error from a stream to s3.upload callback

jbergknoff:
When using `s3.upload` to upload a stream, the AWS SDK currently does not handle `error` events raised by the input stream. The user has two options [1]:

1. Handle the error, in which case `s3.upload` **does not call back**
2. Do not handle the error, in which case the process will crash

This PR makes it so that an error raised by an input stream will propagate to the `s3.upload` callback as an error.

`s3.putObject` also doesn't handle errors on the input stream, but it's not clear how or where to modify that behavior.

By the way, it looks like a few tests in test/s3/managed_upload.spec.coffee are using `upload.send` instead of the wrapper `send`, but then checking the global variable `err` (which is not being updated because it's managed by the `send` wrapper). Those tests may be giving false positives.

[1] Here's a sample program demonstrating the behavior I describe. Comment out the `on('error', ...` line to switch between the cases.

```js
var aws = require('aws-sdk');
var stream = require('stream');

var s3 = new aws.S3({ ... });

var errorStream = new stream.Readable(
  { read: () => { errorStream.emit('error', new Error('message')); } }
);

errorStream.on('error', (error) => { console.log('caught error:', error); });

s3.upload(
  { Bucket: '<bucket>', Key: 'abc', Body: errorStream },
  (error, result) => {
    console.log('s3.upload error:', error);
    console.log('s3.upload result:', result);
  }
);
```

### DIFF
--- a/.changes/next-release/bugfix-s3-20314e6c.json
+++ b/.changes/next-release/bugfix-s3-20314e6c.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "s3",
+  "description": "Propagate an error from a stream to s3.upload callback #1169"
+}
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -168,6 +168,7 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
         self.fillQueue = self.fillStream;
         self.partBuffers = [];
         self.body.
+          on('error', function(err) { self.cleanup(err); }).
           on('readable', function() { self.fillQueue(); }).
           on('end', function() {
             self.isDoneChunking = true;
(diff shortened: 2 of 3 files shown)


jbergknoff:
Looks like the test failed on node <4; I'll update the test tomorrow morning.

jbergknoff:
Ping @chrisradek

chrisradek:
Thanks for the PR @jbergknoff 
This change looks good, I'll run it through our integration tests to make sure nothing fails there.

Would you mind running the `add-change` script to generate a changelog entry and add the JSON file to the PR?
https://github.com/aws/aws-sdk-js/tree/master/scripts/changelog#add-change-cli

Thanks!

jbergknoff:
@chrisradek thanks for having a look. I added the changelog entry to the PR.

chrisradek:
This change passes all tests and seems like the right thing to do when a supplied stream encounters an error. Merging!

Thanks for the PR!

PR #1189, opened 2016-10-14
Adds typescript definitions to SDK

chrisradek:
This PR adds a new script that generates TypeScript Definition files for each service.

It also includes a number of `d.ts` files for classes that aren't dynamically generated at runtime.

If it helps with reviewing, I can upload a new PR that doesn't include all the `d.ts` files that are in the `clients` folder. These are the files generated by the script that will be automatically updated with each service update.

These typings were tested using TypeScript 2.0.x, which was recently released.

Developer tools, like vscode, will pick up the correct typings whether you require the whole sdk using `require('aws-sdk')` or if you require individual services using `require('aws-sdk/clients/s3')`.

I will also work on updating the README to describe how to use these typings.

/cc @LiuJoyceC


chrisradek:
Updated the `global.d.ts` and `index.d.ts` files.
Using `Object.assign` was causing the typings to only work when targeting ES6, now these will also work when targeting ES5.

The `index.d.ts` file also exports the typings as a global namespace, `AWS`, so that browser SDK users can get typing support as well.

clients/acm.d.ts
```diff
@@ -0,0 +1,350 @@
...
+   */
+  resendValidationEmail(params: ACM.ResendValidationEmailRequest, callback?: (err: AWSError, data: {}) => void): Request<{}, AWSError>;
+  /**
+   * Resends the email that requests domain ownership validation. The domain owner or an authorized representative must approve the ACM Certificate before it can be issued. The certificate can be approved by clicking a link in the mail to navigate to the Amazon certificate approval website and then clicking I Approve. However, the validation email can be blocked by spam filters. Therefore, if you do not receive the original mail, you can request that the mail be resent within 72 hours of requesting the ACM Certificate. If more than 72 hours have elapsed since your original request or since your last attempt to resend validation mail, you must request a new certificate.
+   */
+  resendValidationEmail(callback?: (err: AWSError, data: {}) => void): Request<{}, AWSError>;
+}
+declare namespace ACM {
```
LiuJoyceC: Is there a reason the service model's shapes need to be on this namespace? When I type `AWS.ACM.`, I don't think my code completion tool should try to complete `AddTagsToCertificateRequest` or `CertificateBody`, since these aren't actually valid properties on `AWS.ACM`. Instead, this namespace should contain `apiVersions`, `serviceIdentifier`, and `services` (unless these are private, in which case just don't have an `ACM` namespace). The shapes can live in a separate namespace that isn't exported, and you just have to change the references above. I've tested this by changing this namespace name to be `Shapes` (and changing references above), and I still get the correct code completion suggestions when I'm creating my `params` in a request, and I no longer get shape name suggestions when I type `AWS.ACM.`.

chrisradek: I wanted to export all the interfaces so that users could cast a result if they needed to, but I'm not sure if that's necessary. I'll have to take a look at some other TypeScript libraries to see if that's common practice.

You should only see these interfaces when looking at the service client constructors, but not on the service client instances.

clients/acm.d.ts
```diff
@@ -0,0 +1,350 @@
+import {Request} from '../lib/request';
+import {Response} from '../lib/response';
+import {AWSError} from '../lib/error';
+import {Service} from '../lib/service';
+declare class ACM extends Service {
+  /**
```
LiuJoyceC: This might not be a must-have,  but it would be nice to have code completion work when I'm binding params to a service client on construction. You can define a constructor here (perhaps something like `constructor(options?: ServiceConfigurationOptions & ServiceParams)`) and define an interface below like:
One way you could do this is have a constructor such as:
`constructor(options?: ServiceConfigurationOptions & ServiceParams)`
And then near the bottom of the file you can define:
```
interface ServiceParams {
    params?: Shapes.AddTagsToCertificateRequest &
                    Shapes.DeleteCertificateRequest &
                    ...
}
```
Since the `TSGenerator` has to iterate through every operation input shape anyway, it could just keep track of a list of all of the input shape names.

chrisradek: That's an interesting idea. I don't think this will work for S3, or services that have custom config options (anything with dualstack, just S3 for now) if we stop exporting the service interfaces though. I'll take a look.

LiuJoyceC: Well, for S3, since you currently have `constructor(options?: ServiceConfigurationOptions & UseDualstackConfigOptions)`, couldn't that just be changed to `constructor(options?: ServiceConfigurationOptions & UseDualstackConfigOptions & ServiceParams)`?

chrisradek: Yeah, but I'd have to import ServiceParams from somewhere, presumably the base S3 class.

LiuJoyceC: Also, as a modification to what I said above, instead of merging together all of the input shapes, it would be better if each potential param is explicitly listed, since they are all optional here, and we don't want to keep the requirements of the input params. Most services reuse most of their input params, so for many services the list of possible params may actually be shorter than the list of input shapes. So instead of what I have above, it could be more like:
```
interface ServiceParams {
    params?: ACM.AllInputParams
}
```
and inside the `ACM` namespace have:
```
interface AllInputParams {
    CertificateArn?: Arn;
    Tags?: TagList;
    CertificateStatuses?: CertificateStatuses;
    ...
}
```

clients/acm.d.ts
LiuJoyceC: Right, it would only be on the constructor, but there are valid reasons a customer may want to look at the properties of the constructor such as looking up what `apiVersions` are available for a service, or for the case of S3 and DynamoDB, they want to access the `ManagedUpload` or `DocumentClient`. When they try to access these, the constructors will autocomplete with a large number of properties that aren't real and make it difficult to find the actual property they are trying to access.

clients/acm.d.ts
LiuJoyceC: Another reason to have a constructor definition for each service class is to allow code completion for API version dates (by defining the optional `apiVersion` as an enum in `ServiceParams`), especially if we want to encourage users to specify API versions when constructing the service clients.

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    return serviceModel;
+};
+
+/**
+ * Determines if a member is required by checking for it in a list.
+ */
+TSGenerator.prototype.checkRequired = function checkRequired(list, member) {
+    if (list.indexOf(member)) {
```
LiuJoyceC: `if (list.indexOf(member) !== -1)`

LiuJoyceC: Otherwise it's always returning true if the member isn't actually required, and it's returning false if the member happens to be the first one in the list

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+    } else if (type === 'list') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = ' + shape.member.shape + '[];\n';
+    } else if (type === 'map') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = {[key: string]: ' + shape.value.shape + '};\n';
+    } else if (type === 'string' || type === 'character') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = string;\n';
+    } else if (['double', 'long', 'short', 'biginteger', 'bigdecimal', 'integer', 'float'].indexOf(type) >= 0) {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = number;\n';
+    } else if (type === 'timestamp') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = Date;\n';
```
LiuJoyceC: Perhaps this should be `Date|string`?

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+    return customConfigurations;
+};
+
+/**
+ * Generates a type or interface based on the shape.
+ */
+TSGenerator.prototype.generateTypingsFromShape = function generateTypingsFromShape(shapeKey, shape, tabCount) {
+    // some shapes shouldn't be generated if they are javascript primitives
+    if (['string', 'boolean', 'number', 'Date', 'Blob'].indexOf(shapeKey) >= 0) {
+        return '';
```
LiuJoyceC: Instead of skipping shapes that happen to be named these, we could just add an underscore in front to differentiate the name of the shape from the javascript primitives.
It could still be useful to generate a typing for these shapes. For example, some services (like CodeCommit) have a shape named `"Date"` that is modeled as a string rather than a timestamp. If we rename the shape rather than skip it and have it be typed as a JavaScript `Date` object, then it could prevent errors before runtime, such as calling a `Date` method like `getMonth()` on the string.

LiuJoyceC: So for example, the typing could look like:
```
export type _Date = string;
```
And the references to this shape would have to have an underscore added in front as well.

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+            }
+            var required = self.checkRequired(shape.required || [], memberKey) ? '?' : '';
+            code += tabs(tabCount + 1) + memberKey + required + ': ' + member.shape + ';\n';
+        });
+        code += tabs(tabCount) + '}\n';
+    } else if (type === 'list') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = ' + shape.member.shape + '[];\n';
+    } else if (type === 'map') {
+        code += tabs(tabCount) + 'export type ' + shapeKey + ' = {[key: string]: ' + shape.value.shape + '};\n';
+    } else if (type === 'string' || type === 'character') {
```
LiuJoyceC: It would be helpful to support enums.
(after checking for complex types and before starting to check primitive types, so between map and string):
```
else if (Array.isArray(shape.enum)) {
    code += tabs(tabCount) + 'export type ' + shapeKey + ' = ' + shape.enum.map(JSON.stringify).join('|') + ';\n';
```
So for example, Lambda's `Runtime` declaration would look like:
```
export type Runtime = "nodejs"|"nodejs4.3"|"java8"|"python2.7";
```
instead of just a generic string
```
export type Runtime = string;
```

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+    var output = operation.output;
+    var operationName = operation.name.charAt(0).toLowerCase() + operation.name.substring(1);
+
+    var inputShape = input ? className + '.' + input.shape : '{}';
+    var outputShape = output ? className + '.' + output.shape : '{}';
+
+    code += this.generateDocString(operation.documentation, tabCount);
+    code += tabs(tabCount) + operationName + '(params: ' + inputShape + ', callback?: (err: AWSError, data: ' + outputShape + ') => void): Request<' + outputShape + ', AWSError>;\n';
+    code += this.generateDocString(operation.documentation, tabCount);
+    code += tabs(tabCount) + operationName + '(callback?: (err: AWSError, data: ' + outputShape + ') => void): Request<' + outputShape + ', AWSError>;\n';
```
LiuJoyceC: Perhaps we could check whether the input has any required params, and if it has at least one, then don't generate this second second declaration where params is optional.

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+
+/**
+ * Generates class method types for a waiter.
+ */
+TSGenerator.prototype.generateTypingsFromWaiters = function generateTypingsFromWaiters(className, waiterState, waiter, underlyingOperation, tabCount) {
+    var code = '';
+    tabCount = tabCount || 0;
+    var operationName = waiter.operation.charAt(0).toLowerCase() + waiter.operation.substring(1);
+
+    var docString = 'Waits for the ' + waiterState + ' state by periodically calling the underlying ' + className + '.' + operationName + 'operation every ' + waiter.delay + ' seconds (at most ' + waiter.maxAttempts + ' times).';
```
LiuJoyceC: Add a space before the word 'operation'.

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+    code += this.generateDocString(operation.documentation, tabCount);
+    code += tabs(tabCount) + operationName + '(callback?: (err: AWSError, data: ' + outputShape + ') => void): Request<' + outputShape + ', AWSError>;\n';
+
+    return code;
+};
+
+/**
+ * Generates class method types for a waiter.
+ */
+TSGenerator.prototype.generateTypingsFromWaiters = function generateTypingsFromWaiters(className, waiterState, waiter, underlyingOperation, tabCount) {
```
LiuJoyceC: `waiterState` needs to have its first letter lowercased. For example, using the state `"BucketExists"` will result in an error, but `"bucketExists"` is correct.

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+        inputShape = className + '.' + underlyingOperation.input.shape;
+    }
+    if (underlyingOperation.output) {
+        outputShape = className + '.' + underlyingOperation.output.shape;
+    }
+
+    code += this.generateDocString(docString, tabCount);
+    code += this.tabs(tabCount) + 'waitFor(state: "' + waiterState + '", params: ' + inputShape + ', callback?: (err: AWSError, data: ' + outputShape + ') => void): Request<' + outputShape + ', AWSError>;\n';
+    code += this.generateDocString(docString, tabCount);
+    code += this.tabs(tabCount) + 'waitFor(state: "' + waiterState + '", callback?: (err: AWSError, data: ' + outputShape + ') => void): Request<' + outputShape + ', AWSError>;\n';
```
LiuJoyceC: Same comment as above about required params. If the underlying operation has any required params, that perhaps we shouldn't generate this second typing where params is optional.

chrisradek:
Posting my sample tsconfig.json file here:
```javascript
{
    "compilerOptions": {
        "target": "es5",
        "module": "commonjs",
        "outDir": "dist",
        "declarationDir": "dist",
        "declaration": true,
        "lib": [
            "es5",
            "es2015.promise",
            "dom"
        ]
    },
    "include": [
        "src/**/*"
    ]
}
```

Assumes that `.ts` files are located under a `src` directory.

lib/cloudfront/signer.d.ts
```diff
@@ -0,0 +1,48 @@
+export class Signer {
+    /**
+     * A signer object can be used to generate signed URLs and cookies for granting access to content on restricted CloudFront distributions.
+     * 
+     * @param {string} keyPairId - The ID of the CloudFront key pair being used.
+     * @param {string} privateKey - A private key in RSA format.
+     */
+    constructor(keyPaidId: string, privateKey: string);
```
LiuJoyceC: typo: `keyPairId` instead of `keyPaidId`

clients/acm.d.ts
chrisradek: It looks like I'll have to create a custom shape, since only top-level params are actually allowed:
https://github.com/aws/aws-sdk-js/blob/master/lib/service.js#L163

lib/cloudfront/signer.d.ts
chrisradek: Nice catch!

scripts/lib/ts-generator.js
chrisradek: Good catch!

scripts/lib/ts-generator.js
chrisradek: Agreed, I had meant to get to this but hadn't yet. I think we should also still add the primitive type (string) as well, so that users can specify new enums as they become available without having to update the SDK.

scripts/lib/ts-generator.js
chrisradek: Fair point, thanks for finding that.

scripts/lib/ts-generator.js
chrisradek: What's an example of a timestamp in the output that is returned as a string, instead of a Date?
If the shape is modeled as a timestamp, and the value is a string or number, we call [`util.date.parseTimestamp`](https://github.com/aws/aws-sdk-js/blob/master/lib/model/shape.js#L269-L274), and [`parseTimestamp`](https://github.com/aws/aws-sdk-js/blob/master/lib/util.js#L291-L305) always returns a Date object, or throws an error.

Fair point about the above for inputs though.

scripts/lib/ts-generator.js
chrisradek: Thanks, good catch!

scripts/lib/ts-generator.js
```diff
@@ -0,0 +1,437 @@
...
+        var members = shape.members;
+        // cycle through members
+        var memberKeys = Object.keys(members);
+        memberKeys.forEach(function(memberKey) {
+            // docs
+            var member = members[memberKey];
+            if (member.documentation) {
+                code += self.generateDocString(member.documentation, tabCount + 1);
+            }
+            var required = self.checkRequired(shape.required || [], memberKey) ? '?' : '';
```
chrisradek: Thanks, didn't catch this due to the mistake above.

lib/config.d.ts
```diff
@@ -0,0 +1,268 @@
+/// <reference types="node" />
+
+import * as http from 'http';
```
LiuJoyceC: Would doing `import {Agent as httpAgent} from 'http';` and `import {Agent as httpsAgent} from 'https';` make a difference in how much typings data is loaded into memory by Intellisense or by the typescript compiler? I've seen complaints about Intellisense slowing down VS Code or the typescript compiler being too slow when declaration files are large. Since the only typings we need are for `Agent` from both modules, it seems unnecessary to load all of the typings from those modules into memory.

scripts/lib/ts-generator.js
LiuJoyceC: If an output field modeled as a timestamp is sourced from the response header rather than the body, then it remains as a string in the output data. For example, in S3, the `LastModified` shape is modeled as a timestamp. In the output of `getObject` and `headObject`, the `LastModified` field is sourced from the header rather than the body, so the value is a string. In `s3.d.ts`, `LastModified` is typed as a Date, which would be incorrect for the output of these two operations. In some other operations such as `listObjects` or `copyObjects`, the `LastModified` field (modeled by the same `LastModified` shape) is sourced from the response body and the value is converted to a Date object.

scripts/lib/ts-generator.js
LiuJoyceC: So do you mean like this?
```
export type Runtime = "nodejs"|"nodejs4.3"|"java8"|"python2.7"|string;
```
If our main goal is for this to be used for code completion, then this is fine. However, if a user wants to catch compile-time errors, having the enum without the catch-all string type could be really useful in catching typos.

chrisradek: Yes, that's exactly what it'd look like.

I don't want the user to have to update the SDK to use a new enum. If Lambda added support for `nodejs5`, and we didn't specify `string` as an allowable type, they would have to update their SDK to the latest version or else get compilation errors. Today with JavaScript, they could just continue using their current version of the SDK.

chrisradek:
Closed by #1218

PR #1196, opened 2016-10-22
Add AWS_CREDENTIAL_PROFILES_FILE Environment Variable

jans510:
Adding the AWS_CREDENTIAL_PROFILES_FILE Environment Variable, that is for example used in the Java and Python AWS_SDK . Using this variable you can specify a different location of the shared credential file.

### DIFF
--- a/.changes/next-release/feature-EnvironmentVariable-ce9cbf48.json
+++ b/.changes/next-release/feature-EnvironmentVariable-ce9cbf48.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "EnvironmentVariable",
+  "description": "Add AWS_CREDENTIAL_PROFILES_FILE Environment Variable as defined in the Java AWS_SDK"
+}
--- a/doc-src/guide/node-configuring.md
+++ b/doc-src/guide/node-configuring.md
@@ -79,6 +79,9 @@ You may configure credentials for multiple access keys in the same shared
 configuration file using *profiles*. This is discussed in the last part of
 this section.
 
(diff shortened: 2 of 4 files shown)


jans510:
@chrisradek any comments?

jeskew:
Hi @jans510,

It doesn't look like the python SDK supports the `AWS_CREDENTIAL_PROFILES_FILE` environment variable (it uses [AWS_CREDENTIAL_FILE](https://github.com/boto/botocore/blob/5ec04e81d0121ca304bbcc3074d4868df09d2411/botocore/credentials.py#L571) instead, and this feature is maintained for compatibility with some older EC2 CLI tools). Could you tell me a bit more about your use case? I think it would be a breaking change if an environment variable that used to only affect Java applications were to suddenly be picked up by any new JavaScript applications, too, but I'd like to help you find a workaround.

jans510:
Hi @jeskew, thanks for the reply.

I only heard that the python sdk supports this environment variable but didn't looked at it. When we run our java applications in a docker container locally, we use this environment variable to specify the location of the credentials files directory on the container startup. A short time ago we started developing nodejs backends and wanted to start them similar to our java applications, running a container locally and setting this variable with the docker run command, which didn't work as this variable doesn't exists.

I thought it would be a good idea to support the same environment variables on the s3 sdks, but I can understand your concerns, you never know which use cases might be there to use different credentials for different applications on the same host. It would be great if you have any suggestions on solving this problem, at the moment we 'hard code' this environment variable in our nodejs backends, setting the SharedIniFileCredentials manually.

Issue #1018, opened 2016-06-08
Route53.waitFor('resourceRecordSetsChanged') throwing `[Throttling: Resource is not in the state resourceRecordSetsChanged]`


ggoodman:
The `waitFor` method of the `Route53` object occasionally errors with an Error: `[Throttling: Resource is not in the state resourceRecordSetsChanged]`.

The stack trace is as follows:

```
Error submitting Route53 update: Throttling: Resource is not in the state resourceRecordSetsChanged
    at Request.extractError (/path/to/code/node_modules/aws-sdk/lib/protocol/rest_xml.js:44:29)
    at Request.callListeners (/path/to/code/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/path/to/code/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/path/to/code/node_modules/aws-sdk/lib/request.js:615:14)
    at Request.transition (/path/to/code/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/path/to/code/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /path/to/code/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/path/to/code/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/path/to/code/node_modules/aws-sdk/lib/request.js:617:12)
    at Request.callListeners (/path/to/code/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
    at Request.emit (/path/to/code/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/path/to/code/node_modules/aws-sdk/lib/request.js:615:14)
    at Request.transition (/path/to/code/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/path/to/code/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /path/to/code/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/path/to/code/node_modules/aws-sdk/lib/request.js:38:9)
```

Enumerable properties of the error:

```
{ [Throttling: Resource is not in the state resourceRecordSetsChanged]
  message: 'Resource is not in the state resourceRecordSetsChanged',
  code: 'ResourceNotReady',
[...]

PR #1206, opened 2016-11-02
support custom delay and maxAttempts in ResourceWaiter

### LINKED ISSUE #881: DynamoDB waitFor() with custom interval/max_attempts
How would I use waitFor() with a customer interval and/or max_attempts value?

wjordan:
fixes #881.

This PR would allow `delay` and `maxAttempts` parameters to be passed within the `params` object of `ResourceWaiter.wait` (used by the `waitFor` service methods) to override the default settings in the waiter service API definition.

Sample syntax using the example [discussed](https://github.com/aws/aws-sdk-js/issues/1018#issuecomment-227914203) in #1018:

```
var req = route53.waitFor('resourceRecordSetsChanged', {
  Id: "ID",
  delay: 40,
  maxAttempts: 100
}, function (err, data){ /* etc.. */});
```

### DIFF
--- a/lib/resource_waiter.js
+++ b/lib/resource_waiter.js
@@ -136,6 +136,15 @@ AWS.ResourceWaiter = inherit({
       callback = params; params = undefined;
     }
 
+    if (params && params.delay) {
+      this.config.delay = params.delay;
+      delete params.delay;
+    }
+    if (params && params.maxAttempts) {
+      this.config.maxAttempts = params.maxAttempts;
+      delete params.maxAttempts;
+    }
+
     var request = this.service.makeRequest(this.config.operation, params);
     request._waiter = this;
     request.response.maxRetries = this.config.maxAttempts;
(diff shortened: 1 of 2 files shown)


chrisradek:
@wjordan 
Thanks for the PR! One thing we need to be careful of is putting waiter-specific parameters into the same object that defines the parameters that get passed to the underlying operations that get called. Right now there aren't any operations that accept `delay` or `maxAttempts` as top-level params, but we've run into a similar issue with `s3.getSignedUrl` where we allowed setting `Expires`, then an operation added `Expires` as an operation parameter later on.

The safest thing to do would be either:
 * Add a new argument to the function call. Would have to be careful that it's done in a way that doesn't break existing users.
 * Create a simpler way to modify the `delay`/`maxAttempts` value on a request object directly. This would be the least intrusive, but would require an extra call:
`var req = route53.waitFor('resourceRecordSetsChanged', {Id: 'ID'}, callback).setWaitersConfig({});`

What are your thoughts on either of these methods? Personally, the 2nd option seems 'safer' to me, but I could see the 1st option being more intuitive.

PR #1221, opened 2016-11-11
Allows building using noImplicitAny = true

### LINKED ISSUE #1220: Allow noImplicitAny = true from TypeScript
Hi Guys, thanks for the latest update.

When trying to compile a TS app using the library, with the `noImplcitAny` = `true` flag, it fails because the `resolve` method in the CredentialProviderChain class doesn't declares it return type.

I'll send a pull request to fix this in a few minutes

marcote:
Fixes #1220

### DIFF
--- a/lib/credentials/credential_provider_chain.d.ts
+++ b/lib/credentials/credential_provider_chain.d.ts
@@ -8,7 +8,7 @@ export class CredentialProviderChain extends Credentials {
     /**
      * Resolves the provider chain by searching for the first set of credentials in providers.
      */
-    resolve(callback:(err: AWSError, credentials: Credentials) => void);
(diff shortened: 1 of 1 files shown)


chrisradek:
@marcote 
Thanks for the PR! Merging.

blakeembrey:
@marcote Shouldn't you actually enable `noImplicitAny` so this won't happen again? See https://github.com/aws/aws-sdk-js/blob/master/ts/tsconfig.json. Here's the config I usually recommend for people writing type definitions - https://github.com/types/guidelines#tsconfigjson.

maghis:
@blakeembrey see #1224, any other suggestion?

PR #1224, opened 2016-11-13
Typescript: improves chaining experience 

maghis:
Improves on #1220 and adds stricter compiler options as suggested by @blakeembrey

### DIFF
--- a/lib/credentials/credential_provider_chain.d.ts
+++ b/lib/credentials/credential_provider_chain.d.ts
@@ -8,7 +8,7 @@ export class CredentialProviderChain extends Credentials {
     /**
      * Resolves the provider chain by searching for the first set of credentials in providers.
      */
-    resolve(callback:(err: AWSError, credentials: Credentials) => void): any;
+    resolve(callback:(err: AWSError, credentials: Credentials) => void): CredentialProviderChain;
     /**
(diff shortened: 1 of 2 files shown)


blakeembrey:
LGTM 👍 Do you think the `rootDir` should be changed just in case some `.d.ts` files are missed? Not sure if that's an issue here, but looks great 😄

maghis:
@blakeembrey I didn't encounter any issue because of the `rootDir` but I have tried just a small subset. This is so much better than the funky types I was using before :)

blakeembrey:
Sounds good. I'm excited to be using it myself too, once it's working for me 😄 I only brought up `rootDir` because I believe the root is treated at where the `tsconfig.json` file is located otherwise (but I could be wrong, there's too many ways to resolve and I can't remember how that one works because I don't really use it).

chrisradek:
@maghis 
Thanks for the PR! I'll merge to master now. I'm going to try and tackle the other TypeScript issues today, but if I run out of time I'll include this with the next npm publish and follow-up with the rest.

maghis:
@blakeembrey ok, I tested and it looks like it doesn't make any difference.

For reference, I tested by adding an invalid line in one of the root .d.ts and the test failed as expected.

PR #1228, opened 2016-11-15
Removes reference to node typings and stubs dom interfaces

chrisradek:
@blakeembry
I'm only posting a few of the updated service models here since they are updated so frequently.
I removed all instances of
```javascript
///<reference types="node" />
```

I tried to include a `tsconfig.json` with the SDK that specified `dom` in `compilerOptions.lib`, but I still got XMLHttpRequest (and related) errors in my test app if I excluded `dom`.

It looks like the route some libraries have taken is to create empty interfaces. Then, if the user includes `dom` the interfaces are merged. I've done this, however this also reduces the effectiveness of type-checking. Short of creating separate definitions for the browser (difficult since this is a combined project), I'm not sure there's a better way to do this.

Anyway, would love to get some feedback if you have time.

Note: The first commit has most of the changes (I reverted the streams one for now), I uploaded all the updated definitions in case anyone wanted to test against their existing projects.

### DIFF
--- a/clients/acm.d.ts
+++ b/clients/acm.d.ts
@@ -1,10 +1,10 @@
-///<reference types="node" />
 import {Request} from '../lib/request';
 import {Response} from '../lib/response';
 import {AWSError} from '../lib/error';
 import {Service} from '../lib/service';
 import {ServiceConfigurationOptions} from '../lib/service';
 import {ConfigBase as Config} from '../lib/config';
+interface Blob {}
 declare class ACM extends Service {
   /**
    * Constructs a service object. This object has one method for each API operation.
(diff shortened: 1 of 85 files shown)


blakeembrey:
Unfortunately TypeScript doesn't support browser resolution, I have an older issue in the TypeScript repo regarding this as that'd be the best solution for browser type separation. Was there an issue with the streams for now? I haven't used the SDK a ton with TypeScript, but the places you're using these types should mostly be configured based on input options right? Would it make sense to leave it semi-configurable to the user through generics to avoid the messy parts?

Now I think about it, default generic types may also be used here for narrowing the return type, but that can be an unrelated enhancement I'll comment when TypeScript supports it.

chrisradek:
@blakeembrey 
Thanks for the feedback!
The stream support is fine, I just wasn't ready to include them with this PR yet, but I can update the PR with them now.

If you don't have any concerns with mocking the interfaces, I'll go ahead and push this (and the streams) change. Really appreciate the support you've given!

PR #1244, opened 2016-11-29
Signal Initial Validity

### LINKED ISSUE #1243: Valid and Invalid SharedIniFileCredentials Initially Indistinguishable
When one is loading credentials using the role_arn+source_profile configuration in one's `~/.aws/[credentials/config]` setup ([docs](http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html)), the resulting [SharedIniFileCredentials](https://github.com/aws/aws-sdk-js/blob/master/lib/credentials/shared_ini_file_credentials.js) object is indistinguishable whether or not a valid specification has been found (e.g. the profile exists vs. the profile was not found).  Code to demonstrate this follows.

create a temporary directory and then create the following files in that directory.  I'm using the following versions (though they should be irrelevant):
$node --version
v4.2.3
$npm --version
3.10.8

package.json:
```
{
  "name": "tmp",
  "version": "1.0.0",
  "description": "",
  "main": "tmp.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "erikerikson",
  "license": "MIT",
  "dependencies": {
    "aws-sdk": "^2.7.8",
    "fs-extra": "^1.0.0",
    "path": "^0.12.7"
  }
}
```

tmp.js:
```
'use strict';

const fse = require('fs-extra');
const path = require('path');

process.env.AWS_SHARED_CREDENTIALS_FILE = './credentials0';
fse.writeFileSync(
  process.env.AWS_SHARED_CREDENTIALS_FILE,
  '[src]\n' +
  'aws_access_key_id = akid\n' +
  'aws_secret_access_key = secret\n' +
  '\n' +
  '[async]\n' +
  'role_arn = some:role:arn\n' +
  'source_profile = src\n'
);
const validProfileCredentials = new AWS.SharedIniFileCredentials({ profile: 'async' });

process.env.AWS_SHARED_CREDENTIALS_FILE = './credentials1';
fse.writeFileSync(
  process.env.AWS_SHARED_CREDENTIALS_FILE,
  '[src]\n' +
  'aws_access_key_id = akid\n' +
  'aws_secret_access_key = secret\n'
);
const invalidProfileCredentials = new AWS.SharedIniFileCredentials({ profile: 'async' });

let foundDifference = false;
Object.keys(validProfileCredentials).forEach((key) => {
  if (validProfileCredentials[key] !== invalidProfileCredentials[key]) {
    foundDifference = true;
    console.log(`key '${
      key
    }' had different value with valid having '${
      validProfileCredentials[key]
    }' and invalid having '${
      invalidProfileCredentials[key]
    }'`);
  }
});
// secretAccessKey is not enumerable
if (validProfileCredentials.secretAccessKey !== validProfileCredentials.secretAccessKey) {
  foundDifference = true;
  console.log(`key 'secretAccessKey' was different`);
}
if (!foundDifference) {
  console.log('no difference was found');
}
```

then, in your temporary directory, execute:
```
npm install
node tmp.js
```

In the code's current state, one sees `'no difference was found'` printed to the console.

Instead, I would expect to be able to find at least one difference (perhaps the existence of a `roleArn` attribute with the value of `role_arn` in it).  That seems reasonable.

Functionally, this means that while using the aws-sdk one cannot proactively identify and report a credential loading failure.  Instead, one will have to `refresh(callback)` the credentials or more commonly attempt using a feature that requires the use of credentials (thus implying the `refresh` call).

It would be preferable for the credentials to have a discernable difference indicating that the credentials are in a pending load state.

[edit Nov. 29th: working code and instructions to demonstrate the reported issue]

erikerikson:
Signal that the credentials have passed initial validation checks (i.e. the role_arn and source_profiles values were set and the source_profile was successfully loaded for use in invoking STS).  This is a small change to resolve #1243 by exposing the `role_arn` value on the attribute `roleArn` after the initial validation of the settings.

Alter unit tests to validate at least once that the roleArn attribute is created and available in a valid configuration unit test.

See #1243 for more background on the issue and reproduction steps.

### DIFF
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -142,6 +142,8 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       );
     }
 
+    this.roleArn = roleArn;
+
     var sourceCredentials = {
       accessKeyId: sourceProfile['aws_access_key_id'],
       secretAccessKey: sourceProfile['aws_secret_access_key'],
--- a/test/credentials.spec.coffee
+++ b/test/credentials.spec.coffee
@@ -362,6 +362,7 @@ if AWS.util.isNode()
       </AssumeRoleResponse>
       '''
       creds = new AWS.SharedIniFileCredentials()
+      expect(creds.roleArn).to.equal('arn')
       creds.refresh (err) ->
         expect(creds.accessKeyId).to.equal('KEY')
         expect(creds.secretAccessKey).to.equal('SECRET')


erikerikson:
The build errors (from the nodejs 0.10 build) appear to be unrelated to these changes:
```
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
The command "cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js" exited with 8.
```

chrisradek:
@erikerikson 
Thanks for the PR! I think it's a good idea to have the role_arn on the credentials object now that we're able to assume roles with this credential provider, so this seems like an easy merge.

PR #1245, opened 2016-11-30
Fix localStorage check for Safari private browsing

mdurrant:
This PR adds a step to ensure a proper fallback is achieved when in private browsing mode on Safari. 

In private browsing, Safari restricts the localStorage quota at 0, so while `window.localStorage` is a valid object in the browser, any attempt to set data throws an exception.

### DIFF
--- a/lib/credentials/cognito_identity_credentials.js
+++ b/lib/credentials/cognito_identity_credentials.js
@@ -333,8 +333,8 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
     try {
       window.localStorage.setItem('aws.test-storage', 'foobar');
       window.localStorage.removeItem('aws.test-storage');
-      return AWS.util.isBrowser() && window.localStorage !== null && typeof window.localStorage === 'object' ?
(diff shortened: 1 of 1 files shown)


lib/credentials/cognito_identity_credentials.js
```diff
@@ -331,8 +331,10 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
    */
   storage: (function() {
     try {
-      return AWS.util.isBrowser() && window.localStorage !== null && typeof window.localStorage === 'object' ?
```
chrisradek: Can you keep the `AWS.util.isBrowser() && window.localStorage !== null` checks at the top of the try block? As it is, this will always throw an error in node.js (and possibly some 'browser' environments like electron). I know the catch block should handle this case, but I'd like to avoid throwing an error if we can predict it.

mdurrant:
Updated per your comment, let me know if you need any more adjustments.

chrisradek:
Thanks @mdurrant 
Merging this change, thanks for contributing!

Issue #1223, opened 2016-11-11
AWS definitions broke all builds


blakeembrey:
The AWS definitions are incorrect in a number of ways and has broken all builds depending on your TypeScript configuration. Issues I run into:

* Relies on implicit `any`
* Relies on `XMLHttpRequest` (which is not in the standard definition, fails when you don't include the browser-side definitions with TypeScript)
* `Cannot find type definition file for 'node'.` - relies on `@types` which means I can't use the better definitions from Typings or DefinitelyTyped

marcote:
@blakeembrey the bug about implicit `any` it's already fixed in #1221

chrisradek:
@blakeembrey 
What were you using previously to define types for the SDK?

#1221 has just been merged in, I'll go through and see if there are any other methods returning implicit `any`.

Regarding `XMLHttpRequest`, I'll look into that. I wonder if that can be resolved by including a jsconfig.json with this package that includes `dom` under `compilerOptions.lib`.

For your third point, on [typescriptlang.org](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) it mentions that references to paths should not be used in definition files, and using `<reference types="..." />` is preferred. If I removed that reference, wouldn't I also need to remove any imports from that reference as well, such as where I'm importing http/https/stream/etc? Is the best practice to not import those in a library, or am I misunderstanding something?

blakeembrey:
@chrisradek I had my own local definitions that were incomplete (mostly just polyfilled for what I was using). 

> For your third point

It's not entirely true. The `types` reference only causes a error to be emitted by TypeScript, but all the definitions still exist within TypeScript would resolve. Basically, it emits `Cannot find type definition file for 'node'` instead of `Cannot find module 'http'`. It's up to you whether you remove it, though me and a number of others won't be able to use it unless you do and the alternative errors may be preferred - especially if you're typing a definition that shouldn't require `node` anyway (like the browser version of this library, I shouldn't really need to install `node` for that).

midknight41:
Firstly, thank you for creating these definitions! I've been involved in maintaining these definitions at DefinitelyTyped for years and I can't keep up with you folks. :frowning: 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/aws-sdk/aws-sdk.d.ts

That said, this broke my builds too. I've not using @types yet either.

Are you planning to contribute the definitions to DefinitelyTyped? Your definitions should replace mine ideally!

[...]

PR #1249, opened 2016-12-02
Adds support for on method on the S3.ManagedUpload object.

chrisradek:
Reported in #1223


chrisradek:
Added this change into #1240 instead.

Issue #1225, opened 2016-11-14
Typescript def: S3.Body doesn't contain RadableStream as a valid option


mrgleba:
S3.Body should be (as per doc in http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html):
Buffer, Typed Array, Blob, String, ReadableStream

Currenty it is:
export type Body = Buffer|Uint8Array|Blob|string;

forbesmyester:
To further this... This code is valid and should work (it did with DT/@types typings):

    let s3obj = new AWS.S3();
    let body = createReadStream(srcPath).pipe(createGzip());
    let params = {
        Body: body,
        Bucket: getEnv('UPLOAD_S3_BUCKET'),
        ContentEncoding: 'gzip',
        Key: 'some-key-here',
    };
    s3obj.upload(params, next);

jeskew:
Resolved by #1662.

PR #1254, opened 2016-12-06
Add ReadableStream type to Body

goblindegook:
Addresses #1225, allowing the body in S3 requests to be a ReadableStream when using TypeScript.


jeskew:
This was addressed in #1662.

PR #1268, opened 2016-12-14
Add a contructor option to convert empty strings, buffers, and sets to NULL types in DynamoDB document client

jeskew:
This change causes the document client's marshaller to render empty strings, buffers, and sets as `{NULL: true}` instead of `{S: ''}`, `{B: new Buffer('')}`, `{SS: []}`, respectively.

/cc @chrisradek


test/dynamodb/document_client.coffee
```diff
@@ -52,6 +52,39 @@ describe 'AWS.DynamoDB.DocumentClient', ->
         foo: S: 'bar'
       expect(translateInput(input)).to.eql(params)
 
+    it 'translates empty strings', ->
```
chrisradek: For all these tests where you're making sure we remove empty inputs, can you also add tests to verify that we don't translate empty strings/sets/buffers if `convertEmptyValues` isn't set?

PR #1275, opened 2016-12-20
AWS.config.setPromisesDependency(null) forces use of native promises

### LINKED ISSUE #1274: Breaking change of setPromisesDependency() in patch version update
A breaking change was added to the `setPromisesDependency()` and pushed as a patch version update of the `aws-sdk` NPM module in version `v2.6.12` and every version since.

Previously, it was possible to use the native Node.js implementation of the `Promise` object by specifying `null` as the `PromiseDependency` argument of the `addPromises()` function which `setPromisesDependency()` proxies to like this:

```
const sqs = new AWS.SQS();

sqs.config.setPromisesDependency(null);
```

This behaviour was allowed by this line of code up until `v2.6.11`:

https://github.com/aws/aws-sdk-js/blob/v2.6.11/lib/util.js#L753

It was even publicised in the Node.js examples in the Promise support announcement on the official AWS blogpost:

https://aws.amazon.com/blogs/developer/support-for-promises-in-the-sdk/

After `v2.6.12`, only `undefined` (and of course passing the `Promise` object directly) is allowed as a value to default to the native Node.js implementation:

https://github.com/aws/aws-sdk-js/blob/v2.6.12/lib/util.js#L755

The previous snippet would then have to be changed to this to keep the same behaviour:

```
const sqs = new AWS.SQS();

sqs.config.setPromisesDependency();
```
This week some of our long running apps crashed due to this breaking change.

I wonder if it would be possible to add a mention of this in the Changelogs maybe or even allow `null` back as a possible value for the `dep` argument?

The individual commit that broke `v2.6.12` is the following: https://github.com/aws/aws-sdk-js/commit/b0ad8bc1b381ffd5cb1361efe527a9197d5cf205#diff-3deb3f32958bb937ae05c6f3e4abbdf5R754

chrisradek:
Calling AWS.config.setPromisesDependency with null now reverts to native promises. 

Fixes #1274 

/cc @jeskew

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -497,6 +497,8 @@ AWS.Config = AWS.util.inherit({
 
   /**
    * Sets the promise dependency the SDK will use wherever Promises are returned.
+   * Passing `null` will force the SDK to use native Promises if they are available.
+   * If native Promises are not available, passing `null` will have no effect.
    * @param [Constructor] dep A reference to a Promise constructor
    */
   setPromisesDependency: function setPromisesDependency(dep) {
--- a/lib/config.js
+++ b/lib/config.js
@@ -501,9 +501,13 @@ AWS.Config = AWS.util.inherit({
    */
   setPromisesDependency: function setPromisesDependency(dep) {
     PromisesDependency = dep;
+    // if null was passed in, we should try to use native promises
+    if (dep === null && typeof Promise === 'function') {
+      PromisesDependency = Promise;
+    }
     var constructors = [AWS.Request, AWS.Credentials, AWS.CredentialProviderChain];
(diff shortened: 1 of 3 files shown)


lib/config.js
```diff
@@ -501,9 +501,13 @@ AWS.Config = AWS.util.inherit({
    */
   setPromisesDependency: function setPromisesDependency(dep) {
     PromisesDependency = dep;
+    // if null was passed in, we should try to use native promises
+    if (dep === null && typeof Promise === 'function') {
```
jeskew: This should be reflected in the jsdoc comment above: "... If native promises are supported, pass in `null` to use direct the SDK to use them. If native promises are not supported and `null` is provided for this argument, the method is a no-op."

chrisradek:
Good call-out, updated!

PR #1281, opened 2016-12-27
Respect convertEmptyValues option for Maps and Lists in the DynamoDB document client

jcready:
Also only call `typeOf(data)` once


jcready:
This is essentially just an addition to https://github.com/aws/aws-sdk-js/pull/1268

jcready:
Looks like this is being handled in https://github.com/aws/aws-sdk-js/pull/1283 so I'll close this.

PR #1283, opened 2017-01-01
Fix/doc client empty nested members

jeskew:
This PR adds a couple tests to verify #1281 and protect against future regressions.

It also changes the expectations of some tests that assume that the system's path separator will be `/`, which was causing the unit test suite to fail on Windows.

/cc @chrisradek

### DIFF
--- a/lib/dynamodb/converter.js
+++ b/lib/dynamodb/converter.js
@@ -4,35 +4,36 @@ var DynamoDBSet = require('./set');
 
 function convertInput(data, options) {
   options = options || {};
-  if (typeOf(data) === 'Object') {
+  var type = typeOf(data);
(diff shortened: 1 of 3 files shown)


chrisradek:
:shipit:

arun-gupta:
Similar value is not found in Java SDK: http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/ClientConfiguration.html Any idea when is that getting added?

jeskew:
@arun-gupta You will need to open an issue on the [AWS SDK for Java repository](https://github.com/aws/aws-sdk-java).

arun-gupta:
@jeskew thanks, just did!

Issue #1284, opened 2017-01-04
AWS.config.update() type definition is missing CredentialsOptions


yangchristian:
Per [docs](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#constructor-property) and [implementation](https://github.com/aws/aws-sdk-js/blob/master/lib/config.js#L372), `AWS.config.update()` should support the credentials keys like `accessKeyId` and `secretAccessKey` directly.

Minimal repro:
```ts
import * as aws from 'aws-sdk';
aws.config.update({
  accessKeyId: 'abc123',
  secretAccessKey: 'abc123',
});
```

Results in:

> [ts]
Argument of type '{ accessKeyId: any; secretAccessKey: any; region: any; }' is not assignable to parameter of type 'ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions'.
  Object literal may only specify known properties, and 'accessKeyId' does not exist in type 'ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions'.

I should have a quick PR ready soon.

chrisradek:
Closing since this was added with #1285. Thanks for the PR!

PR #1285, opened 2017-01-04
Add CredentialsOptions to type definition of AWS.config.update()

yangchristian:
Addresses aws/aws-sdk-js#1284

### DIFF
--- a/lib/config.d.ts
+++ b/lib/config.d.ts
@@ -63,14 +63,14 @@ export class Config extends ConfigBase {
      * @param {ConfigurationOptions} options - a map of option keys and values.
      * @param {boolean} allowUnknownKeys - Whether unknown keys can be set on the configuration object.
      */
-    update(options: ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions & {[key: string]: any}, allowUnknownKeys: true): void;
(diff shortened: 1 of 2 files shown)


lib/config.d.ts
```diff
@@ -63,14 +63,14 @@ export class Config extends ConfigBase {
...
-    update(options: ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions & {[key: string]: any}, allowUnknownKeys: true): void;
+    update(options: ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions & CredentialsOptions & {[key: string]: any}, allowUnknownKeys: true): void;
     /**
      * Updates the current configuration object with new options.
      * 
      * @param {ConfigurationOptions} options - a map of option keys and values.
      * @param {boolean} allowUnknownKeys - Defaults to false. Whether unknown keys can be set on the configuration object.
      */
-    update(options: ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions, allowUnknownKeys?: false): void;
+    update(options: ConfigurationOptions & ConfigurationServicePlaceholders & APIVersions & CredentialsOptions, allowUnknownKeys?: false): void;
```
chrisradek: Can you add `CredentialsOptions` to the `update` method in `ConfigBase` as well? This should be allowed in service configuration in addition to the global config.

yangchristian:
Thanks @jeskew. Made the change to inline into `ConfigurationOptions` w/`@deprecated` comment and also added a test. Can squash the commits as well if you like.

PR #1303, opened 2017-01-17
Cast string bodies to buffers to ensure correct encoding used

### LINKED ISSUE #1297: S3 putObject is generating incorrect signatures in some cases
When a multibyte utf-8 character is included in the Metadata object, the sdk currently generates an invalid signature causing upload to fail _if_ the `Body` is a string. However if a buffer is used, the correct signature is generated.

Simple example:

```js

AWS.config.update({
  accessKeyId: process.env.AWS_ACCESS_KEY,
  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
});

AWS.config.update({ region: 'us-east-1' });

const bucketName = 'some-s3-bucket';
const bucket = new AWS.S3({ params: { Bucket: bucketName } });

// Will fail
bucket.putObject({
  Body: 'actual string content',
  Key: '/some/file/path',
  Bucket: bucketName,
  Metadata: { 'key': 'multibyte char ç' }
}, (err, res) => {
  console.log('String', err, res);
});

// Will succeed
bucket.putObject({
  Body: new Buffer('not actual string content'),
  Key: '/some/file/path2',
  Bucket: bucketName,
  Metadata: { 'key': 'multibyte char ç' }
}, (err, res) => {
  console.log('Buffer', err, res);
});

```

Output:

```
String { [SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.]
  message: 'The request signature we calculated does not match the signature you provided. Check your key and signing method.',
  code: 'SignatureDoesNotMatch',
  region: null,
  time: Thu Jan 12 2017 17:39:53 GMT-0800 (PST),
  requestId: '45955EF049944D58',
  extendedRequestId: 'TkUKuDhpxI5lYz6uENPSA2udwUUAMgqVT6QSKiGYNcvUYDSGgNj2Dr1Ds7dPhHKmRnwcXQDvyZU=',
  cfId: undefined,
  statusCode: 403,
  retryable: false,
  retryDelay: 76.59807160962373 } null
Buffer null { ETag: '"3690727fb26da2f5984deccf3319c7be"' }
```

jeskew:
This change ensures that bodies are always converted to byte buffers before being passed to an `http.ClientRequest`. It looks like the code that packs headers and short bodies into the same outgoing HTTP packet has a slightly different codepath for non-binary strings, which this change lets the SDK avoid.

Resolves #1297.

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -100,6 +100,9 @@ AWS.NodeHttpClient = AWS.util.inherit({
           total: totalBytes
         });
       });
+      if (typeof body === 'string') {
+        body = new Buffer(body);
+      }
       stream.end(body);
     } else {
       // no request body
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -101,7 +101,8 @@ AWS.NodeHttpClient = AWS.util.inherit({
         });
       });
       if (typeof body === 'string') {
-        body = new Buffer(body);
+        body = typeof Buffer.from === 'function' ?
+          Buffer.from(body) : new Buffer(body);
       }
       stream.end(body);
     } else {
--- a/.changes/next-release/bugfix-S3-3a275bd9.json
+++ b/.changes/next-release/bugfix-S3-3a275bd9.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "S3",
+  "description": "Convert string bodies to buffers to ensure correct encoding is used"
+}
--- a/features/s3/objects.feature
+++ b/features/s3/objects.feature
@@ -28,6 +28,20 @@ Feature: Working with Objects in S3
     And the object "contentlength" should contain "foo"
 
   @multi-byte
+  Scenario: Putting multi-byte metadata
+    When I put a buffer containing "foo" to the key "mbmd" with the metadata key "mdkey" and value "voilà ça marche"
+    Then the object "mbmd" should exist
+    Then I get the object "mbmd"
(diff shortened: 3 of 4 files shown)


lib/http/node.js
```diff
@@ -100,6 +100,9 @@ AWS.NodeHttpClient = AWS.util.inherit({
           total: totalBytes
         });
       });
+      if (typeof body === 'string') {
+        body = new Buffer(body);
```
chrisradek: Since the Buffer constructors are deprecated, we might want to add some feature detection to try `Buffer.from` first, then `new Buffer`.

chrisradek:
Also, can you add a changelog entry using the `./scrips/changelog/add-change.js` script?

lib/http/node.js
```diff
@@ -100,6 +100,10 @@ AWS.NodeHttpClient = AWS.util.inherit({
           total: totalBytes
         });
       });
+      if (typeof body === 'string') {
+        body = typeof Buffer.from === 'function' ?
```
demobox: This unfortunately is broken on Node 4.4.5 :-( `Buffer.from` _is_ a function but fails (because it is actually `Uint8Array.from`)
```
root@45d8cedbf66f:/app# node --version
v4.4.5
root@45d8cedbf66f:/app# node
> Buffer.from === Uint8Array.from
true
> typeof Buffer.from
'function'
> Buffer.from("foo")
TypeError: this is not a typed array.
    at Function.from (native)
...
```

PR #1316, opened 2017-01-19
Limit body casting to just S3

jeskew:
This change limits the scope of #1303 to just S3. It appears that Node does some [special magic](https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L114) to get string bodies into the same packet as headers. This was causing the S3 issue described in #1297 when headers contained characters outside the US-ASCII set, but S3 had no problem accepting UTF-8 characters in a header, provided the body was encoded separately. In our case, this was accomplished by casting the body to a buffer, which had the side effect of disabling the aforementioned special magic.

This feels like a bit of a hack, and it's possible the right answer is just to ask customers with multibyte metadata to URL encode it before passing it to the SDK.

### DIFF
--- a/.changes/next-release/bugfix-Performance-f6176b73.json
+++ b/.changes/next-release/bugfix-Performance-f6176b73.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Performance",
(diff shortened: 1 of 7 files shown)


chrisradek:
:shipit: assuming travis passes (looks like the failures were transient)

PR #1317, opened 2017-01-20
Updates CognitoIdentityCredentials to accept AWS.config options

### LINKED ISSUE #1295: Initiate dynamodb error
Hello, I just started on DynamoDB and I do the following

``` jsx
import { DynamoDB, Config } from 'aws-sdk';

Config.region = 'my region';
Config.credentials = new CognitoIdentityCredentials({
  IdentityPoolId: appConfig.IdentityPoolId
});
const dynamoDB = new DynamoDB(Config);
```

I got Error: Missing credentials in config

How do I pass in my credentials? I have already successfully log into Amazon Cognito and has cognitoUser object.

chrisradek:
Fixes #1295 

/cc @jeskew

### DIFF
--- a/lib/credentials/cognito_identity_credentials.js
+++ b/lib/credentials/cognito_identity_credentials.js
@@ -102,18 +102,35 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
    *     // and multiple users are signed in at once, used for caching
    *     LoginId: 'example@gmail.com'
    *
+   *   }, {
+   *      // optionally provide configuration to apply to the underlying AWS.CognitoIdentity service client
+   *      // if configuration is not provided, then configuration will be pulled from AWS.config
+   *
+   *      // region should match the region your identity pool is located in
+   *      region: 'us-east-1',
+   *
+   *      // specify timeout options
+   *      httpOptions: {
+   *        timeout: 100
+   *      }
    *   });
    * @see AWS.CognitoIdentity.getId
    * @see AWS.CognitoIdentity.getCredentialsForIdentity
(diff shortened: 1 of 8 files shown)


lib/credentials/cognito_identity_credentials.js
```diff
@@ -296,10 +313,14 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
...
     this.webIdentityCredentials = this.webIdentityCredentials ||
       new AWS.WebIdentityCredentials(this.params);
-    this.cognito = this.cognito ||
-      new CognitoIdentity({params: this.params});
+    if (!this.cognito) {
+      var cognitoConfig = AWS.util.merge({}, clientConfig);
+      cognitoConfig.params = this.params;
+      this.cognito = new CognitoIdentity(cognitoConfig);
+    }
     this.sts = this.sts || new STS();
```
jeskew: Should the client config be passed to STS, too?

chrisradek: I went back and forth on that. Since `STS` uses a global endpoint, it's fine if the region the user passes in isn't us-east-1.

I would probably also have to update WebIdentityCredentials to accept client config, since it calls STS as well.

jeskew: There are other configuration options that customers might want to control (e.g., timeouts, retries, http agents, etc.)

chrisradek: Right. Sorry, I wasn't clear in my response. I was initially worried that a different region would cause an issue with STS, but discovered that's not the case. I'll update the PR to pass config to STS (and likewise to WebIdentityCredentials as well).

PR #1325, opened 2017-01-25
Fixes issue wherein the wrong parameter name was used for cross region presigned urls

### LINKED ISSUE #1324: RDS is not injecting correct pre-signed url
## Environment
Using the aws sdk available through Lambda.

## Overview
According to the AWS docs, the RDS `copyDBSnapshot` operation expects `PreSignedUrl`.
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDS.html#copyDBSnapshot-property

## Problem
The code sets `PresignedUrl` instead.
https://github.com/aws/aws-sdk-js/blob/master/lib/services/rds.js#L45-L51

## Symptom
When trying to execute `copyDBSnapshot` for cross-region backups, you get this:
```
InvalidParameterValue: PreSignedUrl could not be authenticated
```

CloudTrail shows neither `PresignedUrl` or `PreSignedUrl` in the `requestParameters`.

chrisradek:
Fixes #1324 

/cc @jeskew

### DIFF
--- a/lib/services/rds.js
+++ b/lib/services/rds.js
@@ -13,11 +13,19 @@ var AWS = require('../core');
      if (crossRegionOperations.indexOf(request.operation) !== -1 &&
          request.params.SourceRegion) {
        request.params = AWS.util.copy(request.params);
-       if (request.params.PresignedUrl ||
+       if (request.params.PreSignedUrl ||
            request.params.SourceRegion === this.config.region) {
          delete request.params.SourceRegion;
        } else {
+         var doesParamValidation = !!this.config.paramValidation;
+         // remove the validate parameters listener so we can re-add it after we build the URL
+         if (doesParamValidation) {
+           request.removeListener('validate', AWS.EventListeners.Core.VALIDATE_PARAMETERS);
+         }
          request.onAsync('validate', this.buildCrossRegionPresignedUrl);
+         if (doesParamValidation) {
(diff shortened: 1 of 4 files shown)


test/event_listeners.spec.coffee
```diff
@@ -215,7 +215,7 @@ describe 'AWS.EventListeners', ->
       makeRequest(->)
       expect(options.timeout).to.equal(15)
 
-    it 'signs only once in normal case', ->
+    it 'signs only once in normal case', (done) ->
```
jeskew: Hopefully, making all of these async should take care of the flickering failures we've been seeing over the past few weeks

lib/services/rds.js
```diff
@@ -13,11 +13,19 @@ var AWS = require('../core');
...
        request.params = AWS.util.copy(request.params);
-       if (request.params.PresignedUrl ||
+       if (request.params.PreSignedUrl ||
            request.params.SourceRegion === this.config.region) {
          delete request.params.SourceRegion;
        } else {
+         var doesParamValidation = !!this.config.paramValidation;
+         // remove the validate parameters listener so we can re-add it after we build the URL
+         if (doesParamValidation) {
+           request.removeListener('validate', AWS.EventListeners.Core.VALIDATE_PARAMETERS);
```
chrisradek: Agreed, we do this often enough throughout the SDK that we should provide a better way to handle this.

PR #1327, opened 2017-01-26
change first selection value of request.httpRequest.region in updateRegion

jaoromi:
I think that request.httpRequest.region shouldn't change value if it have a value. If request.httpRequest.region is undefined, then request.httpRequest.region initialize zones[1] from endpoint domain.


jeskew:
Based on the test failure, it looks like this would be a breaking change. Can I ask what the motivation for this PR was? There might be a way to achieve the goal you had in mind in a backwards compatible way.

jeskew:
Closing, as it looks like this would be a breaking change. Please feel free to reopen if you have any questions or concerns.

PR #1333, opened 2017-01-29
[DynamoDB] Clarify TypeScript typings for item attributes.

pe8ter:
The TypeScript typings for DynamoDB's `AttributeValue` are too loose. Here's a subset:

```
export interface AttributeValue {
  S?: StringAttributeValue;
  N?: NumberAttributeValue;
  ...
}
```

This typing documents that if I have an object that conforms to the `AttributeValue` interface that it can have none, one, some, or all of the specific attribute value types. This is not true. An object that is an `AttributeValue` should have one and only one of the specific attribute value types.

This pull request rectifies this by making each specific attribute value type its own interface and transforms the overloaded `AttributeValue` interface into a type that can be any one of these new interfaces. Here's a subset:

```
export interface StringAttribute {
  S: StringAttributeValue;
}
export interface NumberAttribute {
  N: NumberAttributeValue;
}
...
export type AttributeValue = StringAttribute | NumberAttribute | ...
```

This PR is the minimum needed to add the new clarified typings while not breaking existing consumers of the `AttributeValue` interface.

If you'd like I can add a use case that demonstrates why the loose typing is problematic.

### DIFF
--- a/.changes/next-release/feature-dynamodb-09a8b087.json
+++ b/.changes/next-release/feature-dynamodb-09a8b087.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "dynamodb",
+  "description": "Add specific TypeScript typings for item attributes."
+}
--- a/clients/dynamodb.d.ts
+++ b/clients/dynamodb.d.ts
@@ -185,48 +185,69 @@ declare namespace DynamoDB {
   export type AttributeName = string;
   export type AttributeNameList = AttributeName[];
   export type AttributeUpdates = {[key: string]: AttributeValueUpdate};
-  export interface AttributeValue {
     /**
-     * An attribute of type String. For example:  "S": "Hello" 
-     */
-    S?: StringAttributeValue;
-    /**
-     * An attribute of type Number. For example:  "N": "123.45"  Numbers are sent across the network to DynamoDB as strings, to maximize compatibility across languages and libraries. However, DynamoDB treats them as number type attributes for mathematical operations.
-     */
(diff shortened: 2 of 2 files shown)


jeskew:
Thanks for putting this together, @pe8ter. I agree that the type information provided in this pull request is a more accurate reflection of what the DynamoDB API expects, but it goes beyond [what's expressed in the service model](https://github.com/aws/aws-sdk-js/blob/master/apis/dynamodb-2011-12-05.normal.json#L248) from which the typing files are generated. Those models are shared among several AWS SDKs -- including some written in languages that do not support union types -- so this additional typing information would need to live in the JS SDK in a location that would not be overwritten should the DynamoDB models be updated.

I'll need to confer with other SDK developers on this, but my inclination is to say that the type definitions are not files on which we can accept community contributions.

chrisradek:
Hey @pe8ter 
Thanks, your typings are more accurate for DynamoDB than what's currently available. @jeskew is correct though in that this change goes beyond what is expressed in the model. The TypeScript definitions are currently generated whenever a service model is updated, so to support this change we would need to update the typescript definitions generator to allow for overriding models. This isn't impossible, but something we would likely discourage because what's modeled may eventually drift from what's been customized.

We can accept TypeScript definition file changes, just not for those in the `clients` folder (unless the `scripts/typings-generator.js` file is updated as well).

pe8ter:
Looks like making the change isn't as shallow as I initially thought. We'll continue to use our own custom typings for now.

Thanks for looking into it.

PR #1339, opened 2017-02-02
Add credentialProvider as an optional parameter of ConfigurationOptions

### LINKED ISSUE #1338: [TypeScript] S3.Types.ClientConfiguration and SES.Types.ClientConfiguration missing credentialProvider
The documentation for [`AWS.S3`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#constructor-property) and [`AWS.SES`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SES.html#constructor-property) both state that they accept `credentialProvider`, which is an instance of [`AWS.CredentialProviderChain`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html), as part of the configuration options.

However, the declaration for [`ConfigurationOptions`](https://github.com/aws/aws-sdk-js/blob/67f8e309fb2ab748880a826911573beaa0e7986c/lib/config.d.ts#L155) does not expose such a parameter.

This prevents something like this from compiling:

```typescript
const s3 = new S3({
  region: `${process.env.GC_AWS_REGION}`,
  credentialProvider: new CredentialProviderChain([
    () => { return new EnvironmentCredentials('GC_AWS'); },
    () => { return new EnvironmentCredentials('GC_AMAZON'); },
    () => { return new SharedIniFileCredentials({ profile: 'govcloud' }); },
    () => { return new SharedIniFileCredentials({ profile: 'default' }); },
    () => { return new EC2MetadataCredentials(); }
  ]);
});
```

> foo.ts(45,3): error TS2345: Argument of type '{ region: string; credentialProvider: CredentialProviderChain; }' is not assignable to parameter of type 'ClientConfiguration | undefined'.
  Object literal may only specify known properties, and 'credentialProvider' does not exist in type 'ClientConfiguration | undefined'.

RLovelett:
As discussed in #1338, the documentation for [`AWS.S3`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#constructor-property) and [`AWS.SES`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SES.html#constructor-property) both state that they accept `credentialProvider`, which is an instance of [`AWS.CredentialProviderChain`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html), as part of the configuration options.

However, the declaration for [`ConfigurationOptions`](https://github.com/aws/aws-sdk-js/blob/67f8e309fb2ab748880a826911573beaa0e7986c/lib/config.d.ts#L155) did not expose such a parameter. This patch resolves that issue.

Fixes #1338

### DIFF
--- a/ts/config.ts
+++ b/ts/config.ts
@@ -94,3 +94,12 @@ var config = AWS.config.loadFromPath('/to/path');
 AWS.config.update({
    fake: 'fake' 
 }, true);
+
+// Test constructing with a CredentialProviderChain
+var options = {
+    credentialProvider: new AWS.CredentialProviderChain([
+        () => new AWS.EC2MetadataCredentials()
+    ])
+};
+var s3 = new AWS.S3(options);
+var ses = new AWS.SES(options);
--- a/.changes/next-release/bugfix-TypeScript-5f321e6b.json
+++ b/.changes/next-release/bugfix-TypeScript-5f321e6b.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "TypeScript",
+  "description": "Add `credentialProvider` as an optional parameter of `ConfigurationOptions`."
+}
--- a/lib/config.d.ts
+++ b/lib/config.d.ts
@@ -2,6 +2,7 @@ import {Agent as httpAgent} from 'http';
 import {Agent as httpsAgent} from 'https';
 import {AWSError} from './error';
 import {Credentials, CredentialsOptions} from './credentials';
+import {CredentialProviderChain} from './credentials/credential_provider_chain';
 import {ConfigurationServicePlaceholders, ConfigurationServiceApiVersions} from './config_service_placeholders';
 
 export class ConfigBase extends ConfigurationOptions{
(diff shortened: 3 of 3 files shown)


ts/config.ts
```diff
@@ -93,4 +93,13 @@ var config = AWS.config.loadFromPath('/to/path');
...
+}, true);
+
+// Test constructing with a CredentialProviderChain
+var options = {
+    credentialProvider: new AWS.CredentialProviderChain([
+        () => new AWS.EC2MetadataCredentials()
+    ])
+};
+var s3 = new AWS.S3(options);
+var ses = new AWS.SES(options);
```
RLovelett: @chrisradek I am not exactly sure what the extent of the expected test should be. I feel like this covers the intent of what the change now provides. Obviously, if this is wrong I'll need a little bit more guidance on what to do next.

chrisradek: @RLovelett 
This should be fine. We really just want to make sure the code compiles correctly, and that we don't break it in the future. This captures the intent.

chrisradek:
@RLovelett
Thanks for the PR! Merging!

Issue #1341, opened 2017-02-02
Doesn't work with Webpack 2


Swizec:
After hours of trying to get it to work, I am now convinced that `aws-sdk-js` doesn't work with Webpack 2. Using the latest version from `npm install aws-sdk`.

Here's what I tried:

```javascript
const S3 = require('aws-sdk/clients/s3');

const Config = {
    credentials: {
        accessKeyId: //...,
        secretAccessKey: //...,
    },
    apiVersion: '2006-03-01',
    region: //...
}

let _S3 = new S3(Config);
_S3.putObject(<data>, (err, data) => console.log('done'));
```

Expect to see a file uploaded to S3 and the callback invoked.

Instead this error happened:

```
Uncaught TypeError: regionConfig is not a function
    at features.constructor.initialize (service.js?9975******:53)
    at features.constructor.Service [as constructor] (service.js?9975******:39)
    at features.constructor (util.js?bbc0******:595)
    at new features.constructor (util.js?bbc0******:595)
    at features.constructor.Service [as constructor] (service.js?9975******:30)
    at new features.constructor (util.js?bbc0******:595)
    at Object.putFile (S3Helper.js?be3d******:53)
    at SendControlView.js?0a68**:38
    at HTMLImageElement.canvasImg.onload (DrawingBoardView.js?09cf**:477)
initialize @ service.js?9975******:53
Service @ service.js?9975******:39
features.constructor @ util.js?bbc0******:595
features.constructor @ util.js?bbc0******:595
Service @ service.js?9975******:30
features.constructor @ util.js?bbc0******:595
putFile @ S3Helper.js?be3d******:53
(anonymous) @ SendControlView.js?0a68**:38
canvasImg.onload @ DrawingBoardView.js?09cf**:477
```

The best I can debug it to, is that `var regionConfig = require('./region_config')` in `lib/service.js` returns `Object {rules: Object, patterns: Object}` instead of a function.

I'm 90% sure I'm not doing anything wrong.

I can get it to work using the "load dist version" workaround:

```javascript
import 'aws-sdk/dist/aws-sdk';
const AWS = window.AWS;
```

But this is not ideal because the whole SDK is 1.2MB and I just need S3.

chrisradek:
@Swizec 
Are you adding the `json-loader` to your webpack config?
```javascript
{
  test: /\.json$/, 
  loaders: ['json-loader']
}
```

TheLarkInn:
[...]

PR #1342, opened 2017-02-02
Fix require issue with Webpack 2

Swizec:
Webpack 2 introduced requiring json and js files by default. This seems to cause an issue in some cases where `region_config.json` gets required instead of `region_config.js`.

Renaming the file fixes it for me.

All tests still passing. Not sure how to write a new test for something like this.

Here's the original issue I opened: https://github.com/aws/aws-sdk-js/issues/1341


jeskew:
@Swizec Are you still interested in expanding this PR with tests as discussed in https://github.com/aws/aws-sdk-js/issues/1341#issuecomment-277340517 ? If not, @chrisradek or I could pick this up and add some tests.

jeskew:
Thanks!

PR #1345, opened 2017-02-06
[TypeScript] Add endpoint property to AWS.S3 instances

RLovelett:
The documentation for `AWS.S3` states that it exposes a property, [`endpoint`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#endpoint-property), which is an [`AWS.Endpoint`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Endpoint.html) object.

However, the declaration for `AWS.S3` does not expose the property.

This patch resolves that issue.

I tried to copy all the things I did for #1339.

### DIFF
--- a/.changes/next-release/bugfix-TypeScript-5a92a366.json
+++ b/.changes/next-release/bugfix-TypeScript-5a92a366.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "TypeScript",
+  "description": "Add `endpoint` property to AWS.S3 instance declaration."
+}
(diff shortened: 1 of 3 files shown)


chrisradek:
@RLovelett 
Thanks again for your PR!
The typings in the `clients` folder are all auto-generated with each release, so can't be modified directly.

In this case, all services should expose the `endpoint` parameter. Since each service client extends `Service`, you could add `endpoint: Endpoint` there:
https://github.com/aws/aws-sdk-js/blob/master/lib/service.d.ts

From a testing perspective, I think you can do you could do something like:
```javascript
import {Endpoint} from '../lib/endpoint';
/* existing code */
var endpoint: Endpoint = s3.endpoint;
```

RLovelett:
@chrisradek thank you for the review! I have made the changes you've requested.

chrisradek:
@RLovelett 
Thanks! PR looks good to me, merging.

PR #1350, opened 2017-02-09
Add a presigned POST method to S3

jeskew:
This change adds a means of creating the necessary data for supporting uploads via [presigned multipart POST forms](http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html). The caching mechanism for derived v4 signing keys was changed to allow identifying keys without using a client ID.


jeskew:
@chrisradek Apologies for the rebase; I wanted to grab the browser testing changes from master. The first two commits are unchanged from when you reviewed them before.

lib/signers/v4_credentials.js
```diff
@@ -0,0 +1,95 @@
...
+   * @returns {String}
+   */
+  getSigningKey: function getSigningKey(
+    credentials,
+    date,
+    region,
+    service,
+    shouldCache
+  ) {
+    var cacheKey = [credentials.accessKeyId, date, region, service].join('_');
```
chrisradek: We had 2 issues reported because this cache key wasn't unique enough:
https://github.com/aws/aws-sdk-js/pull/1054

You may need to pass the serviceClientId as well.

ts/s3.ts
```diff
@@ -163,4 +164,16 @@ uploader2.send((err, data) => {
 
 });
 
-var endpoint: Endpoint = s3.endpoint;
\ No newline at end of file
+var endpoint: Endpoint = s3.endpoint;
+
+let presignedPost: PresignedPost = s3.createPresignedPost({
```
chrisradek: Just to be safe, might be useful to also make sure we can access the expected fields (at least `url` and `fields`) on `presignedPost`.

lib/signers/v4_credentials.js
jeskew: The serviceClientId is orthogonal to the uniqueness of this cache key, but I see your point. I'll use an HMAC of the AKID and secret key.

lib/services/s3.js
```diff
@@ -731,6 +731,173 @@ AWS.util.update(AWS.S3.prototype, {
...
+    var boundParams = this.config.params || {};
+    var bucket = params.Bucket || boundParams.Bucket,
+      self = this,
+      config = this.config,
+      endpoint = AWS.util.copy(this.endpoint);
+    if (!config.s3BucketEndpoint) {
+      endpoint.pathname = '/' + bucket;
+    }
+
+    function finalizePost() {
```
jeskew: This function should only be invoked once credentials have been resolved, so it needs to be called from the callback.

ts/s3.ts
jeskew: Sure; will update.

RLovelett:
I'm not sure if this counts as a review or not but I've replaced my custom pre-signed post generation code with this and it works well. This seems like a positive addition to me.

Thanks @jeskew for pitching this and @chrisradek for reviewing. 👍

lib/services/s3.d.ts
```diff
@@ -1,6 +1,67 @@
 import {Service} from '../service';
 import {ManagedUpload} from '../s3/managed_upload';
 import S3 = require('../../clients/s3');
+
+export interface PresignedPostParams {
```
RLovelett: Typically after importing `S3` in TypeScript I can do something like this:

```
import { S3 } from 'aws-sdk';
const params: S3.Types.PresignedPostParams = {
  // ...
};
```

Unfortunately, it seems that the `PresignedPostParams` are exposed this way (likewise `PresignedPost`).

jeskew:
@chrisradek I moved the type definitions around to support access via `S3.PresignedPost` and `S3.Types.PresignedPost` per @RLovelett's feedback. This is my first time contributing to the generated TS definitions and I'm not sure if everything is correct.

PR #1356, opened 2017-02-10
Fix code syntax on Route53 listTrafficPolicyInstancesByHostedZone

robbiet480:
Noticed that all text on the docs after `IsTruncated` was broken. It's because of a hanging/malformed `<code/>`.

[Check out the problem here.](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#listTrafficPolicyInstancesByHostedZone-property) Notice that the font and style is completely different.


jeskew:
Hi @robbiet480,

Thanks for your contribution, but we are unable to accept changes to API models. They are automatically generated from an internal representation, so the next release would wipe out these changes. I'll follow up with the service team to make sure this change is included in their next release.

robbiet480:
Thanks for the explanation @jeskew! :)

Issue #1366, opened 2017-02-20
Credential Provider Chain Resolution


RLovelett:
This is a question about expected behavior.

Let's say I have the following code (Node 7.5.0 because of async/await).

```
(async function () {
  const govcloudCredential = new CredentialProviderChain([
    () => { return new EnvironmentCredentials('GC_AWS'); },
    () => { return new EnvironmentCredentials('GC_AMAZON'); },
    () => { return new SharedIniFileCredentials({ profile: 'govcloud' }); },
    () => { return new SharedIniFileCredentials({ profile: 'default' }); },
    () => { return new EC2MetadataCredentials(); }
  ]);

  const s3 = new S3({
    region: 'us-gov-west-1',
    credentials: undefined,
    credentialProvider: govcloudCredential
  });

  try {
    const foo = await s3.listObjectsV2({
      Bucket: 'bucket'
    }).promise();
    console.log(foo.KeyCount);
  } catch (error) {
    console.error(error);
  }
})()
```

And `~/.aws/credentials` file:

```
[default]
aws_access_key_id = valid-for-commercial
aws_secret_access_key = valid-for-commercial

[govcloud]
aws_access_key_id = valid-for-gov-cloud
aws_secret_access_key = valid-for-gov-cloud

[commercial]
aws_access_key_id = valid-for-commercial
aws_secret_access_key = valid-for-commercial
```

I would expect that the resulting `S3` instance would pull from the explicit provider chain. Such that if I ran the code it would pull from the `govcloud` section and work fine. However, it pulls from `default` and thus prevents my call from succeeding. Just to be sure I checked by removing the `default` group. Then the call actually pulls from `govcloud`.

This behavior, to me, seems counter-intuitive. Can someone help my intuition?

### Update

```
(async function () {
  const govcloudCredential = new CredentialProviderChain([
    () => { return new EnvironmentCredentials('GC_AWS'); },
    () => { return new EnvironmentCredentials('GC_AMAZON'); },
    () => { return new SharedIniFileCredentials({ profile: 'govcloud' }); },
    () => { return new SharedIniFileCredentials({ profile: 'default' }); },
    () => { return new EC2MetadataCredentials(); }
  ]);

  const creds = new Promise<Credentials>((resolve, reject) => {
    govcloudCredential.resolve((error, creds) => {
      if (error) { reject(error); }
      else { resolve(creds); }
    });
  });

  const s3 = new S3({
    region: 'us-gov-west-1',
[...]

PR #1367, opened 2017-02-21
[TypeScript] Allow the service configuration credentials to be nulled

RLovelett:
This enables the service to use a supplied credential provider chain and to ignore those resolved by the global configuration.

As discussed in #1366.


chrisradek:
Thanks @RLovelett 
Merging this in!

PR #1370, opened 2017-02-23
Expose DynamoDB set and converter

jeskew:
This PR exposes the DynamoDB DocumentClient converter and set functionality. It also adds unit and TS tests.

/cc @chrisradek

### DIFF
--- a/ts/dynamodb.ts
+++ b/ts/dynamodb.ts
@@ -19,7 +19,12 @@ if (set.type === 'String') {
     const binary: DynamoDB.DocumentClient.binaryType|undefined = set.values.pop();
 }
 
-const av: DynamoDB.AttributeValue = DynamoDB.Converter.input('string');
+const converter: DynamoDB.Converter = DynamoDB.Converter;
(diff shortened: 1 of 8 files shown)


ts/dynamodb.ts
```diff
@@ -9,6 +9,18 @@ const params: DynamoDB.DocumentClient.GetItemInput = {
     }
 };
```
chrisradek: Can you add tests to make sure we can explicitly set a variable to the `DynamoDB.Converter` type? Also, would be nice to verify we can access `DocumentClient.ConverterOptions`.

Can be simple, like:
```javascript
const converter: Converter = DynamoDB.Converter;
// and a test for input with converter options
const converterOptions: DynamoDB.DocumentClient.ConverterOptions = {convertEmptyValues: true};
DynamoDB.Converter.input('string', converterOptions);
```

How hard would it be to also expose ConverterOptions on the Converter namespace? Just feels a little odd having to access it off the DocumentClient namespace instead.

chrisradek:
Aside from the TS comment, looks good! Thanks for writing so many tests!

PR #1371, opened 2017-02-23
Fixes XHR abort issue

chrisradek:
I changed the XHR abort test to abort once headers are received. Calling `abort` on `send` always created the error and triggered the callback, however it did not appear to actually abort a request.

/cc @jeskew


jeskew:
🐑 🇮🇹

PR #1377, opened 2017-02-28
Generate the list of services supported by the SDK to a separate file.

jeskew:
The list of supported services in the README file has fallen out of date, so this PR adds a script to generate one automatically. The new list also reports whether a service supports CORS.

/cc @chrisradek


chrisradek:
:shipit:

PR #1380, opened 2017-03-01
Use HTTP status code as error code if provided body cannot be parsed.

### LINKED ISSUE #1379: When AWS is down, JSON parser breaks...
protocol/json.js should check the htttp status code and/or content-type before attempting to parse the result.
Current AWS outage in US East causing 
```
 statusCode :503 
<html><body><b>Http/1.1 Service Unavailable</b></body> </html>
```

jeskew:
If a response is received that cannot be parsed, the SDK should treat the response just as if a response with no body had been received.

This should resolve #1379, and I don't believe there are any BC concerns here.

/cc @chrisradek

### DIFF
--- a/dist-tools/test/helpers.coffee
+++ b/dist-tools/test/helpers.coffee
@@ -3,7 +3,7 @@ fs = require('fs')
 evalCode = (code, preamble) ->
   eval """
     (function() {
-      var window = GLOBAL;
+      var window = global;
       #{preamble};
       return #{code};
     })();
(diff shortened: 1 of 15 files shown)


chrisradek:
Awesome!
Can you also update the `request.d.ts` definition to include the `statusMessage`:
https://github.com/aws/aws-sdk-js/blob/master/lib/request.d.ts#L125
and update the request documentation:
https://github.com/aws/aws-sdk-js/blob/master/lib/request.js#L249-L254

chrisradek:
:shipit:

PR #1382, opened 2017-03-01
Fix ManagedUpload never finishes when stream size == part size

johanneswuerbach:
When uploading a stream with an unknown initial length and a total length
exactly equal to a multiple of the part size and a queueSize of `1`, the upload callback is
never finished.

This is caused by only checking whether the upload finished after uploading
the last part, but not when the end event is received.

In our case the end event is received after the last part has been uploaded,
but there is also no more data to be uploaded, resulting in a hanging upload.

I would love to add a test, but couldn't figure out how to trigger receiving the end event after the stream was read, any hints?

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -173,7 +173,11 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
           on('end', function() {
             self.isDoneChunking = true;
             self.numParts = self.totalPartNumbers;
-            self.fillQueue.call(self);
+            if (self.isDoneChunking && self.totalPartNumbers >= 1 && self.doneParts === self.numParts) {
+              self.finishMultiPart();
+            } else {
+              self.fillQueue.call(self);
+            }
           });
       }
     }
--- a/test/s3/managed_upload.spec.coffee
+++ b/test/s3/managed_upload.spec.coffee
@@ -368,6 +368,60 @@ describe 'AWS.S3.ManagedUpload', ->
             expect(e.message).to.equal('message')
             done()
 
+        it 'can send a stream that is exactly equal to part size', (done) ->
+          partSize = 5 * 1024 * 1024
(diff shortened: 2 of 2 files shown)


johanneswuerbach:
Thanks for the hit @jeskew. Tests added and confirmed that they were hanging before the fix and pass after it.

mramato:
@jeskew, I apologize if I should create a new issue instead of commenting here, but I believe this change is causing my uploads to s3 to become corrupted.  While I'm still debugging the problem, in my case sometimes (but not all of the time) `self.finishMultiPart();` is being called too early and the upload completes "successfully" but before all parts are actually complete.  The end result is that the file on s3 is truncated and missing data.  If I roll back to a version of the sdk before this change, everything works 100% of the time (like it has for months).

My server code that performs the upload is really simple:

```javascript
S3DataStore.prototype.uploadForProcessing = function (id, stream) {
    var size = 0;
    var uploadStream = new PassThrough();

    stream.on('data', function (data) {
        size += data.length;
        var canContinue = uploadStream.write(data);
        if (!canContinue) {
            stream.pause();
            uploadStream.once('drain', function () {
                stream.resume();
            });
        }
    });

    stream.on('error', function (error) {
        uploadStream.emit('error', error);
    });

    stream.on('end', function () {
        uploadStream.end();
    });

    return s3.upload({
        Bucket: this._bucket,
        Key: this._prefix + id,
        Body: uploadStream
    }).promise().then(function () {
        return {
            path: id,
            size: size
        };
    });
};
```

Should I create a new issue with the above information?  I can try and open a PR if I can figure out the problem in more detail, but this is a major regression for me in the meantime.

johanneswuerbach:
I also just hit the problem and are working on a fix as we speak.

mramato:
Awesome, thanks!

johanneswuerbach:
https://github.com/aws/aws-sdk-js/pull/1395

PR #1383, opened 2017-03-01
Fixes yardoc generation when service models use mixed-casing

chrisradek:
Currently, if a service model contains mixed casing, yard will fail to generate documentation.

/cc @jeskew


jeskew:
:shipit:

PR #1391, opened 2017-03-07
Add support for loading shared config

### LINKED ISSUE #993: Assume Role Profiles don't follow the CLI and Boto Convention for Configuration
https://github.com/aws/aws-sdk-js/pull/926  apparently added the ability to have assumeRole items in the aws credentials file

but the set up for this doesn't seem to be documented anywhere and it also doesn't follow what works in CLI and Boto and IS documented here: http://docs.aws.amazon.com/cli/latest/topic/config-vars.html#using-aws-iam-roles

so now if i want to use the cli on my machine and the aws sdk my config file and my credentials file needs 

```
...
[assumeroleprofile]
role_arn = arn:aws:iam::accountnum:role/SuperRole
source_profile = main
...
```

could we just get the JS sdk to read the ~/.aws/config like both the CLI and Boto do?

### LINKED ISSUE #1093: Release separate parts of the whole SDK to npm?
Hey there.

Have you ever thought about / do you think you could / would it be possible to release separate working parts of the whole SDK, possibly configurable, e.g. separate AWS.S3, to NPM?

Similarly as it is possible to [Build your own version of the AWS SDK for JavaScript](https://sdk.amazonaws.com/builder/js/) at [amazonaws.com](https://sdk.amazonaws.com/builder/js/), it would be cool to have it on NPM and programatically just connect the separate parts of the whole SDK. The whole SDK is quite large and most of the programmers are not using most of it anyway.

screen shot 2016-08-11 at 16 09 07

Would you be interested in collaboration on this?

### LINKED ISSUE #1296: AWS_PROFILE to respect config role_arn
Support for AWS_PROFILE is great, but without automatically assuming the `role_arn` in `~/.aws/config` for that profile makes it useless for the most common use case for even having profiles (master account delegating access to other accounts).  The node.js SDK appears to just ignore this `role_arn` and try to perform actions on the "master" account.

This could be potentially dangerous, since it behaves differently from the CLI, one might issue commands on an unintended account.

jeskew:
This PR adds support for loading configuration from `~/.aws/config`, including credentials, the default region to use, and role assumption metadata. This is an opt-in feature and will only take effect if the `AWS_SDK_LOAD_CONFIG` environment variable is set to something truthy.

I also added support for some configuration environment variables that we do not currently support, namely `AWS_SHARED_CREDENTIALS_FILE` and `AWS_CONFIG_FILE`, both of which will only be used if `AWS_SDK_LOAD_CONFIG` is set.

This should resolve #1196 as well (via `AWS_SHARED_CREDENTIALS_FILE`).

Resolves #1296 
Resolves #1093 
Resolves #993

### DIFF
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -1,7 +1,7 @@
 var AWS = require('../core');
 var path = require('path');
 var SharedIniFile = require('../shared_ini');
-var STS = require('../../clients/sts');
+require('../../clients/sts');
 
 /**
  * Represents credentials loaded from shared credentials file
@@ -176,7 +176,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       );
     }
 
-    var sts = new STS({
+    var sts = new AWS.STS({
       credentials: new AWS.Credentials(sourceCredentials)
     });
 
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -1,10 +1,7 @@
 var AWS = require('../core');
 var path = require('path');
+var SharedIniFile = require('../shared_ini');
 var STS = require('../../clients/sts');
-
-var configOptInEnv = 'AWS_SDK_LOAD_CONFIG';
-var sharedFileEnv = 'AWS_SHARED_CREDENTIALS_FILE';
-var defaultProfile = 'default';
 
 /**
  * Represents credentials loaded from shared credentials file
@@ -57,7 +54,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
     options = options || {};
 
     this.filename = options.filename;
-    this.profile = options.profile || process.env.AWS_PROFILE || defaultProfile;
+    this.profile = options.profile || process.env.AWS_PROFILE || AWS.util.defaultProfile;
     this.disableAssumeRole = Boolean(options.disableAssumeRole);
     this.get(function() {});
   },
@@ -76,19 +73,27 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
   refresh: function refresh(callback) {
     if (!callback) callback = function(err) { if (err) throw err; };
     try {
-      var profile = {};
-      if (process.env[configOptInEnv]) {
-        var config = new AWS.SharedIniFile({
+      var profiles = {};
+      if (process.env[AWS.util.configOptInEnv]) {
+        var config = new SharedIniFile({
           isConfig: true,
-          filename: process.env.AWS_CONFIG_FILE
+          filename: process.env[AWS.util.sharedConfigFileEnv]
         });
-        profile = AWS.util.merge(profile, config.getProfile(this.profile));
+        for (var i = 0, availableProfiles = config.getProfiles(); i < availableProfiles.length; i++) {
+          profiles[availableProfiles[i]] = config.getProfile(availableProfiles[i]);
+        }
       }
-      var creds = new AWS.SharedIniFile({
+      var creds = new SharedIniFile({
         filename: this.filename ||
-          (process.env[configOptInEnv] && process.env[sharedFileEnv])
+          (process.env[AWS.util.configOptInEnv] && process.env[AWS.util.sharedCredentialsFileEnv])
       });
-      profile = AWS.util.merge(profile, creds.getProfile(this.profile));
+      for (var i = 0, availableProfiles = creds.getProfiles(); i < availableProfiles.length; i++) {
+        profiles[availableProfiles[i]] = AWS.util.merge(
+          profiles[availableProfiles[i]] || {},
+          creds.getProfile(availableProfiles[i])
+        );
+      }
+      var profile = profiles[this.profile] || {};
 
       if (Object.keys(profile).length === 0) {
         throw AWS.util.error(
@@ -98,7 +103,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       }
 
       if (profile['role_arn']) {
-        this.loadRoleProfile(creds, profile, callback);
+        this.loadRoleProfile(profiles, profile, callback);
         return;
       }
 
@@ -126,7 +131,8 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
     if (this.disableAssumeRole) {
       throw AWS.util.error(
         new Error('Role assumption profiles are disabled. ' +
-                  'Failed to load profile ' + this.profile),
+                  'Failed to load profile ' + this.profile +
+                  ' from ' + creds.filename),
         { code: 'SharedIniFileCredentialsProviderFailure' }
       );
     }
@@ -144,7 +150,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       );
     }
 
-    var sourceProfile = creds.getProfile(sourceProfileName);
+    var sourceProfile = creds[sourceProfileName];
 
     if (typeof sourceProfile !== 'object') {
       throw AWS.util.error(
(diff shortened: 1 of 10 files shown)


alsmola:
Anything holding back a merge here? This functionality would be really useful for compatibility with other SDKs.

lib/node_loader.js
```diff
@@ -59,7 +62,19 @@ AWS.util.update(AWS.Config.prototype.keys, {
     return new AWS.CredentialProviderChain();
   },
   region: function() {
-    return process.env.AWS_REGION || process.env.AMAZON_REGION;
+    var env = process.env;
+    var region = env.AWS_REGION || env.AMAZON_REGION;
+    if (!region && env.AWS_SDK_LOAD_CONFIG) {
```
chrisradek: I believe we're actually supposed to check the credentials file for region, then the config file, similar to the CLI/boto.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -2,9 +2,14 @@ var AWS = require('../core');
 var path = require('path');
 var STS = require('../../clients/sts');
 
+var configOptInEnv = 'AWS_SDK_LOAD_CONFIG';
+var sharedFileEnv = 'AWS_SHARED_CREDENTIALS_FILE';
+var defaultProfile = 'default';
```
chrisradek: Should we grab `defaultProfile` from `AWS.util.defaultProfile` instead?

test/credentials.spec.coffee
```diff
@@ -209,6 +233,53 @@ if AWS.util.isNode()
...
+        '''
+        helpers.spyOn(AWS.util, 'readFileSync').andReturn(mock)
+
+        creds = new AWS.SharedIniFileCredentials()
+        creds.get();
+        validateCredentials(creds)
+        expect(AWS.util.readFileSync.calls[0].arguments[0]).to.match(/[\/\\]home[\/\\]user[\/\\].aws[\/\\]config/)
+        expect(AWS.util.readFileSync.calls[1].arguments[0]).to.equal(process.env.AWS_SHARED_CREDENTIALS_FILE)
+
+      it 'loads credentials from ~/.aws/config if AWS_SDK_LOAD_CONFIG is set', ->
```
chrisradek: Can we add a test to make sure the credentials from `~/.aws/credentials` is used preferentially over the credentials in `~/.aws/config` if the same profile exists in both files?

chrisradek:
One thing I'm not clear on, does the following scenario work:
-  profile in `credentials` file contains a roleArn and source profile
- source profile is read from `config` file

test/credentials.spec.coffee
jeskew: I'll add a test for that.

jeskew:
I had to force push due to a merge conflict in AWS.util. The last commit add the ability to assume a role for a profile defined in `~/.aws/credentials` even if the source profile is defined `~/.aws/config` and vice versa.

test/credentials.spec.coffee
```diff
@@ -209,6 +238,83 @@ if AWS.util.isNode()
...
+        helpers.spyOn(AWS.util, 'readFileSync').andReturn(mock)
+
+        creds = new AWS.SharedIniFileCredentials()
+        creds.get();
+        validateCredentials(creds)
+        expect(AWS.util.readFileSync.calls[0].arguments[0]).to.match(/[\/\\]home[\/\\]user[\/\\].aws[\/\\]config/)
+
+      it 'prefers credentials from ~/.aws/credentials if AWS_SDK_LOAD_CONFIG is set', ->
+        process.env.AWS_SDK_LOAD_CONFIG = '1'
+        mock = '''
```
chrisradek: This mock can probably be removed.

test/credentials.spec.coffee
```diff
@@ -370,6 +491,84 @@ if AWS.util.isNode()
...
+          <Credentials>
+            <AccessKeyId>KEY</AccessKeyId>
+            <SecretAccessKey>SECRET</SecretAccessKey>
+            <SessionToken>TOKEN</SessionToken>
+            <Expiration>1970-01-01T00:00:00.000Z</Expiration>
+          </Credentials>
+        </AssumeRoleResult>
+      </AssumeRoleResponse>
+      '''
+      debugger;
```
chrisradek: Extra statement :)

test/credentials.spec.coffee
```diff
@@ -370,6 +491,84 @@ if AWS.util.isNode()
...
+          role_arn = arn
+          source_profile = foo
+          '''
+        else
+          '''
+          [profile foo]
+          aws_access_key_id = akid2
+          aws_secret_access_key = secret2
+          '''
+      helpers.mockHttpResponse 200, {}, '''
```
chrisradek: I'm not sure this test is actually ensuring that the creds from `profile foo` are used instead of `default`.

What do you think about spying on `AWS.STS` or `AWS.Credentials` to get the accessKeyId that was used as the source?

test/credentials.spec.coffee
```diff
@@ -370,6 +491,84 @@ if AWS.util.isNode()
...
+          <Credentials>
+            <AccessKeyId>KEY</AccessKeyId>
+            <SecretAccessKey>SECRET</SecretAccessKey>
+            <SessionToken>TOKEN</SessionToken>
+            <Expiration>1970-01-01T00:00:00.000Z</Expiration>
+          </Credentials>
+        </AssumeRoleResult>
+      </AssumeRoleResponse>
+      '''
+      debugger;
```
chrisradek: Extra statement.

eric-tucker:
Am I to understand that the SDK currently doesn't support ~/.aws at all?  Or just doesn't support using assumed roles through the ~/.aws entries (which has only recently been added to PHP and worked a while in Python).

chrisradek:
@eric-tucker 
The SDK currently sources credentials using `~/.aws/credentials` by default. It actually can use assumed roles if they are defined in the credentials file. This change also allows `~/.aws/config` to be used when sourcing credentials, since most of the other SDKs support this now as well.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -87,11 +87,8 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
         filename: this.filename ||
           (process.env[AWS.util.configOptInEnv] && process.env[AWS.util.sharedCredentialsFileEnv])
       });
-      for (var i = 0, availableProfiles = creds.getProfiles(); i < availableProfiles.length; i++) {
-        profiles[availableProfiles[i]] = AWS.util.merge(
-          profiles[availableProfiles[i]] || {},
-          creds.getProfile(availableProfiles[i])
-        );
+      for (i = 0, availableProfiles = creds.getProfiles(); i < availableProfiles.length; i++) {
```
chrisradek: Nit: Can we either use `var i` here, or declare `var i` at the top of the function? Just worried if we someday change `i` in the for loop above, then this `i` becomes a global.

dotchev:
Why not load `~/.aws/confi` by default?

chrisradek:
@dotchev 
We can't do that until we make a major version bump, because this could lead to the SDK loading different credentials without any code changes in consuming code. We do plan on loading `~/.aws/config` by default like the CLI does when we do a major bump, but I can't give a timeline on when that will be.

lucleray:
Hi,

This page is still confusing : https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html#setting-region-config-file.

**"The SDK for JavaScript automatically searches for a config file when it loads."**

I understand here that I need to set `AWS_SDK_LOAD_CONFIG` to something truthy to get that behavior. Is this documented somewhere ?

jeskew:
The `AWS_SDK_LOAD_CONFIG` environment variable is only documented on that page in the "Order of Precedence for Setting the Region" section. I'll try get a the "Using a Shared Config File" section updated to reflect this.

lucleray:
Thanks!

villasv:
@chrisradek any update on the version bump? Having this inconsistency for over a year is enough justification for a major release IMHO

PersianIronwood:
It would have been nice if you could add an example of using this feature. I am having issues of using assume role with JS.

iwaduarte:
Any updated on this? It would be very regarded if anyone:
1) Provide a wait/gist of assuming the role specified in the config file.
2) Fix the mfa feature for roles that requires it.

This would help a lot in cases where we need to **test it in local environments** without assuming any other dependencies tools. I am not acquainted with the aws-sdk code but I would certainly help figure out if someone pointed me where should I look.

Issue #740, opened 2015-09-29
React Native compatibility


Lohit9:
Is the AWS node js SDK compatible with React native?

AdityaManohar:
I'm not sure of the internal details of React Native, but the SDK should work similarly to any other `npm` modules you may use in your React Native applications.

Let us know if you run into any trouble using the SDK in your React Native applications.

Lohit9:
Ok, I will be implementing a sample in the next 2 days and will mention in case I encounter an issue.

Lohit9:
React Native runs JS on JavaScriptCore (when on the device or simulator) and on Chrome itself (when using Chrome debugging), so modules that depend on built-in Node.js modules won't work. 

The AWS node SDK has many core node dependencies, due to which you cannot use it in your React native app. Unless you find a fork of the SDK that doesn't use any core node modules

AdityaManohar:
@Lohit9 Thanks for the update!

Is there a way you can bundle external dependencies into a React Native app? 

If so you can use the browser version of the AWS SDK for JavaScript by downloading it here: [https://sdk.amazonaws.com/js/aws-sdk-2.2.4.min.js](https://sdk.amazonaws.com/js/aws-sdk-2.2.4.min.js)

You can also build a custom version of the SDK for the browser here: [https://sdk.amazonaws.com/builder/js](https://sdk.amazonaws.com/builder/js)

Hope this helps!

Lohit9:
@AdityaManohar Thanks for the links, trying out stuff to see if the javascript sdks does the job.

chrisradek:
@Lohit9 
Did you have any luck with the browser version of the SDK?

Lohit9:
@chrisradek No I could not get it working. I don't remember the exact issue I had, but the javascript files in the browser version of the SDK were not being imported into the react native javascript code. I believe it was expecting a different kind of file structure/or maybe some dependency related issue.

What I finally did was to use the AWS Mobile SDK and natively bridge it with the react native javascript code. The downside is that I had to do this twice - for Android and iOS, would love to hear from someone who got their react native app working with the browser version of the SDK.

morenoh149:
I'm currently seeing
```

WARNING in ./~/aws-sdk/lib/util.js
Critical dependencies:
50:30-45 the request of a dependency is an expression
53:11-53 the request of a dependency is an expression
 @ ./~/aws-sdk/lib/util.js 50:30-45 53:11-53

[...]

PR #1393, opened 2017-03-07
[WIP] Initial attempt at react-native support

chrisradek:
Initial attempt at supporting react-native.
So far I've tested several services manually, including S3 and Lambda. I need to get the full test suite running in mobile environments before merging this.

Right now we have to create a build of the SDK that runs in react-native, like we do for the browser SDK. This is due to a bug in react native's packager:
https://github.com/facebook/react-native/pull/10451

If this bug is fixed, then the SDK can be required the same as in node.js. 

See updated README for installation instructions.

Addresses #740


lib/react_native_loader.js
```diff
@@ -0,0 +1,24 @@
+var util = require('./util');
+
+// browser specific modules
+util.crypto.lib = require('crypto-browserify');
```
jeskew: Does this get resolved to the pure-js version used in browsers or something specific to react native?

package.json
```diff
@@ -25,14 +25,16 @@
...
     "karma": "^1.4.1",
     "karma-jasmine": "^1.1.0",
     "karma-phantomjs-launcher": "^1.0.2",
     "mocha": "*",
     "repl.history": "*",
     "semver": "*",
     "typescript": "2.0.8",
-    "uglify-js": "2.x"
+    "uglify-js": "2.x",
+    "webpack": "^1.14.0"
```
jeskew: Why not `^2.0.0`? :)

lib/react_native_loader.js
chrisradek: This gets resolved to the pure-js version, same as what browsers use.

package.json
chrisradek: I'm actually hoping we won't need webpack at all. This is a temporary work-around until the issue can be fixed on react-native's side. I used webpack 1 since I had tested with that before webpack 2 was released, and the fields I'm using don't exist in webpack 2 anymore.

If for whatever reason the react native issue can't be fixed, I'll update this to 2+

chrisradek:
Agreed on the canary. Really, we need to do this regardless of supporting react native since we've made it a point to support webpack.

vitalis:
👍  looks great!

bobmccracken:
Have you guys had any internal discussion about replacing https://github.com/awslabs/aws-sdk-react-native with this? The [same question](https://github.com/awslabs/aws-sdk-react-native/issues/50) has been raised over there, but it seems like no one is watching that repo anymore.

MattyK14:
Any update on the react native support?

chrisradek:
#1486 has just been merged into master. The react-native distributable will be available in the next version of the SDK.

PR #1394, opened 2017-03-07
Minor Enhancement

scrpatlolla:


jeskew:
Aren't `.js` files covered by the `'.*'` just to the left of this addition?

jeskew:
Closing due to age. I believe this change would be a no-op (`.js` files should be caught by the `.*` glob), but please comment or reopen if I'm wrong about that.

PR #1395, opened 2017-03-08
Fixes missing last part on managed uploads

johanneswuerbach:
When using queueSize 1 and receiving an end event before the last part
was read, the last part wasn't uploaded and the upload corrupted.

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -173,10 +173,10 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
           on('end', function() {
             self.isDoneChunking = true;
             self.numParts = self.totalPartNumbers;
+            self.fillQueue.call(self);
+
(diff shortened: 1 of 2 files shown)


lib/s3/managed_upload.js
```diff
@@ -173,10 +173,10 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
           on('end', function() {
             self.isDoneChunking = true;
             self.numParts = self.totalPartNumbers;
+            self.fillQueue.call(self);
```
johanneswuerbach: Calling this without further data results in a noop, but when the stream has more data to read it increments the `self.numParts (self.totalPartNumbers )` https://github.com/aws/aws-sdk-js/blob/master/lib/s3/managed_upload.js#L449 causing the if below to be skipped.

jeskew:
Looks like we had a real testing gap on using multipart_upload with streams. Thanks for your help on this.

PR #1405, opened 2017-03-15
Customize waiter configuration

### LINKED ISSUE #881: DynamoDB waitFor() with custom interval/max_attempts

jeskew:
This PR is a continuation of #1206. Rather than keep `delay` and `maxAttempts` as top-level members of the hash passed to `ResourceWaiter`, the waiter will look for them in a key (`@waiter`) that won't collide with any operation arguments.

/cc @chrisradek  

Resolves #1206 
Resolves #881

### DIFF
--- a/lib/resource_waiter.js
+++ b/lib/resource_waiter.js
@@ -136,13 +136,15 @@ AWS.ResourceWaiter = inherit({
       callback = params; params = undefined;
     }
 
-    if (params && params.delay) {
-      this.config.delay = params.delay;
-      delete params.delay;
-    }
-    if (params && params.maxAttempts) {
-      this.config.maxAttempts = params.maxAttempts;
-      delete params.maxAttempts;
+    if (params && params.$waiter) {
+      params = AWS.util.copy(params);
+      if (typeof params.$waiter.delay === 'number') {
+        this.config.delay = params.$waiter.delay;
+      }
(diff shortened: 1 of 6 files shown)


lib/resource_waiter.js
```diff
@@ -136,6 +136,16 @@ AWS.ResourceWaiter = inherit({
       callback = params; params = undefined;
     }
 
+    if (params && params.$waiter) {
+      if (params.$waiter.delay) {
+        this.config.delay = params.$waiter.delay;
```
chrisradek: Just in case someone wants to enter `0`, maybe use a `typeof x === 'number'` check instead.

lib/resource_waiter.js
```diff
@@ -136,6 +136,16 @@ AWS.ResourceWaiter = inherit({
...
+      }
+      if (params.$waiter.maxAttempts) {
+        this.config.maxAttempts = params.$waiter.maxAttempts;
+      }
+      delete params.$waiter;
```
chrisradek: We typically try to treat user input as immutable, that way they can reuse the same object in another call if desired.

jeskew: Good point. I added a call to AWS.util.copy to clone the params so that we're not passing any unknown keys to the service operation.

PR #1408, opened 2017-03-16
Adds support for additional authtraits

chrisradek:
/cc @jeskew

### DIFF
--- a/lib/event_listeners.js
+++ b/lib/event_listeners.js
@@ -117,7 +117,7 @@ AWS.EventListeners = {
         var body = req.httpRequest.body || '';
         if (authtype.indexOf('unsigned-body') >= 0) {
           req.httpRequest.headers['X-Amz-Content-Sha256'] = 'UNSIGNED-PAYLOAD';
-          done();
+          return done();
         }
         AWS.util.computeSha256(body, function(err, sha) {
           if (err) {
--- a/lib/event_listeners.js
+++ b/lib/event_listeners.js
@@ -110,9 +110,15 @@ AWS.EventListeners = {
 
     addAsync('COMPUTE_SHA256', 'afterBuild', function COMPUTE_SHA256(req, done) {
       req.haltHandlersOnError();
-      if (!req.service.api.signatureVersion) return done(); // none
+      var operation = req.service.api.operations[req.operation];
+      var authtype = operation ? operation.authtype : '';
+      if (!req.service.api.signatureVersion && !authtype) return done(); // none
       if (req.service.getSignerClass(req) === AWS.Signers.V4) {
         var body = req.httpRequest.body || '';
+        if (authtype.indexOf('unsigned-body') >= 0) {
+          req.httpRequest.headers['X-Amz-Content-Sha256'] = 'UNSIGNED-PAYLOAD';
+          done();
+        }
         AWS.util.computeSha256(body, function(err, sha) {
           if (err) {
             done(err);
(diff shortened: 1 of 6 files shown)


lib/service.js
```diff
@@ -280,10 +280,15 @@ AWS.Service = inherit({
...
+  getSignerClass: function getSignerClass(request) {
     var version;
+    // get operation authtype if present
+    var operation = request ? request.service.api.operations[request.operation] : null;
+    var authtype = operation ? operation.authtype : '';
     if (this.config.signatureVersion) {
       version = this.config.signatureVersion;
+    } else if (authtype === 'v4' || authtype === 'v4-unsigned-body') {
+      version = 'v4';
     } else {
```
jeskew: Cognito Identity Provider uses ["none"](https://github.com/aws/aws-sdk-js/blob/master/apis/cognito-idp-2016-04-18.normal.json#L455) as an authtype; should support for that be added here?

lib/signers/v4.js
```diff
@@ -169,10 +171,13 @@ AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
...
 
   hexEncodedBodyHash: function hexEncodedBodyHash() {
-    if (this.isPresigned() && this.serviceName === 's3' && !this.request.body) {
+    var request = this.request;
+    if (this.isPresigned() && this.serviceName === 's3' && !request.body) {
       return 'UNSIGNED-PAYLOAD';
-    } else if (this.request.headers['X-Amz-Content-Sha256']) {
-      return this.request.headers['X-Amz-Content-Sha256'];
+    } else if (this.operation && this.operation.authtype && this.operation.authtype.indexOf('unsigned-body') >= 0) {
+      return 'UNSIGNED-PAYLOAD';
```
jeskew: Wouldn't this case be covered by the change to event_listeners (`this.request.headers['X-Amz-Content-Sha256']` should already be `'UNSIGNED-PAYLOAD'`)?

lib/service.js
jeskew: Makes sense

lib/signers/v4.js
```diff
@@ -11,10 +11,12 @@ var expiresHeader = 'presigned-expires';
  * @api private
  */
 AWS.Signers.V4 = inherit(AWS.Signers.RequestSigner, {
-  constructor: function V4(request, serviceName, signatureCache) {
+  constructor: function V4(request, serviceName, options) {
     AWS.Signers.RequestSigner.call(this, request);
     this.serviceName = serviceName;
-    this.signatureCache = signatureCache;
+    options = options || {};
```
jeskew: If someone were constructing a signer with `signatureCache = true`, options would still be a boolean here, so `options.signatureCache` would be undefined.

lib/signers/v4.js
```diff
@@ -11,10 +11,12 @@ var expiresHeader = 'presigned-expires';
...
+    this.signatureCache = options.signatureCache || true;
```
jeskew: Err, is there any way to turn off the signature cache? This will convert `false` to `true`. What about `this.signatureCache = options.signatureCache !== false;`

chrisradek: Good catch! Fixed this and added additional tests.

lib/event_listeners.js
```diff
@@ -110,9 +110,15 @@ AWS.EventListeners = {
...
       req.haltHandlersOnError();
-      if (!req.service.api.signatureVersion) return done(); // none
+      var operation = req.service.api.operations[req.operation];
+      var authtype = operation ? operation.authtype : '';
+      if (!req.service.api.signatureVersion && !authtype) return done(); // none
       if (req.service.getSignerClass(req) === AWS.Signers.V4) {
         var body = req.httpRequest.body || '';
+        if (authtype.indexOf('unsigned-body') >= 0) {
+          req.httpRequest.headers['X-Amz-Content-Sha256'] = 'UNSIGNED-PAYLOAD';
+          return done();
```
jeskew: Yeah, that seems like an important `done` call.

PR #1418, opened 2017-03-21
Adds jitter to DynamoDB retries and allows custom retry logic

chrisradek:
/cc @jeskew

### DIFF
--- a/lib/services/dynamodb.js
+++ b/lib/services/dynamodb.js
@@ -47,18 +47,12 @@ AWS.util.update(AWS.DynamoDB.prototype, {
    * @api private
    */
   retryDelays: function retryDelays(retryCount) {
-    var retryDelayOptions = this.config.retryDelayOptions;
+    var retryDelayOptions = AWS.util.copy(this.config.retryDelayOptions);
 
     if (typeof retryDelayOptions.base !== 'number') {
         retryDelayOptions.base = 50; // default for dynamodb
     }
-
-    var delay;
-    if (typeof retryDelayOptions.customBackoff === 'function') {
-      delay = AWS.util.calculateRetryDelay(retryCount, retryDelayOptions);
-    } else {
-      delay = retryCount > 0 ? AWS.util.calculateRetryDelay(Math.max(0, retryCount - 1), retryDelayOptions) : 0;
-    }
+    var delay = AWS.util.calculateRetryDelay(retryCount, retryDelayOptions);
     return delay;
   }
 });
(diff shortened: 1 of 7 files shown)


lib/services/dynamodb.js
```diff
@@ -47,7 +47,18 @@ AWS.util.update(AWS.DynamoDB.prototype, {
...
+
+    if (typeof retryDelayOptions.base !== 'number') {
+        retryDelayOptions.base = 50; // default for dynamodb
+    }
+
+    var delay;
+    if (typeof retryDelayOptions.customBackoff === 'function') {
+      delay = AWS.util.calculateRetryDelay(retryCount, retryDelayOptions);
+    } else {
+      delay = retryCount > 0 ? AWS.util.calculateRetryDelay(Math.max(0, retryCount - 1), retryDelayOptions) : 0;
```
jeskew: I'm not sure this is in line with the recommendation. [DynamoDB recommends](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.RetryAndBackoff) that the first retry be delayed up to 50 ms, the second up to 100 ms, etc.; whereas this will follow that sequence for retries 2-10 with the first retry being dispatched with no delay.

Applying the same delay calculation to all retries also means that you can just return `AWS.util.calculateRetryDelay(retryCount, retryDelayOptions)` instead of having extra branching logic.

See https://github.com/aws/aws-sdk-php/issues/1186 for discussion on this in another SDK.

test/config.spec.coffee
```diff
@@ -60,8 +60,6 @@ describe 'AWS.Config', ->
       expect(configure(maxRetries: 2).maxRetries).to.equal(2)
 
   describe 'retryDelayOptions', ->
-    it 'defaults to "base: 100"', ->
```
jeskew: Optional: you might want to add a test to service.spec.coffee verifying that a normal client still gets a retry base of 100 set.

lib/services/dynamodb.js
chrisradek: I think you're right based on the documentation on DynamoDB's page. Today the SDK will perform the first retry immediately, but happy to change that (and simplify the code!)

test/config.spec.coffee
chrisradek: There is a separate test that implicitly tests this, but happy to add an explicit test.

dcollinsf5:
@chrisradek I was just working on hacking this to allow this functionality, so this is perfectly timed!  How long does it generally take for a merge like this to make it's way to the NPM registry? Excuse my ignorance, this is the first time I've needed something that was just merged.

chrisradek:
@dcollinsf5 
Glad this came with good timing! We tend to release multiple times per week, so, not long :)

dcollinsf5:
OK, thanks!

PR #1421, opened 2017-03-22
Updates smoke tests to work with temporary credentials

chrisradek:
Some of our tests just won't work using temporary credentials, so they are tagged with `@nosession`.

We can skip running those tests by typing: `cucumber.js -t ~@nosession`

Some of the S3 tests were also failing when using temporary credentials since they were relying on credentials being resolved prior to running them.

/cc @jeskew

### DIFF
--- a/features/s3/step_definitions/objects.js
+++ b/features/s3/step_definitions/objects.js
@@ -102,8 +102,11 @@ module.exports = function () {
   });
 
   this.When(/^I get a pre\-signed URL to GET the key "([^"]*)"$/, function(key, callback) {
-    this.signedUrl = this.s3.getSignedUrl('getObject', {Bucket: this.sharedBucket, Key: key});
-    callback();
+    var world = this;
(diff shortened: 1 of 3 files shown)


features/s3/step_definitions/objects.js
```diff
@@ -151,24 +158,25 @@ module.exports = function () {
...
-      body += 'submit\r\n';
-      body += '--' + this.postBoundary + '--\r\n';
-      this.postBody = body;
-      this.postAction = postData.url;
-      callback();
+      this.s3.createPresignedPost(params, function(err, postData) {
+        var body = Object.keys(postData.fields).reduce(function(body, fieldName) {
+          body += '--' + boundary + '\r\n';
+          body += 'Content-Disposition: form-data; name="' + fieldName + '"\r\n\r\n';
+            return body + postData.fields[fieldName] + '\r\n';
```
jeskew: Nit: lines 165-178 have an extra indentation.

jeskew:
🐑 🇮🇹

PR #1425, opened 2017-03-22
Add a means of specifying tags for s3.upload that applies to both single part and multipart uploads

### LINKED ISSUE #1412: Tagging is not applied by S3.upload() for large files
If I upload a small file using `S3.upload()` and specify tagging then the tags are applied. If however, the file is larger than 15M and under the hood `upload` uses a multipart upload then the tags are silently not applied. 

### Example
```js
const fs = require('fs')

const s3 = new AWS.S3({
  region: 'ap-southeast-2',
  credentials: new AWS.SharedIniFileCredentials(),
})

s3.upload(
  {
    Bucket: 'vigil.media.dev-damon',
    Key: 'test.upload.big',
    Tagging: 'someKey=someValue',
    Body: fs.createReadStream('./some-big.file'),
  },
  (err, data) => {
    if (err) {
      console.errro(err)
    } else {
      console.dir(data)
    }
  }
)
```

And then check if tags applied:
```bash
$ aws s3api get-object-tagging --bucket vigil.media.dev-damon --region ap-southeast-2 --key test.upload.big
{
    "TagSet": []
}
```

If `./some-big.file` is changed to something less than 15M then checking the tags afterwards show they have been applied. 

### Environment
AWS SDK 2.28.0
Node v7.7.3
MacOS 10.12.3

jeskew:
The difference between how S3 expects tags to come in on multipart uploads vs PutObject operations is big enough that it makes sense to add an abstraction rather than perform magic on the parameters passed in.

Resolves #1412 

/cc @chrisradek

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -71,6 +71,11 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
    *     partSize: 10 * 1024 * 1024, queueSize: 1,
    *     params: {Bucket: 'bucket', Key: 'key', Body: stream}
    *   });
+   * @example Creating an uploader with tags
+   *   var upload = new AWS.S3.ManagedUpload({
+   *     params: {Bucket: 'bucket', Key: 'key', Body: stream},
+   *     tags: [{Key: 'tag1', Value: 'value1'}, {Key: 'tag2', Value: 'value2'}]
+   *   });
    * @see send
    */
   constructor: function ManagedUpload(options) {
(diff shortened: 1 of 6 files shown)


lib/s3/managed_upload.js
```diff
@@ -60,6 +60,8 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
    *   failures manually.
    * @option options service [AWS.S3] an optional S3 service object to use for
    *   requests. This object might have bound parameters used by the uploader.
+   * @option options tags [Array<map>] The tags to apply to the uploaded object.
+   *   Each tag should have a `Key` and `Value` keys.
```
chrisradek: Minor: Might be worth adding an example that uses tags.

chrisradek:
The change to throw an error on bad tags rather than silently ignore them looks reasonable. Still approved!

damonmaria:
Just checking. I gather from the commits that this will apply to `S3.upload()` as well as `ManagedUpload`. If so, what is happening with `S3.upload({ Tagging: '...' })`? Will that still exist as it is now (therefore not work if the upload switches to multipart)?

PR #1143, opened 2016-09-16
Update Dependencies

gurpreetatwal:
Update some of the packages the sdk relies on, started out updating xmlbuilder only added the rest in the interest of consolidating the updates into one PR
### Updated:
- xmlbuilder 2.6.2 -> 4.2.1
- sax 1.1.5 -> 1.2.1
- xml2js 0.4.15 -> 0.4.17
- uuid 3.0.0 -> 3.0.1
### xmlbuilder

The main reason for updating xmlbuilder is that the current version relies on lodash 3, which although not deprecated is an older release. Most projects that depend on lodash will most likely install 4 and installing aws-sdk-js in their project will install lodash 3 alongside lodash 4.

I also didn't update to the latest version because version 5 and up [drop support for `node < 0.10`](https://github.com/oozcitak/xmlbuilder-js/commit/586433a144777926341cafb2b4c691a1d1f1621a) and aws-sdk supports `node >= 0.8`. Allthough upgrading to the latest version wouldn't be the worst thing, it would just spit out a warning which `escodegen` currently does as well.

According to the [xmlbuilder changelog](https://github.com/oozcitak/xmlbuilder-js/blob/master/CHANGELOG.md) there don't seem to be any changes that would break the sdk. I also ran the unit tests locally and they all passed. 
### sax changes
- fix bug when object is used as a chunk
- add new event
- [changes between 1.1.5 and 1.2.1](https://github.com/isaacs/sax-js/compare/v1.1.5...master)
### xml2js changes
- added option to allow whitespace only text-nodes
- update xmlbuilder dep to 8.2.0  
  - Note: this will cause the warning to show for `node < 0.10`
- [changes between 0.4.15 and master](https://github.com/Leonidas-from-XIV/node-xml2js/compare/0.4.15...master)

### uuid
- no meaningful changes, internal code reorg
- [changes](https://github.com/kelektiv/node-uuid/compare/v3.0.0...v3.0.1)

PR #1428, opened 2017-03-26
upgrade xmlbuilder and xml2js

baweaver:
While looking for older versions of lodash, I'd noticed you were using an older version of xmlbuilder that transitively includes lodash ~3.5.0.

This can cause any project relying on the aws-sdk to gain about 2M of dependencies if they're using Lodash 4 or above:

```
➜  ember.js git:(master) du -sh ./node_modules/xmlbuilder/node_modules/lodash
1.8M    ./node_modules/xmlbuilder/node_modules/lodash
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
     "sax": "1.1.5",
     "url": "0.10.3",
     "uuid": "3.0.0",
-    "xml2js": "0.4.15",
-    "xmlbuilder": "2.6.2"
+    "xml2js": "^0.4.17",
+    "xmlbuilder": "^8.2.0"
   },
   "main": "lib/aws.js",
   "browser": {


jeskew:
Hi @baweaver,

Thanks for putting this together. There's another PR (#1143) that updates `sax` and `uuid` as well and keeps `xmlbuilder` at version 4.2.1. Later versions of xmlbuilder do not support node 0.8, which the AWS SDK does.

Was the concern with having two versions of lodash the size of the generated bundle, or did you encounter any errors building code that relies on the SDK?

baweaver:
Node v0.8 has been out of support for a while now, over two years at this point, so that may be a bit extreme. See: https://github.com/nodejs/LTS

jeskew:
True, but we still have users running Node 0.8 and NPM v1. `4.2.1` is the latest version of xmlbuilder we would be able to use without introducing a breaking change for those users, and [it does use lodash 4](https://github.com/oozcitak/xmlbuilder-js/blob/v4.2.1/package.json#L25).

baweaver:
I won't drag that topic on then.

The primary concern of this one was to try and get Lodash into a reconcilable version to prevent package bloat where possible, so that would be a valid solution.

PR #1445, opened 2017-04-03
Fix doc example gen

chrisradek:
This update now treats example doc generation errors as a warning instead of a hard failure. A helpful message will be outputted to aide in troubleshooting.

I also discovered that services with a `prefix` field in the `metadata.json` file were not being represented by the example doc generator. Fixed that as well.

/cc @jeskew

### DIFF
--- a/doc-src/templates/api-versions/plugin.rb
+++ b/doc-src/templates/api-versions/plugin.rb
@@ -276,7 +276,9 @@ eof
   end
 
   def load_examples(name, version)
-    paths = Dir[File.join($APIS_DIR, "#{name}-#{version}.examples.json")]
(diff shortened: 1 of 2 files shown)


doc-src/templates/api-versions/plugin.rb
```diff
@@ -276,7 +276,9 @@ def class_info_for(prefix)
   end
 
   def load_examples(name, version)
-    paths = Dir[File.join($APIS_DIR, "#{name}-#{version}.examples.json")]
+    @info ||= JSON.parse(File.read(File.join($APIS_DIR, 'metadata.json')))
```
jeskew: I can't believe that's a single operator. :)

PR #1457, opened 2017-04-12
Fix/stop stringing payload blobs

jeskew:
The SDK is currently converting members of output structures that represent the entire response payload to strings (unless the member is tagged as "streaming" in the model). This is unwanted behavior in APIGateway's `getSdk` method, where the payload is a binary blob that would be corrupted by being converted to a string, and the SDK therefore has a customization to use the raw payload instead for that operation. It is, however, behavior on which customers might be relying for `AWS.Lambda.invoke`, `AWS.IotData.deleteThingShadow`, `AWS.IotData.getThingShadow`, and `AWS.IotData.updateThingShadow`. 

This PR flips that setup so that the latter operations have a customization and returning the raw payload for operations like `AWS.APIGateway.getSdk` is the default. That way, any future operation that returns binary data will expose it safely to customers (i.e., as a buffer) rather than casting it to a string.

### DIFF
--- a/lib/services/apigateway.js
+++ b/lib/services/apigateway.js
@@ -19,20 +19,8 @@ AWS.util.update(AWS.APIGateway.prototype, {
     if (request.operation === 'getExport') {
       var params = request.params || {};
       if (params.exportType === 'swagger') {
-        request.addListener('extractData', this.parsePayload);
+        request.addListener('extractData', AWS.util.convertPayloadToString);
       }
-    }
-  },
-
-  /**
-   * @api private
-   */
-  parsePayload: function parsePayload(resp) {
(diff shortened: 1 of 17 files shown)


lib/services/apigateway.js
```diff
@@ -16,18 +16,23 @@ AWS.util.update(AWS.APIGateway.prototype, {
...
+        request.addListener('extractData', this.parsePayload);
+      }
     }
   },
 
-  useRawPayload: function useRawPayload(resp) {
+  /**
+   * @api private
+   */
+  parsePayload: function parsePayload(resp) {
```
jeskew: Sure. Will centralize.

PR #1459, opened 2017-04-12
add customUserAgent to ConfigurationOptions

ericb:
Allows setting `customUserAgent` configuration option in Typescript

### DIFF
--- a/lib/config.d.ts
+++ b/lib/config.d.ts
@@ -176,7 +176,9 @@ export abstract class ConfigurationOptions {
      */
     correctClockSkew?: boolean
     /**
-     * Custom User-Agent
+     * Sets a custom User-Agent string.
+     * In node environments this will set the User-Agent header, but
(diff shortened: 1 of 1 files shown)


jeskew:
I believe this property is not exposed because we can't append anything to the user agent in a browser environment (in browsers, the SDK's user agent are sent [via the `X-Amz-User-Agent` header](https://github.com/aws/aws-sdk-js/blob/f83d12c83542d8c22652ac39e6e8c3a9a722b02b/lib/http.js#L114)). That said, I don't have any objection to exposing this property so long as that limitation is called out in documentation. Could you update the doc string to mention that the `User-Agent` header will only be modified in a node environment?

ericb:
Thanks @jeskew - I've updated the docstring to mention both environments.

jeskew:
Thanks for your contribution!

PR #1462, opened 2017-04-13
Bind responses to the resolved value of promises

### LINKED ISSUE #1461: using .promise() makes it difficult to get the requestId property
Because only the response body is provided  as the resolution value, unlike the original library that the `.promise()` function came from (aws-sdk-promise), it's not possible to get the requestId.

Instead, I have to do this:
```javascript
ec2 = new aws.EC2({region: us-east-1});
let request = ec2.describeRegions();
let data = await request.promise();
let requestId = request.response.requestId; // <---- wow ugly.
```

This is not so nice.  Is there a better way to do this?  Can you add some way to allow the requestId to be passed back?

jeskew:
This PR adds a `$response` property to the value passed to the resolution handler by promises created by `AWS.Request.promise`. This will allow users to access pagination methods, retryCount, requestId, etc.

This should resolve #1461

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -769,7 +769,9 @@ AWS.Request.addPromisesToClass = function addPromisesToClass(PromiseDependency) 
         if (resp.error) {
           reject(resp.error);
         } else {
-          resolve(resp.data);
+          var data = AWS.util.copy(resp.data);
+          data.$response = resp;
+          resolve(data);
         }
       });
       self.runTo();
--- a/.changes/next-release/feature-Promises-14caa793.json
+++ b/.changes/next-release/feature-Promises-14caa793.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "Promises",
+  "description": "Binds response object to the data object with which successful request promises are resolved"
+}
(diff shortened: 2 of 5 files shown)


lib/request.js
```diff
@@ -769,7 +769,9 @@ AWS.Request.addPromisesToClass = function addPromisesToClass(PromiseDependency)
         if (resp.error) {
           reject(resp.error);
         } else {
-          resolve(resp.data);
+          var data = AWS.util.copy(resp.data);
```
chrisradek: Do we need to copy `resp.data` here? I'm a little worried about copying Buffers for operations that return binary data. Since you can't mix promises/callbacks/streams, we shouldn't really have to worry about mutating the `data` object.

jeskew: AWS.util.copy is a shallow object clone, so this would just create a new object and assign the members from the output shape structure to that new object. Doing the copy prevents a reference cycle of `data.$response.data.$response.data...` etc.

chrisradek: Ah gotcha, looks good then!

wryun: This has messed up our mocking of the AWS SDK (which relies on... some messiness):

https://gist.github.com/wryun/347283d8e715afc4f904ffe5838c536c

Do you have any suggestions for how to do this in a nicer way?

PR #1467, opened 2017-04-18
Bump "buffer" module version to support old Android Browsers

yokoe-naosuke-peco:
With old Android Browser (ex. Android 4.4.2),  [s3.upload](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property) fails with undefined method error on ArrayBuffer.

This is because old Android Browser supports ArrayBuffer but not ArrayBuffer.isView.
The issue seems to be fixed on buffer@v5.0.6. 
https://github.com/feross/buffer/commit/2ac965b088d5e9019930c3d9016c7a213994e2fb

I tried `npm run unit` and `npm run browsertest` on aws-js-sdk with buffer v5.0.6 and there was no failure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "uglify-js": "2.x"
   },
   "dependencies": {
-    "buffer": "4.9.1",
+    "buffer": "5.0.6",
     "crypto-browserify": "1.0.9",
     "jmespath": "0.15.0",
     "querystring": "0.2.0",


jeskew:
I verified that with this fix, the following commands still run without an issue on Node 0.8:
* a clean `npm install --production`
* `node node_modules/.bin/mocha -- test/protocol/ test/query/ test/json/ test/xml/ test/model/`
* `node node_modules/.bin/cucumber.js --tags ~@file --tags ~@streaming`

Thanks for putting this PR together!

PR #1478, opened 2017-04-21
Ignore some development files in .npmignore

jorrit:


jeskew:
Thanks for the contribution!

PR #1480, opened 2017-04-24
fix typo in the documentation of retryDelayOptions.base

jfromaniello:
I think there is a typo in the documentation of `retryDelayOptions.base`, it says:

> The base number of milliseconds to use in the exponential backoff for operation retries. Defaults to 100 ms for all services **accept** DynamoDB, where it defaults to 50ms

I think it should be "except". 

Please note that I am not a native speaker and I could be wrong.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -100,7 +100,7 @@ var PromisesDependency;
  *     Currently supported options are:
  *
  *     * **base** [Integer] &mdash; The base number of milliseconds to use in the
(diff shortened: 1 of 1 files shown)


chrisradek:
@jfromaniello 
You are correct, thanks for the correction!

jfromaniello:
Glad to help! This is the best SDK I have ever used ❤️

PR #1487, opened 2017-04-27
Fixed TypeScript compile error

simplegsb:
When attempting to compile a simple TypeScript project that uses aws-sdk, I was getting the following error:

node_modules/aws-sdk/lib/config.d.ts(39,37): error TS2693: 'Promise' only refers to a type, but is being used as a value here.

This change has resolved the error but I'm not 100% confident it is correct.

...also I hope master is the correct branch to be submitting the request for...

### DIFF
--- a/lib/config.d.ts
+++ b/lib/config.d.ts
@@ -36,7 +36,7 @@ export class ConfigBase extends ConfigurationOptions{
     /**
      * Gets the promise dependency the SDK will use wherever Promises are returned.
      */
-    getPromisesDependency(): typeof Promise | void;
+    getPromisesDependency(): Promise<any> | void;
     /**
      * Sets the promise dependency the SDK will use wherever Promises are returned.
      * @param {function} dep - a reference to a Promise constructor


jeskew:
Master would be the correct branch, but the `getPromisesDependency` function should return a promise constructor and not an instance of a promise. Can you share some information about the environment and context in which you're seeing this issue? (E.g., the OS, node version, your tsconfig.json file, etc.)

simplegsb:
Ah right. I'm not sure what TypeScript is complaining about then...

My setup is:

- Ubuntu: 16.04.1
- Node: v6.9.2
- npm: 3.10.9

My tsconfig.json is very simple:

    {
      "compilerOptions":
      {
        "outDir": "./lib",
        "rootDir": "./src",
        "sourceMap":  true
      }
    }

And my app is very simple too:

	import { config, SES } from 'aws-sdk';
	import * as bodyParser from 'body-parser';
	import * as express from 'express';
	import * as nodemailer from 'nodemailer';
	
	config.region = 'us-west-2';
	
	let app = express();
	app.use(bodyParser.json());
	
	// CORS
	app.use((req: express.Request, res: express.Response, next: express.NextFunction) =>
	{
		res.header('Access-Control-Allow-Headers', 'Content-Type');
		res.header('Access-Control-Allow-Methods', 'POST');
		res.header('Access-Control-Allow-Origin', req.get('origin'));
	
		if (req.method === 'OPTIONS')
		{
			res.end();
		}
		else
		{
			next();
		}
	});
	
	app.post('/shopping-list', (req: express.Request, res: express.Response) =>
	{
		let transporter = nodemailer.createTransport({ SES: new SES() } as any);
	
		let mailOptions: nodemailer.SendMailOptions =
		{
			from: 'xxx',
			to: 'xxx',
			subject: 'Shopping List for ' + req.body.familyName + ', ' + req.body.startDate,
			text: 'Shopping List for ' + req.body.familyName + ', ' + req.body.startDate,
			attachments:
			[
				{
					content: req.body.shoppingList,
					filename: 'Shopping List for ' + req.body.familyName + ', ' + req.body.startDate + '.csv'
				}
			]
		};
	
		transporter.sendMail(mailOptions, (err: any, info: nodemailer.SentMessageInfo) =>
		{
			if (err)
			{
				console.log(err);
				res.status(500);
			}
	
			res.end();
		});
	});
	
	app.listen(process.env.PORT || 3000);

jeskew:
Could you try adding a `"lib"` section to your tsconfig file? I believe that by default `es2015.promise` is not included. The [example tsconfig.json](https://github.com/aws/aws-sdk-js/blob/master/ts/tsconfig.json#L8) in the SDK uses `["es5", "es2015.promise"]` for its lib section.

simplegsb:
Ah that worked! Thank you for your help!

I also needed to remove the @types/aws-sdk package from my dependencies as it was giving me this error:

`node_modules/@types/aws-sdk/index.d.ts(349,33): error TS2304: Cannot find name 'Blob'.`

Thanks again, I'll close this PR.

PR #1496, opened 2017-05-03
Ensure req.send is not called until an error handler is registered when creating readable streams

### LINKED ISSUE #1479: .getObject().createReadStream invalid error handling
```.getObject((err, data) => {})``` takes a callback which traps errors with the object such as invalid bucketname, key etc. 

If this callback is specified then ```.createReadStream``` will have write after end errors causing the process to end.

If the callback is omitted the stream will function correctly **however** all callback, rather than read errors aren't firing the ```stream.emit('error, ...)```, causing the stream to hang.

jeskew:
This PR fixes #1479.

Essentially, `AWS.Request::createReadStream` was calling `send` on the underlying request before attaching an error handler to itself that piped emitted errors onto the stream. Any synchronous errors (e.g., validation) would simply be swallowed and the stream would stall.

I also had to update the mock streams used in tests to support a `pipe` operation.


jeskew:
Our tests are currently failing when run on Node 7.10. I'll address that issue in a separate PR and rebase.

chrisradek:
Shipt it!

Issue #1300, opened 2017-01-16
S3 getObject hangs after "write after end" error, when concurrently downloading several files


perhallstroem:
Using one of the standard Node examples for downloading to a file (http://docs.aws.amazon.com/de_de/sdk-for-javascript/v2/developer-guide/requests-using-stream-objects.html), I am seeing frequent "write after end" errors. Using aws-sdk 2.7.21, Node 7.4.0 and running on Mac OS 10.11.6 – same happens on Ubuntu 16.04. Having searched through the issues here, I notice that there are quite a large number of issues related to getObject, of which some are similar to this one. Most of them are old, and some of them are resolved in the "_well, we'll try this and close this for now ..._" way. Hence, I can't tell for sure if this issue duplicates any other issue.

The described happens somewhat randomly, but the below code snippet can be used to reproduce the issue. It seems to happen more often the more concurrent files I am downloading. It also seems to happen more frequently when downloading from a geographically close bucket, i.e. when transfer rate is higher. Surprisingly, the error event is emitted by the **read stream** – that is, the stream returned by the call to `createReadStream()`. I am not sure who or what attempts to write to that stream, especially "after end".

```
"use strict";

const AWS               = require('aws-sdk');
const createWriteStream = require('fs').createWriteStream;

const s3 = new AWS.S3({region: 'ap-southeast-2', sslEnabled: true, maxRetries: 0});

const download = (key) => {
  console.log(`Starting download of ${key}`);
  const file = createWriteStream(`/tmp/${key}`);
  s3.getObject({Bucket: 'b7f', Key: key, RequestPayer: 'requester'}, (err, data) => {
    if (err) throw new Error(`Unable to download ${key}: ${err}`);
    console.log(`Successfully downloaded ${key}`);
  }).createReadStream().pipe(file);
};

s3.listObjectsV2({Bucket: 'b7f', RequestPayer: 'requester'}, (err, data) => {
  if (err) throw err;
  data.Contents.forEach((key) => download(key.Key));
});
```

```Starting download of 13a69rk6ifb40wk8oc4o4wkgk
Starting download of 19bsdj5ghpmo008ogck4kg4cc
Starting download of 1ioz372xmx8g48ocsgk0k0gok
Starting download of 1j0lmsd3ucxs0w8g0o0wo800o
Starting download of 1wr9ryp33i74kcgksw4c0c0k8
Starting download of 293etv4hbvms4wok0wwwos8ww
[...]

PR #1507, opened 2017-05-10
Fix socket read timeout

### LINKED ISSUE #1392: Callback isn't called when connection hangs after receiving some data.
Suppose we make a request to dynamodb. The request starts and receives some data, e.g.
```
httpDownloadProgress { loaded: 789073, total: '2170168' }
```
Then suppose that the data transfer hangs (this happens occasionally even with the production DynamoDB).

There is how the timeout is handled in the code (lib/http/node.js):
```
stream.setTimeout(httpOptions.timeout || 0, function() {
      if (cbAlreadyCalled) return; cbAlreadyCalled = true;
```
Note that `cbAlreadyCalled` here is `true`, because we've already received some data.
Hence the callback isn't called.
The request hangs forever. =(

chrisradek:
Fixes #1392 

Currently, when the SDK receives response headers after sending a request, it internally sets some state that prevents response errors/timeouts from affecting an AWS Request's lifecycle.

This caused problems (reported in #1392) wherein the SDK would receive response headers, but then the connection would hang. Since the socket read timeout was essentially being ignored in this case, the AWS Request could never enter a `complete` state. This prevented a user's callback from being triggered. 

This should mitigate #1300 as well, since now socket read timeouts should eventually be triggered.

/cc @jeskew

### DIFF
--- a/test/mocks/shaky-stream.js
+++ b/test/mocks/shaky-stream.js
@@ -28,7 +28,7 @@ util.inherits(ShakyStream, Readable);
 ShakyStream.prototype._read = function _read(size) {
     if (!this._didStart) {
         this._didStart = true;
-        this.push(Buffer.from('{"Count":1,"Items":[{"id":{"S":"2016-12-11"},"dateUTC":{"N":"1481494545591"},'));
+        this.push(new Buffer('{"Count":1,"Items":[{"id":{"S":"2016-12-11"},"dateUTC":{"N":"1481494545591"},'));
     }
     if (this._didStart && this._isPaused) {
         return;
@@ -36,7 +36,7 @@ ShakyStream.prototype._read = function _read(size) {
         this._isPaused = true;
         var self = this;
         timeoutFn(function() {
-            self.push(Buffer.from('"javascript":{"M":{"foo":{"S":"bar"},"baz":{"S":"buz"}}}}],"ScannedCount":1}'));
+            self.push(new Buffer('"javascript":{"M":{"foo":{"S":"bar"},"baz":{"S":"buz"}}}}],"ScannedCount":1}'));
             self.push(null);
         }, this._shakyTime);
     }
--- a/test/mocks/shaky-stream.js
+++ b/test/mocks/shaky-stream.js
@@ -1,0 +1,45 @@
+var stream = require('stream');
+var util = require('util');
+
+var Readable = stream.Readable;
+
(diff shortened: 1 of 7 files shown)


test/request.spec.js
```diff
@@ -758,14 +758,13 @@ describe('AWS.Request', function() {
...
-          expect(error).to.be["null"];
+          expect(error).to.not.be["null"];
         });
         s.on('data', function(c) {
           return data += c.toString();
         });
         return s.on('end', function() {
-          expect(error).to.be["null"];
-          expect(data).to.equal('FOOBARBAZQU');
+          expect(error).to.not.be["null"];
```
jeskew: Is this a breaking change? Previously, users would get up to `content-length` bytes and no more if the returned stream exceeded that length; now, the stream will emit an error event in that case. Was there ever a case where `content-length` would be wrong but the user was OK with receiving a truncated stream?

test/mocks/shaky-stream.js
```diff
@@ -0,0 +1,45 @@
...
+var util = require('util');
+
+var Readable = stream.Readable;
+
+var timeoutFn = typeof setTimeoutOrig === 'function' ? setTimeoutOrig : setTimeout;
+
+/**
+ * ShakyStream will send data in 2 parts, pausing between parts.
+ */
+function ShakyStream(options) {
```
jeskew: +1 to having stream fixtures that aren't just event emitters.

test/request.spec.js
chrisradek: I don't think giving the user a truncated stream is really ok. I think this test was checking the behavior the SDK exhibited, rather than what was intended.

[Here](https://github.com/aws/aws-sdk-js/blob/master/lib/request.js#L606-L611), there's a check to see if the incoming data matched the content-length. It should be throwing an error when the content-length is less than the data received. When I was testing though, I discovered that when the body is larger than the content-length, node throws a `ParseError`. However, because we were swallowing these errors (due to already receiving response headers), and node.js still gave us access to the body up to the content-length, it appeared as though we could never detect when data streamed in exceeded the expected amount.

jeskew: I can't imagine that the data received up to `content-length` would be usable in most cases... you would end up with an unparsable partial XML or JSON document unless the service was returning a byte stream. I think the change in this PR is correct, but we should be on the lookout for issues where users were receiving partial byte streams and are now encountering stream errors.

jeskew:
:shipit:

PR #1517, opened 2017-05-22
Fixes issue found in React Native on Android when sending request bodies without a Content-Type

### LINKED ISSUE #1490: aws-sdk-js/react-native android network issue
screen shot 2017-04-28 at 2 10 33 pm
I am using aws-sdk-js/react-native branch to collect user data with aws mobile analytics service. Everything works fine with IOS. But for android for some reason XMLHttpRequest doesn't work.

Any help is appreciated.

chrisradek:
React Native's Android XMLHttpRequest implementation requires the Content-Type header to be set when the request contains a body. For some protocols, e.g. rest-json, the SDK doesn't set a Content-Type. Now, if a request contains a body and the Content-Type is not already set, nor is it a presigned url, it will be set to an empty string.

This change only affects the React Native build of the SDK.

Fixes #1490 

Tested on Android and iOS.

/cc @jeskew

### DIFF
--- a/lib/react-native/add-content-type.js
+++ b/lib/react-native/add-content-type.js
@@ -1,0 +1,15 @@
+function addContentType(req) {
+  var httpRequest = req.httpRequest || {};
+  var headers = httpRequest.headers;
+  // We don't want to force a content type on presigned urls
+  if (headers && !req.isPresigned()) {
+    if (httpRequest.body && !headers['Content-Type']) {
+      // React Native's android XHR requires Content-Type to be defined if there is a body
+      headers['Content-Type'] = '';
+    }
+  }
+}
+
+module.exports = {
+    addContentType: addContentType
+};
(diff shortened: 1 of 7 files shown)


lib/react-native/add-content-type.js
```diff
@@ -0,0 +1,15 @@
+function addContentType(req) {
+  var httpRequest = req.httpRequest || {};
+  var headers = httpRequest.headers;
+  // We don't want to force a content type on presigned urls
+  if (headers && !req.isPresigned()) {
+    if (httpRequest.body && !headers['Content-Type']) {
+      // React Native's android XHR requires Content-Type to be defined if there is a body
```
jeskew: Not a blocker, but it would be worthwhile to open an issue on React Native or comment on an existing one. (This looks related to https://github.com/facebook/react-native/issues/10362, but there might be another one that's still open.)

PR #1530, opened 2017-05-26
Allow empty string bodies

### LINKED ISSUE #1528: S3 empty file upload fails
I can't upload an empty file to S3.

Empty body (`""`) is valid. I think the comparison must be `self.body === undefined`.

https://github.com/aws/aws-sdk-js/blob/master/lib/s3/managed_upload.js#L260

jeskew:
Fixes #1528 

The managed uploader already allows the uploading of empty buffers, and providing an empty string is another way of expressing the same intention to upload a marker file. The uploader will still throw when `params.Body` is `null` or `false`

### DIFF
--- a/.changes/next-release/bugfix-S3-37b3e4ef.json
+++ b/.changes/next-release/bugfix-S3-37b3e4ef.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "S3",
+  "description": "Update ManagedUploader body verification to allow empty strings"
+}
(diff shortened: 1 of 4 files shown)


jeskew:
This PR also locks the version of chai used by the SDK to a specific major version, as our test suite does not appear to be compatible with today's release of [Chai 4.0.0](https://github.com/chaijs/chai/releases/tag/4.0.0).

iurisilvio:
It works for me. :+1:

chrisradek:
I gave it a shipit, but don't forget the changelog entry too!

chrisradek:
:shipit:!!!!

PR #1532, opened 2017-05-29
Check options parameter

renatogil:
Will fail if config is undefined.
```
.../node_modules/aws-sdk/lib/config.js:474
    if (options.accessKeyId && options.secretAccessKey) {
               ^
```

```
TypeError: Cannot read property 'accessKeyId' of undefined
    at Config.extractCredentials (.../node_modules/aws-sdk/lib/config.js:474:16)
```

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -495,7 +495,7 @@ AWS.Config = AWS.util.inherit({
    * @api private
    */
   extractCredentials: function extractCredentials(options) {
-    if (options.accessKeyId && options.secretAccessKey) {
+    if (options && options.accessKeyId && options.secretAccessKey) {
       options = AWS.util.copy(options);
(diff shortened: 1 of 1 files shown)


jeskew:
Hi @renatogil,

Thanks for your contribution! I think this case is covered by [an `if` statement in the config object's constructor](https://github.com/aws/aws-sdk-js/pull/1532/files#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R273); the `extractCredentials` function is marked `@api private` and is not meant to be called from outside of the `AWS.Config` class. What triggered the error this change is meant to address, and would it be possible to create a unit test that exercises that behavior?

jeskew:
Closing due to inactivity. Please feel free to reopen if you have any questions or concerns.

PR #1536, opened 2017-05-30
Call out in documentation that requests may only be sent once.

### LINKED ISSUE #1520: Retried S3 request uses a bogus hostname (the bucket name is repeated)
This is a duplicate of issue #585 and possibly regresses #636 
Using sdk version 2.55.0

Ive attached a log where I have placed a console.log(listener) in callListeners right before listener.apply and a console.log in populateUri that prints as console.log(b + "." + endpoint.hostname)

As is shown from the sequence, the host name is getting doubled up on retry regressing the solution from #636

This is on an ElasticBeanstalk Node Js instance running "64bit Amazon Linux 2016.09 v4.0.0 running Node.js" using the Amazon provided aws-sdk version 2.55.0 (taken from package.json in the node_modules/aws-sdk) folder.

[issue_sequence.txt](https://github.com/aws/aws-sdk-js/files/1020674/issue_sequence.txt)

jeskew:
This PR resolves #1520 by making explicit the expectation the requests only be sent once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ node_js:
   - "0.12"
   - "4.2"
   - "4.4"
-  - "5.11"
-  - "6.1"
-  - "node"
+  - "5"
+  - "6"
+  - "7"
+  - "8"
 
 before_install:
   - "if [[ `node -v` == v0.8.* ]]; then npm install -g npm || exit 0; fi"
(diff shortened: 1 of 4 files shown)


jeskew:
The browser builder rake task appears to be incompatible with NPM 5.0.0.

PR #1540, opened 2017-05-31
Add AppVeyor config

jeskew:
This PR adds configuration for AppVeyor so that PRs are automatically tested on Windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,11 @@
 environment:
   matrix:
-    nodejs_version: "4"
-    nodejs_version: "5"
-    nodejs_version: "6"
-    nodejs_version: "7"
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,16 @@
+environment:
+  nodejs_version: "6"
```
chrisradek: Should we test multiple platforms too?
https://www.appveyor.com/docs/lang/nodejs-iojs/#testing-under-multiple-versions-of-nodejs-or-iojs

jeskew: I think Travis gives us good coverage of multiple node versions, though there might be some environmental issues that only affect certain versions of node.

Issue #1538, opened 2017-05-31
documentation for leavePartsOnError scenario 


davidmaxwaterman:
I see the mention of `leavePartsOnError` option in the docs. It explains that it doesn't clear up the chunks to enable a more aggressive recovery, but it doesn't explain how the user might do such a recovery. I see some examples in issues here, eg:

https://github.com/aws/aws-sdk-js/issues/778

but that just does the recovery, and doesn't seem to do any clearup. The docs for abort() suggest that calling that would normally do the clearup, but that it doesn't do so if the 'leavePartsOnError' option is set. So, I am left wondering how, after an unsuccessful 'more aggressive' recovery, how I then actually use the SDK to clear up the parts.

I know the docs are quite, er, 'extensive', so perhaps I'm missing some reference somewhere. If not, is it possible to add some example code for this scenario?

davidmaxwaterman:
I am also wondering if the 'leavePartsOnError' could be used for a 'pause/resume' feature?

chrisradek:
@davidmaxwaterman 
We can add some documentation on how to use this. I revisited the issue you linked to, and there does appear to be a bug where resuming an aborted upload results in a smaller file being uploaded. In at least some cases, this occurs when all but 1 'part' has been uploaded. The next time the upload is triggered, since there is only 1 part left, it performs a `putObject` request instead of `uploadPart`. I'm working on a fix for that now.

ajredniwja:
Closing this issue since the feature was implemented.

PR #1550, opened 2017-06-05
Fixes S3 ManagedUpload leavePartsOnError issue when resuming Buffer uploads

chrisradek:
Addresses #1538 & #778 

When the S3 ManagedUpload is configured with `leavePartsOnError` set to `true`, it does not clean up data that has been uploaded to S3.

Calling `send` multiple times when the Body is a Buffer should attempt to upload the data that was not already saved to S3.

There is an issue where calling `send` a 2nd time would instead upload the 1st part using a `putObject` request, resulting in a 5MB file being uploaded, instead of the full file. This happens when all parts have been queued for upload. On a subsequent `send`, the uploader considers the payload to be 'chunked', and assumes the first part it encounters to be the full file.

This change resets the `isDoneChunking` state when an upload fails/is aborted.

Note that resuming an upload still will not work for streams, since streams are exhausted on failure. Supporting resume on streams should be an enhancement we make at some point.

/cc @jeskew

### DIFF
--- a/lib/s3/managed_upload.js
+++ b/lib/s3/managed_upload.js
@@ -590,8 +590,18 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
       self.body.resume();
     }
 
+    // cleanup multipartReq listeners
+    if (self.multipartReq) {
+      self.multipartReq.removeAllListeners('success');
+      self.multipartReq.removeAllListeners('error');
+      self.multipartReq.removeAllListeners('complete');
+      delete self.multipartReq;
+    }
+
     if (self.service.config.params.UploadId && !self.leavePartsOnError) {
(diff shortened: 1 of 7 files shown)


lib/s3/managed_upload.js
```diff
@@ -590,8 +590,18 @@ AWS.S3.ManagedUpload = AWS.util.inherit({
       self.body.resume();
     }
 
+    // cleanup multipartReq listeners
+    if (self.multipartReq) {
```
jeskew: CodeCov is being a tad too extreme on this PR, but I think it would be helpful to add a unit test that ensures listeners are removed and that `isDoneChunking` is left `false`.

PR #1558, opened 2017-06-08
Switches S3 default signatureVersion from v2 to v4

chrisradek:
Original attempt: #1023 
Reverted attempt: #1085 

/cc @jeskew


lib/services/s3.js
```diff
@@ -143,6 +154,24 @@ AWS.util.update(AWS.S3.prototype, {
   /**
    * @api private
    */
+  validateBucketName: function validateBucketName(req) {
+    var service = req.service;
+    var signatureVersion = service.getSignatureVersion(req);
+    // Only validate keys when using sigv4
```
jeskew: Nit: "Only validate **buckets**"

chrisradek: Nice catch!

PR #1563, opened 2017-06-12
Fix markdown in README.md

mmorearty:
The markdown in README.md had a few problems that caused it to be displayed incorrectly by some markdown parsers, even though GitHub displayed it correctly. For example, http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/index.html displayed the lists incorrectly.


jeskew:
Thanks for your contribution!

PR #1587, opened 2017-06-20
S3 bucket slashes

chrisradek:
Updates `s3.upload` to use the same signature version as the client calling it.

Also allows forward slashes in S3 bucket names when an operation supports a Bucket and Key parameter. This is to maintain backwards compatibility with code written when signature version 2 was the default.

/cc @jeskew

### DIFF
--- a/features/s3/objects.feature
+++ b/features/s3/objects.feature
@@ -215,3 +215,63 @@ Feature: Working with Objects in S3
   Scenario: Error handling
     Given I put "data" to the invalid key ""
     Then the status code should be 400
+
+  @bucket-slashes
+  Scenario: Sigv4 Bucket with trailing slash
+    Given I use signatureVersion "v4"
+    When I put "" to the key "" with bucket suffix "/"
+    Then the object "/" should exist
+  
+  @bucket-slashes
+  Scenario: Sigv4 Bucket with dobule trailing slashes
+    Given I use signatureVersion "v4"
+    When I put "" to the key "" with bucket suffix "//"
+    Then the object "//" should exist
+
+  @bucket-slashes
+  Scenario: Sigv4 Bucket with slashes without trailing slash
+    Given I use signatureVersion "v4"
+    When I put "" to the key "" with bucket suffix "/a/b"
+    Then the object "a/b/" should exist
+
+  @bucket-slashes
+  Scenario: Sigv4 Bucket with slashes with Key
+    Given I use signatureVersion "v4"
+    When I put "" to the key "foo" with bucket suffix "/a/b"
+    Then the object "a/b/foo" should exist
+
+  @bucket-slashes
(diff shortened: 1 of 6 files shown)


features/s3/objects.feature
```diff
@@ -215,3 +215,63 @@ Feature: Working with Objects in S3
...
+  Scenario: Sigv2 Bucket with slashes with Key
+    Given I use signatureVersion "v2"
+    When I put "" to the key "foo" with bucket suffix "/a/b"
+    Then the object "a/b/foo" should exist
+
+  @bucket-slashes
+  Scenario: Sigv2 Bucket with trailing slashes with Key
+    Given I use signatureVersion "v2"
+    When I put "" to the key "foo" with bucket suffix "/a/b/"
+    Then the object "a/b//foo" should exist
```
jeskew: We can't change it, but this behavior does not seem completely intuitive. Bash at least will treat one or more slashes in a sequence as a single directory separator. Just something to keep in mind for any future high-level interfaces, especially those that allow/encourage users to interact with S3 as if it were a filesystem.

test/services/s3.spec.js
```diff
@@ -205,11 +205,62 @@
             done();
           });
         });
+        it('using sigv4 and the "Bucket" parameter contains forward slashes and "Key" exists', function(done) {
```
jeskew: This test is using sigv2

features/s3/objects.feature
chrisradek: Agreed, and I can't imagine this would be desired behavior. I wonder if other SDKs collapse multiple slashes into one.

lib/services/s3.js
```diff
@@ -162,10 +162,20 @@ AWS.util.update(AWS.S3.prototype, {
...
+    var slashIndex = bucket && bucket.indexOf('/');
+    if (bucket && slashIndex >= 0) {
+      if (typeof key === 'string') {
+        req.params = AWS.util.copy(req.params);
+        // Need to include trailing slash to match sigv2 behavior
+        var prefix = bucket.substr(slashIndex + 1) || '';
+        req.params.Key = prefix + '/' + key;
+        req.params.Bucket = bucket.substr(0, slashIndex);
+      } else {
+        var msg = 'Bucket names cannot contain forward slashes. Bucket: ' + bucket;
```
chrisradek: If a user did not supply a key at all, then they would receive a missing required param error. That said, it is possible to turn off param validation so maybe I should just make sure Key exists as a possible parameter.

test/services/s3.spec.js
chrisradek: Copy paste error, I'll fix that.

lib/services/s3.js
jeskew: That sounds like a fairly extreme edge case. I'd be fine with leaving as is.

PR #1596, opened 2017-06-24
Roll buffer dependency back to 4.9.1

### LINKED ISSUE #1565: IE 10 JS SDK not working. buf.write is undefined
When trying to authenticate using JS SDK clientside in IE 10, we are receiving an error in the console saying the buf.write is undefined.

I believe it has something to do with the fact IE10 doesn't allow you to assign to the prototype, but in the `dist/aws-sdk.js` file there are two `createBuffer` functions being defined; one that checks `TYPED_ARRAY_SUPPORT` and one that doesn't

1. Line 111334
```
function createBuffer (that, length) {
  if (kMaxLength() < length) {
    throw new RangeError('Invalid typed array length')
  }
  if (Buffer.TYPED_ARRAY_SUPPORT) {   //<-- Checks for support (IE10) before trying to set proto, which is good.
    that = new Uint8Array(length)
    that.__proto__ = Buffer.prototype
  } else {
    if (that === null) {
      that = new Buffer(length)
    }
    that.length = length
  }

  return that
}
```

2. Line 112985
```
function createBuffer (length) {
  if (length > K_MAX_LENGTH) {
    throw new RangeError('Invalid typed array length')
  }
  var buf = new Uint8Array(length)
  buf.__proto__ = Buffer.prototype //<- cannot do this in ie10, and because this function is defined after the first one, this is the implementation that runs.
  return buf
}
```

jeskew:
This PR fixes #1565 by rolling back our buffer dependency from 5.0.6 to 4.9.1. Buffer 5.x explicitly only supports IE11, whereas this package supports IE10.

The original impetus for moving to 5.0.6, #1467, will need to be addressed in another way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "webpack": "^1.15.0"
   },
   "dependencies": {
-    "buffer": "5.0.6",
+    "buffer": "4.9.1",
     "crypto-browserify": "1.0.9",
     "jmespath": "0.15.0",
     "querystring": "0.2.0",


chrisradek:
@jeskew 
So will reverting this change break android 4.4.2 users?

chrisradek:
Ship it!

PR #1604, opened 2017-06-29
Prevents automatically filling in Content-Length header for unsigned operations

chrisradek:
The SDK currently will populate the Content-Length header for all requests if it isn't provided. For operations that accept a stream, such as S3 or Glacier, this has limited us to only allowing streams sourced from a file to be used.

The LexRuntime service does not require knowing the Content-Length of a request ahead of time, and can accept a stream as an input to postContent. This change uses the authtype field of an operation to determine if the body is unsigned, and then skips automatically populating the Content-Length. Currently Lex is the only service that uses this authtype field to indicate it uses an unsigned body. Aside from services like S3/Glacier, we can infer that if a service expects a payload to be unsigned, they likely don't expect to know the size of the payload upfront.

/cc @jeskew

### DIFF
--- a/test/event_listeners.spec.js
+++ b/test/event_listeners.spec.js
@@ -170,6 +170,19 @@
         contentLength = function(body) {
           return sendRequest(body).httpRequest.headers['Content-Length'];
         };
+        it('ignores Content-Length for operations with an unsigned authtype', function(done) {
+          var service = new AWS.Lambda();
+          service.api.operations.updateFunctionCode.authtype = 'v4-unsigned-body';
+          req = service.makeRequest('updateFunctionCode', {
+            FunctionName: 'fake',
+            ZipFile: new Buffer('fake')
+          });
+          req.runTo('sign', function(err) {
+            expect(typeof req.httpRequest.headers['Content-Length']).to.equal('undefined');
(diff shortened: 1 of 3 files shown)


test/event_listeners.spec.js
```diff
@@ -170,6 +170,19 @@
         contentLength = function(body) {
           return sendRequest(body).httpRequest.headers['Content-Length'];
         };
+        it('ignores Content-Length for operations with an unsigned authtype', function(done) {
```
jeskew: Not a blocker, but the tests are much easier to read with empty lines separating them.

test/event_listeners.spec.js
```diff
@@ -170,6 +170,19 @@
...
+            FunctionName: 'fake',
+            ZipFile: new Buffer('fake')
+          });
+          req.runTo('sign', function(err) {
+            expect(typeof req.httpRequest.headers['Content-Length']).to.equal('undefined');
+            delete service.api.operations.updateFunctionCode.authtype;
+            done();
+          });
+        });
         it('builds Content-Length in the request headers for string content', function() {
```
jeskew: ditto

PR #1618, opened 2017-07-07
Updates React Native build to allow same style imports as JS SDK

chrisradek:
This change makes it possible to use the SDK in react native environments by importing the same as in node.js or the browser, instead of the current way.

```javascript
/* New way */
import * as AWS from 'aws-sdk';
import S3 from 'aws-sdk/clients/s3';
```
```javascript
/* Current way */
import * as AWS from 'aws-sdk/dist/aws-sdk-react-native';
```

This comes with a few benefits:
1. Typings will now work without any workarounds
2. Libraries that currently depend on the SDK will not need to change how they import the SDK when running in a react native environment.
3. Individual services can be imported, which should reduce application sizes.

Because the SDK has circular dependencies throughout, and react native's bundler doesn't support that, I had to get a little fancy with the react-native build process. Nearly every file imports `lib/core.js`, which itself creates circular dependencies, so I used webpack to create a bundle of just that. Any file that could be an entry point (such as the `clients` directory) has been updated to reference classes from `lib/core.js` instead of the submodule directly, to prevent the metro-bundler from choking on circular deps.

/cc @jeskew

### DIFF
--- a/dist-tools/client-creator.js
+++ b/dist-tools/client-creator.js
@@ -159,14 +159,12 @@ ClientCreator.prototype.generateAllServicesSource = function generateAllServices
   var self = this;
   var code = '';
   code += 'require(\'../lib/node_loader\');\n';
-  //code += 'var AWS = require(\'../lib/core\');\n\n';
   code += 'module.exports = {\n';
 
   services.forEach(function(service, idx) {
     var className = metadata[service].name;
     var tab = '  ';
     var isLast = idx === services.length - 1;
-    //code += self.generateDefinePropertySource('AWS', service, className);
     code += self.tabs(1) + className + ': require(\'./' + service + '\')' + (isLast ? '' : ',') + '\n';
   });
   code += '};';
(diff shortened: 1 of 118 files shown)


dist-tools/client-creator.js
```diff
@@ -159,7 +159,7 @@ ClientCreator.prototype.generateAllServicesSource = function generateAllServices
   var self = this;
   var code = '';
   code += 'require(\'../lib/node_loader\');\n';
-  code += 'var AWS = require(\'../lib/core\');\n\n';
+  //code += 'var AWS = require(\'../lib/core\');\n\n';
```
chrisradek: Thanks for spotting this!

lib/react-native-loader.js
```diff
@@ -8,9 +7,21 @@ util.querystring = require('querystring/');
 util.environment = 'js-react-native';
 
 var AWS = require('./core');
+module.exports = AWS;
+
+// // react-native specific modules
+ AWS.util.crypto.lib = util.crypto.lib;
```
chrisradek: I was actually able to just import `lib/core.js` on the first line and mutate `util` directly on that, instead of pulling in `util`.

Issue #1628, opened 2017-07-14
getObject + callback + createReadStream = corrupted file downloads


broofa:
While using getObject() + createReadStream() to save an S3 file, we found we needed to pick up the case where the s3 key was not recognized.  The only way to do this is to provide a callback function to `getObject()`, thus we ended up with the following code: 

```
    const file = fs.createWriteStream(filepath);
    const s3Obj = s3.getObject(params, (err, data) => {
      if (err) callback(err);
    });
    s3Obj.createReadStream(). stream.pipe(file);
```

***However***... this resulted in corrupted files being saved to disk.   In our case, the S3 file was expected to be a ~640KB image file but what we ended up with was a ~900KB file where the first and last parts were correct, but that ~260-300KB of additional garbage in the middle.

After some experimentation, we found this bug only happened if we provided the getObject() callback.  Without the callback the file saves correctly.  Or, alternatively, getting the content via `data.Body`  in the callback also works.

So, the issue here is that 1) it's possile to use getObject+createReadStream in a way that produces corrupt files, and 2) the documentation does not describe how to capture invalid key errors in the context of getObject()+createReadStream() 

cc @pipemachine @chrisgilmerproj

chrisradek:
@broofa 
When a request is called with a callback and createReadStream, it triggers a race condition where 2 requests are opened at once, but the data from both are piped to the same destination. There are some details about this [here](https://github.com/aws/aws-sdk-js/issues/959#issuecomment-209066648).

Instead of passing a callback function to `s3.getObject` to catch the error from the service, you should attach an error event listener to the s3 stream.

Example:
```javascript
var fs = require('fs');
var path = require('path');
var S3 = require('aws-sdk/clients/s3');

var s3 = new S3({
    region: 'us-west-2'
});

var name = 'fake.file';

var fileStream = fs.createWriteStream(path.join(__dirname, name));
var s3Stream = s3.getObject({Bucket: 'BUCKET', Key: name}).createReadStream();

// Listen for errors returned by the service
s3Stream.on('error', function(err) {
    // NoSuchKey: The specified key does not exist
    console.error(err);
});

s3Stream.pipe(fileStream).on('error', function(err) {
    // capture any errors that occur when writing data to the file
    console.error('File Stream:', err);
}).on('close', function() {
    console.log('Done.');
});
```

Our plan is to throw an error when mixing callbacks with streams are attempted, since doing so will usually cause unexpected behavior.

broofa:
[...]

Issue #959, opened 2016-04-12
s3.getObject - passing a callback and creating a readable stream causes corrupt downloads


inversion:
This is reliably reproducible with the following code. I know it's weird to use the callback and stream, and the callback was there as a consequence of over-defensive programming, but this shouldn't fail silently in this way.

I'm guessing some data doesn't hit the stream as it's assumed to be handled in the callback, while I would expect it to hit both.

Environment: 

```
Node v5.8.0
Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 x86_64
aws-sdk 2.3.3
Bucket in eu-west-1 with versioning enabled. File stored as STANDARD_IA
```

Code to reproduce:

```
'use strict';

const cp = require( 'child_process' );
const fs = require( 'fs' );
const aws = require( 'aws-sdk' );

// Configure these
const file = 'blah.zip';
const key = 'blah.zip';
const hash = 'known_sha256_hash';
const bucket = 'yourbucket';

const s3 = new aws.S3( {
    params: {
        Bucket: bucket
    },
    logger: process.stdout
} );

const get = function() {

    var callback = ( err ) => {

        if ( err ) {
            console.log( err );
        }

    };

    // Uncomment me to fix it
    // callback = undefined;

    return new Promise( ( fulfil, reject ) => {

        var writeStream = fs.createWriteStream( file );

        var req = s3.getObject( {
            Key: key
        }, callback );

        req.on( 'error', reject );

        const readStream = req.createReadStream();

        readStream.on( 'error', reject );

        readStream.pipe( writeStream );

        writeStream.on( 'error', reject );

        writeStream.once( 'finish', () => {

            fulfil();
        } );

    } );

};

get().then( () => {

    console.log( file );

    const actual = cp.execSync( 'shasum -a 256 ' + file, {
        encoding: 'utf-8'
    } ).split( ' ' )[ 0 ];

    console.log( actual === hash, actual, hash );

    if ( actual !== hash ) {

        process.exit( 1 );

    }
} );
```

inversion:
Also reproducible with aws-sdk 2.2.42

chrisradek:
@inversion 
[...]

PR #1631, opened 2017-07-17
Triggering send on an AWS.Request multiple times will cause an error to be thrown

chrisradek:
Currently, triggering a `send` on `AWS.Request` multiple times can result in multiple http requests to be sent. This can lead to data corruption in some cases, since each http request will contribute buffer chunks to the parent AWS.Request.

If `AWS.Request.send` and `AWS.Request.createReadStream` are called using the same request, the stream will emit an error that can be listened for. If AWS.Request.send is called multiple times, then each call beyond the first will throw an error that needs to be caught.

In the latter case, I could not pass the error to the callback the user provides without failing all calls. It might make sense to completely fail a request if it is called multiple times.

/cc @jeskew 

Addresses #1628 and #959

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -327,7 +327,7 @@ AWS.Request = inherit({
     this.response = new AWS.Response(this);
     this._asm = new AcceptorStateMachine(fsm.states, 'validate');
     this._haltHandlersOnError = false;
-
+    this._didSend = false;
     AWS.SequentialExecutor.call(this);
     this.emit = this.emitEvent;
   },
@@ -357,6 +357,7 @@ AWS.Request = inherit({
    *     request.send();
    */
   send: function send(callback) {
+    this.preventMultipleCalls();
     if (callback) {
       // append to user agent
       this.httpRequest.appendToUserAgent('callback');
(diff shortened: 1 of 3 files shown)


lib/request.js
```diff
@@ -563,13 +578,20 @@ AWS.Request = inherit({
    * @!macro nobrowser
    */
   createReadStream: function createReadStream() {
+    this.preventMultipleCalls();
```
chrisradek: I might actually remove this line. This will cause an error to be thrown if createReadStream is called after a request is sent, but it would need to be wrapped in a try/catch. That might be confusing if the user expects to listen for an error on the stream instead.

lib/request.js
```diff
@@ -357,19 +357,34 @@ AWS.Request = inherit({
...
+    this._didSend = true;
     this.runTo();
 
     return this.response;
   },
 
   /**
+   * @api private
+   */
+  preventMultipleCalls: function preventMultipleCalls() {
```
jeskew: Maybe instead of throwing, this method could take a callback? That would allow the RequestAlreadyTriggeredError to be handled by user callbacks when `send` is called or emitted onto streams when `createReadStream` is called.

ajredniwja:
Hey @chrisradek,

Are you still working on this?

PR #1673, opened 2017-08-08
FIX: error TS2503: Cannot find namespace 'AWS'

theluk:
the following file triggers an error when transpiling. it uses `Promise<AWS.Credentials>` as a type even though `AWS` is not declared. I believe only `Credentials` is wanted

https://github.com/aws/aws-sdk-js/pull/1672/files#diff-171e32a9b2d89468799d3311ac2c638b

my tsconfig

```
{
  "compilerOptions": {
    "sourceMap": true,
    "noImplicitAny": true,
    "module": "commonjs",
    "target": "es5",
    "jsx": "react",
    "allowJs": true,
    "lib": [
        "es5",
        "es2015.promise"
    ]
  }
}
```

rule in webpack

```
{
            test: /\.tsx?$/,
            use: 'ts-loader',
            exclude: [
                path.resolve(__dirname, 'node_modules'),
                path.resolve(__dirname, 'bower_components')
            ]
        }
```

### DIFF
--- a/lib/credentials/credential_provider_chain.d.ts
+++ b/lib/credentials/credential_provider_chain.d.ts
@@ -12,7 +12,7 @@ export class CredentialProviderChain extends Credentials {
     /**
      * Return a Promise on resolve() function
      */
-    resolvePromise(): Promise<AWS.Credentials>;
+    resolvePromise(): Promise<Credentials>;
     /**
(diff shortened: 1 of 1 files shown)


rucas:
@theluk Getting this error too on release v2.95.0 went down a version and seems to work. Would love to get this merged!

jeskew:
Thanks for your contribution!

Issue #1612, opened 2017-07-03
Asynchronous getSignedUrl callback is not really asynchronous


colinhemphill:
The AWS docs offer both a synchronous and asynchronous version of the `getSignedUrl` request. I default to using the async option since it has better authentication support. However, it appears that if you don't use one of the asynchronous authentication methods, the async version of the request isn't _actually_ asynchronous.

Here's an example.

```
async.eachLimit(items, 5, function (item, callback) {
  s3.getSignedUrl('getObject', {
    Bucket: 'my-bucket',
    Key: item.s3Key
  }, function (err, url) {
    if (err) return callback(err);
    // do something with the url
    return callback();
  });
}, function (err) {
  // done with async loop
});
```

If `s3.getSignedUrl(operation, params, callback)` was really asynchronous when you use a callback, the above code would work great. With a large enough array of `items`, this fails and gives me a `RangeError: Maximum call stack size exceeded` error from the aws-sdk module.

If I wrap the callback in `async.setImmediate()`, the above works as I expect it to. This indicates to me that the callbacks are happening on the same tick in the event loop as the `getSignedUrl` function that's calling them. Can we force this to wait for the next tick if you're using the asynchronous version?

jeskew:
Resolved in #1674 and #1676.

PR #1674, opened 2017-08-09
Force setImmediate on s3.getSignedUrl when a callback is provided

colinhemphill:
This is to address issue #1612.

If a callback is not provided to `s3.getSignedUrl`, then it returns the url as usual. If `process.nextTick` is a valid function, and an asynchronous callback is expected , it forces the callback to return on the next tick. As a fallback for other environments, it wraps the callback in a `setTimeout(function() {}, 0)`.

I've tested this update in my own Node environment with both a synchronous and an asynchronous call to `s3.getSignedUrl()`. Making many calls to this service in an `async.each()` loop now works as expected without exceeding the stack size. I also ran the `npm test` script.

### DIFF
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -772,11 +772,11 @@ AWS.util.update(AWS.S3.prototype, {
     var expires = params.Expires || 900;
     delete params.Expires; // we can't validate this
     var request = this.makeRequest(operation, params);
-    var hasNextTick = typeof process === 'object' && typeof process.nextTick === 'function';
+    var hasSetImmediate = typeof setImmediate === 'function' && setImmediate;
     if (!callback) {
       return request.presign(expires, callback);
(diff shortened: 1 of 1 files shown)


lib/services/s3.js
```diff
@@ -772,7 +772,18 @@ AWS.util.update(AWS.S3.prototype, {
     var expires = params.Expires || 900;
     delete params.Expires; // we can't validate this
     var request = this.makeRequest(operation, params);
-    return request.presign(expires, callback);
+    var hasNextTick = typeof process === 'object' && typeof process.nextTick === 'function';
```
jeskew: Could this check for and use `setImmediate` instead of `process.nextTick`? `setImmediate` is also supported in IE and Edge.

colinhemphill: Good call - just updated 👍

PR #1676, opened 2017-08-09
Add test for #1674

jeskew:
This PR includes #1674 and adds a test to appease CodeCov. The S3 tests were a bit of a mess following the conversion from coffeescript to JS for our test suite, so that conversion is included as a separate commit.


jeskew:
@AllanFly120 when you review, you can ignore the commit with the message "Un-coffeescript test/services/s3.spec.js." That commit is huge and consists entirely of removing artifacts of the coffeescript compilation process from `test/services/s3.spec.js`.

colinhemphill:
Good idea on the util method. Thanks for this 👍

jeskew:
@colinhemphill Appeasing CodeCov turned out to be more involved than I thought it would be. Thanks for the original work (and identifying the issue, too)

PR #1711, opened 2017-09-08
Add AWS.S3.getSignedUrlPromise

### LINKED ISSUE #1008: S3 Get Signed URL accepts callback but not promise
I'm trying to get away from the callback pattern and go with the promise pattern since Async functions will rely on promises in esnext. If it's actually synchronous, why provide a callback option? If callback is required, shouldn't promise be an option?

IsaiahJTurner:
This closes #1008. Few questions:
1. Do I need to add anything to `/ts/s3.ts`?
2. It seems odd that `S3.getSignedUrl` throws even when a callback is provided. I'd expect the error to be passed to the callback. There does not appear to be tests for this. The `errors if ContentLength is passed as parameter` test uses the no-callback version but there is no test for errors on `getSignedUrl`. Because of this, I only added basic error expectation testing. The error I would be expecting is `ContentLength is not supported in pre-signed URLs` but the error I get is `[TypeError: Cannot read property 'httpMethod' of undefined] to be a function`. This is because the getSignedUrlPromise throws errors directly, not through the promise.
3. I didn't test for the actual credentials refresh request happening using a spy or any of the other features in `getSignedUrl` since it seemed redundant as `getSignedUrlPromise` calls `getSignedUrl`.

Appreciate any feedback or comments on things I missed/did incorrectly. This is my first time contributing. Thanks! Tagging @jeskew.

### DIFF
--- a/lib/services/s3.d.ts
+++ b/lib/services/s3.d.ts
@@ -13,6 +13,11 @@ export class S3Customizations extends Service {
     getSignedUrl(operation: string, params: any): string;
 
     /**
+     * Returns a 'thenable' promise that will be resolved with a pre-signed URL for a given operation name.
+     */
+    getSignedUrlPromise(operation: string, params: any): Promise<void>;
+
+    /**
      * Get the form fields and target URL for direct POST uploading.
      */
     createPresignedPost(
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -802,13 +802,9 @@ AWS.util.update(AWS.S3.prototype, {
    *      work for.
    *   @option params Expires [Integer] (900) the number of seconds to expire
    *      the pre-signed URL operation in. Defaults to 15 minutes.
-   *   @param callback [Function] if a callback is provided, this function will
-   *      pass the URL as the second parameter (after the error parameter) to
-   *      the callback function.
-   *   @callback fulfilledCallback function()
-   *     Called if the promise is fulfilled. When this callback is called, it
-   *     means refreshed credentials information has been loaded into the object
-   *     (as the `accessKeyId`, `secretAccessKey`, and `sessionToken` properties).
+   *   @callback fulfilledCallback function(url)
+   *     Called if the promise is fulfilled.
+   *     @param url [String] the signed url
    *   @callback rejectedCallback function(err)
    *     Called if the promise is rejected.
    *     @param err [Error] if an error occurred, this value will be filled
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -784,8 +784,7 @@ AWS.util.update(AWS.S3.prototype, {
 
   /**
    * @!method  getSignedUrlPromise()
-   *   Returns a 'thenable' promise.
-   *   Get a pre-signed URL for a given operation name.
+   *   Returns a 'thenable' promise that will be resolved with a pre-signed URL for a given option name.
    *
    *   Two callbacks can be provided to the `then` method on the returned promise.
    *   The first callback will be called if the promise is fulfilled, and the second
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -782,6 +782,56 @@ AWS.util.update(AWS.S3.prototype, {
     }
   },
 
+  /**
+   * @!method  getSignedUrlPromise()
+   *   Returns a 'thenable' promise.
+   *   Get a pre-signed URL for a given operation name.
+   *
+   *   Two callbacks can be provided to the `then` method on the returned promise.
+   *   The first callback will be called if the promise is fulfilled, and the second
+   *   callback will be called if the promise is rejected.
+   *   @note Not all operation parameters are supported when using pre-signed
+   *      URLs. Certain parameters, such as `SSECustomerKey`, `ACL`, `Expires`,
+   *      `ContentLength`, or `Tagging` must be provided as headers when sending a
+   *      request. If you are using pre-signed URLs to upload from a browser and
+   *      need to use these fields, see {createPresignedPost}.
+   *   @param operation [String] the name of the operation to call
+   *   @param params [map] parameters to pass to the operation. See the given
+   *      operation for the expected operation parameters. In addition, you can
+   *      also pass the "Expires" parameter to inform S3 how long the URL should
+   *      work for.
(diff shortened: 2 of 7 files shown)


lib/services/s3.js
```diff
@@ -782,6 +782,52 @@ AWS.util.update(AWS.S3.prototype, {
     }
   },
 
+  /**
+   * @!method  getSignedUrlPromise()
+   *   Returns a 'thenable' promise.
```
chrisradek: Can you amend this to state a URL will be returned?

Something like:

> Returns a 'thenable' promise that will be resolved with a pre-signed URL.

IsaiahJTurner:
Should I add anything to `/.changes/next-release/feature-getSignedUrl-9194166f.json`? Couldn't find any documentation for how that works/naming convention.

lib/services/s3.js
```diff
@@ -1068,3 +1114,20 @@ AWS.util.update(AWS.S3.prototype, {
     return uploader;
   }
 });
+
+
+/**
+ * @api private
+ */
+AWS.S3.addPromisesToClass = function addPromisesToClass(PromiseDependency) {
+  this.prototype.getSignedUrlPromise = AWS.util.promisifyMethod('getSignedUrl', PromiseDependency);
```
chrisradek: The `AWS.util.promisifyMethod` function currently only works for functions that accept a callback function as the first parameter. For example, `AWS.Request.send` and `AWS.Credentials.get` both accept just a callback.

The `s3.getSignedUrl` method accepts an operation name, params, and a callback, so using this method won't work. You should be able to set `getSignedUrlPromise` to a function that returns a new Promise. This promise can simply wrap the getSignedUrl function.

test/services/s3.spec.js
```diff
@@ -2541,6 +2541,36 @@ describe('AWS.S3', function() {
...
+      var catchFunction, err;
+      err = null;
+      catchFunction = function(e) {
+        err = e;
+      };
+      beforeEach(function() {
+        AWS.util.addPromises(AWS.S3, Promise);
+        err = null;
+      });
+      it('resolves when getSignedUrl is successful', function() {
```
chrisradek: This test is actually returning a false positive. The `catchFunction` is being called, but since there are no `expect` calls within it that can fail, mocha assumes the test passes.

chrisradek:
Regarding your question about the TypeScript tests, can you add the definition for getSignedUrlPromise to https://github.com/aws/aws-sdk-js/blob/master/lib/services/s3.d.ts, and then in `/ts/s3.ts` just attempt to use the new method (this makes sure that the TypeScript compiler recognizes the new definition correctly)?

Let us know if you need help adding the TypeScript typings. It can look foreign if you haven't had to deal with them before.

lib/services/s3.js
IsaiahJTurner: Happy to do that but that solution could lead replicated code down the line. For example, if/when this functionality is added to `AWS.CloudFront.Signer.getSignedUrl`, we'd need to wrap it with the same logic.

What do you think about modifying `AWS.util.promisifyMethod` so that the generated promise function accepts arguments? 

Something like this:
```js
function promisifyMethod(methodName, PromiseDependency) {
    return function promise() {
        var self = this;
        var args = Array.prototype.slice.call(arguments);
        return new PromiseDependency(function(resolve, reject) {
            args.push(function(err, data) {
                if (err) {
                    reject(err);
                } else {
                    resolve(data);
                }
            });
            self[methodName].apply(null, args);
        });
    };
}
```
My solution does lead to one of two downsides:
- Supplying too many arguments will give unexpected behavior like a "callback is not a function" error.
- To prevent that with a more informative error like "Too many arguments supplied" I'd need to read `self[methodName].length` to get the expected arguments count but this would be incompatible with any functions that support the arguments style `Array.prototype.push([element1[, ...[, elementN]]])` since `Array.prototype.push.length` would only return 1. I'm not sure if any functions in the SDK work like this.

Regardless of which downside chosen, I still think improving `AWS.util.promisifyMethod` rather than wrapping the function directly is the best option since it is the most extensible. Also, how `AWS.util.promisifyMethod` currently works in contrast with how other `promisify` functions within other JS libraries like Bluebird, es6-promisify, and promisify-node work.

IsaiahJTurner: Went ahead and pushed my concept. Let me know what you think, can easily change it back and use your idea.

lib/services/s3.d.ts
```diff
@@ -13,6 +13,11 @@ export class S3Customizations extends Service {
     getSignedUrl(operation: string, params: any): string;
 
     /**
+     * Returns a 'thenable' promise that will be resolved with a pre-signed URL for a given operation name.
+     */
+    getSignedUrlPromise(operation: string, params: any): Promise<void>;
```
jeskew: Shouldn't the return value be `Promise<string>`?

test/services/s3.spec.js
```diff
@@ -2536,11 +2536,61 @@ describe('AWS.S3', function() {
...
+      afterEach(function(done) {
+        AWS.util.date.getDate = date;
+        done();
+      });
 
+      it('resolves when getSignedUrl is successful', function() {
+        return s3.getSignedUrlPromise('getObject', {
+          Bucket: 'bucket',
+          Key: 'key'
+        }).then(resolveFunction).catch(catchFunction).then(function() {
```
jeskew: Rather than setting the resolved value or error to function-scoped variables, why not just make assertions in the success handler passed to `then` and throw in the error handler passed to `catch`?

The file to which these tests are being added was originally written in CoffeeScript and subsequently replaced with its compiled JavaScript form, so many of the odd conventions used were injected by the CoffeeScript compiler.

lib/util.js
```diff
@@ -793,14 +793,16 @@ var util = {
   promisifyMethod: function promisifyMethod(methodName, PromiseDependency) {
     return function promise() {
       var self = this;
+      var args = Array.prototype.slice.call(arguments);
```
jeskew: Since most methods take a variable number of arguments, I'm not sure if this is the best approach. I would just add a simple wrapper for `getSignedUrl` instead.

trivikr:
Fixed in #2827

PR #1728, opened 2017-09-20
Mention in documentation examples that strings provided for binary properties will be base64 encoded

jeskew:
It can be confusing if you're providing base64 encoded data to an operation to find out that is being base64 encoded a second time. This change adds a comment to the usage samples in the documentation that points out that base64 strings will be encoded on the user's behalf.
base64-callout-screenshot


chrisradek:
I think applying this change to all binary shapes instead of just base64 shapes actually makes some things more confusing. For example, `s3.putObject` now sounds like it will base64 encode the body if I pass it text.

s3-docs

jeskew:
You're absolutely right; this note should only appear for non-streaming binary properties. In the screenshot below, `Body` is streaming and `SSECustomerKey` is not:
base64-callout-screenshot

Issue #1737, opened 2017-10-03
Upgrade uuid to version 3.1.0


B3ll89:
Will it be possible to upgrade `uuid` to 3.1.0?
It has been out for quite a while now and I'm now aware of any *actual* changes.

chrisradek:
The uuid dependency has been updated and is merged into master. The next version of the SDK will include this change.

3.1.0 did not include any code changes, just documentation. Did you need it upgraded to avoid duplicate copies, or some other reason?

B3ll89:
It is indeed a duplication issue.

Thank you very much!

PR #1738, opened 2017-10-03
Update uuid dependency to 3.1.0

chrisradek:
The UUID dependency was updated from version 3.0.1 to 3.1.0. This update contains zero code changes (probably should have been a patch update?), just updates to its documentation. Updating it so projects don't have to pull in multiple versions of the same package.

Also fixed mocha to a specific version because I was seeing deprecation warnings.

Addresses #1737


jeskew:
There's no way that coverage number can be right given the scope of this PR. :shipit:

PR #1762, opened 2017-10-17
update managed_upload validateBody method

againer:
- API users can pass a Stream object that does not have `.read(...)`.


againer:
*fixing tests.

PR #1785, opened 2017-10-30
Export CognitoIdentityCredentials

villelahdenvuo:
In order to reduce frontend bundle size, it is nice to be able to import things one-by-one instead of importing the whole SDK.
However using CognitoIdentityCredentials alone is not possible because it is not exported even though the types say so.

```typescript
import * as S3 from 'aws-sdk/clients/s3';
import { CognitoIdentityCredentials } from 'aws-sdk/lib/credentials/cognito_identity_credentials';

console.log(S3, CognitoIdentityCredentials);
```

This logs: `f () {...} undefined`

### DIFF
--- a/lib/credentials/cognito_identity_credentials.js
+++ b/lib/credentials/cognito_identity_credentials.js
@@ -375,3 +375,5 @@ AWS.CognitoIdentityCredentials = AWS.util.inherit(AWS.Credentials, {
     }
   })()
 });
+
+exports.CognitoIdentityCredentials = AWS.CognitoIdentityCredentials;


chrisradek:
@villelahdenvuo 
You can accomplish this today by importing from `aws-sdk/global`. This entry point will not auto-load services

```typescript
import {CognitoIdentityCredentials} from 'aws-sdk/global';
```

villelahdenvuo:
@chrisradek Thanks for the quick reply. I couldn't find any examples for this.

PR #1797, opened 2017-11-12
Fix TypeScript definition for CredentialProviderChain constructor

mmorearty:
The constructor's `providers` argument should be marked as optional.

### DIFF
--- a/lib/credentials/credential_provider_chain.d.ts
+++ b/lib/credentials/credential_provider_chain.d.ts
@@ -4,7 +4,7 @@ export class CredentialProviderChain extends Credentials {
     /**
      * Creates a new CredentialProviderChain with a default set of providers specified by defaultProviders.
      */
-    constructor(providers: provider[])
+    constructor(providers?: provider[])
     /**
      * Resolves the provider chain by searching for the first set of credentials in providers.
      */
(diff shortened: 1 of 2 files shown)


mmorearty:
Hmm, I'm not sure how I could do that. For one thing, my change is just a change to the type declarations in the `.d.ts` file, which is useful for those devs who are writing TypeScript code that uses AWS; but your underlying actual code is written in JavaScript, and this commit did not change that code. Also, I can't find a file named `/ts/credentialprovidertest.ts` or, for that matter, any TypeScript tests at all. The closest thing I can find is `/test/credential_provider_chain.spec.js` (JavaScript) which tests the `CredentialProviderChain` class. That code does already have one test that calls `AWS.CredentialProviderChain()` with no arguments. Not that that really matters; since that code is JavaScript, it ignores the `.d.ts` file.

In other words, I don't think it's possible write a test for this.

jeskew:
Hi @mmorearty,

Sorry for the confusion; the file I mentioned is actually called [`/ts/credentialproviderchain.ts`](https://github.com/aws/aws-sdk-js/blob/master/ts/credentialproviderchain.ts). The SDK's tests for the included type definitions can be found in the repository's [`ts` folder](https://github.com/aws/aws-sdk-js/tree/master/ts), which contains TypeScript files using the SDK that are compiled (but not executed) as part of `npm test`.

To test this change, you would want to add an invocation of `CredentialProviderChain` without providing any arguments to `/ts/credentialproviderchain.ts`.

mmorearty:
Ah, ok — done. Let me know if that was not what you had in mind.

jeskew:
Thanks for your contribution!

PR #1799, opened 2017-11-14
Fix Readme link

dijonkitchen:
By removing unnecessary HTML p tags and making link the proper Markdown format.


dijonkitchen:
@jeskew so it's extending the Markdown syntax? Maybe whatever is being used should be compatible with markdown or change the file type?

jeskew:
@dijonkitchen You're right, the documentation tool should be updated to better support markdown or use its own file extension. Until that change lands, however, this PR would add a broken link to our API documentation.

dijonkitchen:
I'm closing this since it looks like the latest docs have already made my change. 

In the event that there are disconnects with the API documentation, just using a plain link will likely work for anything.

PR #1803, opened 2017-11-16
Reverts S3 presigned url default signatureVersion regression introduced in v2.150.0

### LINKED ISSUE #1802: S3 pre-signed URL broken after upgrade to 2.151.0
Hi,

After the upgrade from aws-sdk `2.148.0` to `2.151.0` I'm not able to upload anything to my S3 anymore. 

After digging a bit, it looks like that something is broken with the pre-signed URLs in this new version of the library. I double checked the release notes and didn't see anything about any modifications which would have to be performed in my code regarding this topic.

Without changing any pieces of codes of all servers and apps except upgrading to `2.151.0`, I now face a `403 - Forbidden access` due to a `SignatureDoesNotMatch`

    <Error><Code>SignatureDoesNotMatch</Code>
    <Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>

Rollbacking to `2.148.0` solve the issue.

Best regards

If needed, my server code to generate the pre-signed url which didn't changed since more than a year:

    var AWS = require('aws-sdk');
    
    AWS.config.update({accessKeyId: "myKeyId", secretAccessKey: "mySecret"});
    AWS.config.region = 'eu-west-1';
    
    var s3 = new AWS.S3({useDualstack: true});
     
    var imgName = req.body.imgName;
    var contentType = req.body.contentType;
     
    // Expires in seconds
    var params = {
        Bucket: 'mybucket',
        Key: imgName, Expires: 600,
        ContentType: contentType,
        ContentEncoding: 'base64',
        ACL: 'public-read'};
        s3.getSignedUrl('putObject', params, function (err, url) {
            if (err) {
               res.status(500).json({
                    error: "Signed S3 url for putObject can't be created. " + JSON.stringify(err)
               });
           } else {
              res.json({url: url});
          }
      });

chrisradek:
/cc @jeskew @AllanFly120 

Fixes #1802

### DIFF
--- a/.changes/next-release/bugfix-S3-c91e72e2.json
+++ b/.changes/next-release/bugfix-S3-c91e72e2.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "S3",
+  "description": "Reverts default signatureVersion for S3 presigned URLs to the lowest signatureVersion the region supports. 2.150.0 changed the default signatureVersion of all presigned URLs to v4. This change does not affect S3 clients that had the signatureVersion explicitly defined."
+}
--- a/lib/region_config_data.json
+++ b/lib/region_config_data.json
@@ -25,6 +25,14 @@
     "us-gov-*/sts": {
       "endpoint": "{service}.{region}.amazonaws.com"
     },
+    "us-gov-west-1/s3": "s3signature",
+    "us-west-1/s3": "s3signature",
+    "us-west-2/s3": "s3signature",
+    "eu-west-1/s3": "s3signature",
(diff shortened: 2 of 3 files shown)


billinghamj:
This introduced an (arguably) breaking change: changing S3 URLs from the hyphenated format `s3-eu-west-1.amazonaws.com` to `s3.eu-west-1.amazonaws.com`.

This broke a bunch of our systems due to CSP configuration.

Probably should document the breaking change in the changelog for anyone else who unfortunately upgrades and doesn't realise this has changed.

PR #1806, opened 2017-11-21
Rename ECSCredentials to RemoteCredentials and derive new ECSCredentials from it

AWSSteveHa:
Rename ECSCredentials to RemoteCredentials and create new ECSCredentials derived from RemoteCredentials. The new ECSCredentials renames the RemoteCredentials URI environment variables to be backwards compatible with the previous ECSCredentials.


lib/credentials/remote_credentials.d.ts
```diff
@@ -0,0 +1,33 @@
+import {Credentials} from '../credentials';
+export class RemoteCredentials extends Credentials {
+		/**
+		 * Represents credentials received.
+		 * @param {object} options - Override the default (1s) timeout period.
+		 */
+		constructor(options?: RemoteCredentialsOptions);
+	}
```
chrisradek: Spaces should be realigned.

lib/credentials/ecs_credentials.d.ts
```diff
@@ -1,17 +0,0 @@
-import {Credentials} from '../credentials';
```
jeskew: Can you restore this file? This change leaves the `ECSCredentials` class in place and preserves its interface, so we would still want to provide types for that class.

lib/credentials/remote_credentials.d.ts
```diff
@@ -0,0 +1,33 @@
...
+    interface RemoteCredentialsOptions {
+        httpOptions?: {
+            /**
+             * Timeout in milliseconds.
+             */
+            timeout?: number
+        }
+        maxRetries?: number
+    }
+export class ECSCredentials extends RemoteCredentials {
```
chrisradek: Can you keep the ECSCredentials in it's own module?

lib/credentials/remote_credentials.js
```diff
@@ -217,3 +234,15 @@ AWS.ECSCredentials = AWS.util.inherit(AWS.Credentials, {
     });
   }
 });
+
+AWS.ECSCredentials = AWS.util.inherit(AWS.RemoteCredentials, {
```
jeskew: `AWS.ECSCredentials` should be in `lib/credentials/ecs_credentials.js` (the file path previously used), and that file should require `AWS.RemoteCredentials`. Some customers may be directly importing ECS Credentials as a submodule, so its location on disk is a public interface.

lib/credentials/remote_credentials.d.ts
jeskew: Ah, I see that this was moved. Instead, it should remain in the same file as before in case someone is importing ECSCredentials directly as a submodule.

lib/credentials/remote_credentials.d.ts
```diff
@@ -0,0 +1,33 @@
+import {Credentials} from '../credentials';
+export class RemoteCredentials extends Credentials {
+		/**
```
jeskew: The indenting throughout this file should be normalized. Please use 4 spaces.

lib/node_loader.js
```diff
@@ -26,7 +26,7 @@ require('./http/node');
 
 // Load custom credential providers
 require('./credentials/ec2_metadata_credentials');
-require('./credentials/ecs_credentials');
```
jeskew: If you make the changes described above, you will still need to require `ecs_credentials`

AWSSteveHa:
I'm still getting errors on the last Travis test having to do with ecs_credentials.d.ts. I'm not sure what the import should look like. I've tried it a couple of ways now.

lib/credentials/ecs_credentials.d.ts
```diff
@@ -1,11 +1,11 @@
...
-export class ECSCredentials extends Credentials {
-		/**
-		 * Represents credentials received from relative URI specified in the ECS container.
-		 * @param {object} options - Override the default (1s) timeout period.
-		 */
-		constructor(options?: ECSCredentialsOptions);
-	}
+import {RemoteCredentials} from '../remote_credentials';
```
chrisradek: `remote_credentials` is located in the same directory, so it should be `./remote_credentials`

AWSSteveHa: Ah of course. I added the fix and all tests appear to be passing now. Thx

lib/credentials/remote_credentials.js
```diff
@@ -0,0 +1,248 @@
...
+        } catch (dataError) {
+          err = dataError;
+        }
+      }
+      callbacks(err, creds);
+    });
+  }
+});
+
+AWS.ECSCredentials = AWS.util.inherit(AWS.RemoteCredentials, {
```
jeskew: Isn't this a duplicate of what's in `lib/credentials/ecs_credentials.js`?

AWSSteveHa: Thanks for the catch. I should have removed that when I put it in ecs_credentials.js. Added fix to PR.

AWSSteveHa:
Please let me know if there are any additional changes I need to make. If possible it would be really great if this could be pulled by EOD tomorrow. Thank you.

PR #1808, opened 2017-11-22
Update presigner to only hoist headers in v2 presigned URLs

### LINKED ISSUE #1804: S3 Uploading pre-signed URL with signature v4, Content Disposition causes signature mismatch
I am currently using pre-signed uploading to directly upload from browser to s3. I can upload with no problems when I do not set the content disposition in options, for instance:

> var s3upload = new AWS.S3({
      signatureVersion: 'v4',
      accessKeyId: process.env.S3_UPLOAD_CLIENT_KEY,
      secretAccessKey: process.env.S3_UPLOAD_CLIENT_SECRET,
    });
  var baseOptions = {
      Bucket: process.env.S3_BUCKET,
      ACL: 'public-read',
      Key: uuidv4(),
    }

This works fine, http request:
works1

This also works fine when I set encryption

> var baseOptions = {
      Bucket: process.env.S3_BUCKET,
      ACL: 'public-read',
      Key: uuidv4(),
     ServerSideEncryption: 'AES256'
    }
works2

But if I add content disposition to options, I always get Signature mismatch
> var baseOptions = {
      Bucket: process.env.S3_BUCKET,
      ACL: 'public-read',
      Key: uuidv4(),
     ContentDisposition: 'attachment',
     ServerSideEncryption: 'AES256'
    }

notworking

I get the following response:
`<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>`

What is the problem and how can I fix it ?

jeskew:
Resolves #1804. All integration tests were run locally and pass.

### DIFF
--- a/lib/signers/presign.js
+++ b/lib/signers/presign.js
@@ -24,11 +24,7 @@ function signedUrlBuilder(request) {
         code: 'InvalidExpiryTime', message: message, retryable: false
       });
     }
-
-    var parsedUrl = AWS.util.urlParse(request.httpRequest.path);
-    var queryParams = parsedUrl.search
-      ? AWS.util.queryStringParse(parsedUrl.search.substr(1)) : {};
-    queryParams['Expires'] = expires;
+    request.httpRequest.headers[expiresHeader] = expires;
   } else if (signerClass === AWS.Signers.S3) {
     var now = request.service ? request.service.getSkewCorrectedDate() : AWS.util.date.getDate();
     request.httpRequest.headers[expiresHeader] = parseInt(
--- a/lib/signers/presign.js
+++ b/lib/signers/presign.js
@@ -24,7 +24,11 @@ function signedUrlBuilder(request) {
         code: 'InvalidExpiryTime', message: message, retryable: false
       });
     }
-    request.httpRequest.headers[expiresHeader] = expires;
+
+    var parsedUrl = AWS.util.urlParse(request.httpRequest.path);
+    var queryParams = parsedUrl.search
+      ? AWS.util.queryStringParse(parsedUrl.search.substr(1)) : {};
+    queryParams['Expires'] = expires;
   } else if (signerClass === AWS.Signers.S3) {
     var now = request.service ? request.service.getSkewCorrectedDate() : AWS.util.date.getDate();
     request.httpRequest.headers[expiresHeader] = parseInt(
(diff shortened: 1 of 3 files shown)


lib/signers/presign.js
```diff
@@ -24,7 +24,11 @@ function signedUrlBuilder(request) {
         code: 'InvalidExpiryTime', message: message, retryable: false
       });
     }
-    request.httpRequest.headers[expiresHeader] = expires;
+
+    var parsedUrl = AWS.util.urlParse(request.httpRequest.path);
+    var queryParams = parsedUrl.search
+      ? AWS.util.queryStringParse(parsedUrl.search.substr(1)) : {};
+    queryParams['Expires'] = expires;
```
chrisradek: It doesn't look like queryParams is used after it's created.

lib/signers/presign.js
```diff
@@ -49,31 +53,31 @@ function signedUrlSigner(request) {
...
-  var auth = queryParams['Authorization'].split(' ');
+  var auth = request.httpRequest.headers['Authorization'].split(' ');
   if (auth[0] === 'AWS') {
     auth = auth[1].split(':');
     queryParams['AWSAccessKeyId'] = auth[0];
     queryParams['Signature'] = auth[1];
+
+    AWS.util.each(request.httpRequest.headers, function (key, value) {
+      if (key === expiresHeader) key = 'Expires';
+      if (key.indexOf('x-amz-meta-') === 0) {
```
chrisradek: Do you know if there are any headers that we hoist to the query string today that don't also have to be sent as headers when using sigv4 presigned urls? Maybe that's just a feature of sigv2, but I'm wondering if we'll lose out on that feature with this change.

jeskew: There are, and they are all handled [here](https://github.com/aws/aws-sdk-js/blob/master/lib/signers/v4.js#L44). The current V4 presigner will always produce an invalid signature when any other headers are present.

chrisradek: Ah ok. So we may want to add additional headers to that section you pointed to in the future, but as of now, this change won't cause any existing cases to break. This part looks good then!

lib/signers/presign.js
jeskew: Will remove.

PR #1812, opened 2017-11-28
Add general purpose RemoteCredentials and derive ECSCredentials from it

AWSSteveHa:
Similar to previous PR but RemoteCredentials is using same environment variables as ECSCredentials.


jeskew:
Merged in as part of #1813

PR #1848, opened 2017-12-28
Adding constructor in HttpRequest typings

### LINKED ISSUE #1847: Wrong d.ts file for HttpRequest
[Here](https://github.com/aws/aws-sdk-js/blob/master/lib/http.js#L95) in constructor we have two arguments:

```ts
constructor: function HttpRequest(endpoint, region) {
    endpoint = new AWS.Endpoint(endpoint);
    ...
    this.region = region;
    ...
  },
...
```

but in [typings](https://github.com/aws/aws-sdk-js/blob/master/lib/http_request.d.ts) there is no constructor.

I suggest to add something like:

```ts
    /**
     * Constructs HttpRequest object with provided endpoint and region
     */
    constructor(endpoint: Endpoint, region: string);
```

slawiko:
This PR fixes #1847 issue: it adds constructor in HttpRequest typings

### DIFF
--- a/lib/http_request.d.ts
+++ b/lib/http_request.d.ts
@@ -4,6 +4,10 @@ import {Endpoint} from './endpoint';
  */
 export class HttpRequest {
     /**
+     * Constructs HttpRequest object with provided endpoint and region
+     */
+    constructor(endpoint: Endpoint, region: string);
+    /**
      * The part of the path excluding the query string.
      */
     pathname(): string;


jeskew:
I would not be inclined to include the constructor in our typings for HttpRequest. If for whatever reason the SDK needs to support an environment that requires more instantiation data for requests, I would want that implementation to be compatible with the existing HttpRequest interface.

TypeScript [allows constructor definitions to be defined separately](https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes), and I wouldn't have any objection to including a separate `HttpRequestConstructor` interface. Would that address the issue that spurred #1847?

slawiko:
@jeskew so you propose to replace `HttpRequest` class with `HttpRequest` interface and `HttpRequestConstructor`? In this case I don't understand for what? Which problem does it solve?

But anyway yes, it would fix #1847.

jeskew:
@slawiko That's a good point; I thought we exposed an interface rather than a class definition. Thanks for your contribution!

PR #1857, opened 2018-01-05
Only pull in the small subset of crypto-browserify that is used by the SDK

### LINKED ISSUE #1602: Vulnerability in crypto-browserify
AWS SDK Version 2.77.0

The AWS NodeJS SDK uses crypto-browserify v1.0.9 (current version is 3.11.0) which has two known vulnerabilities.

screen shot 2017-06-27 at 12 33 55

This report was generated using owdit (https://github.com/KurtPattyn/owdit) which fetches vulnerabilities from OSSIndex (https://ossindex.net/start/npm/).

jeskew:
This PR replaces the SDK's dependency on `crypto-browserify` with direct dependencies on the portions of the crypto API on which the SDK depends, namely `createHash` and `createHmac`. As a result, we are no longer pulling in any form of CSPRNG or AES implementation, which should reduce the size of the SDK's bundle somewhat.

Resolves #1602


KurtPattyn:
Great!

WChoy:
Breaking the crypto in ReactNative and Browser.   "Unable to resolve module 'stream'."

create-hmac and create-hash use a hash-base module. That does a
"var Transform = require('stream').Transform". That was defined in node.js core (NOT a standard JS function, see https://nodejs.org/api/stream.html#stream_new_stream_transform_options)

Reproducible in https://github.com/awslabs/aws-mobile-react-native-starter

PR #1859, opened 2018-01-06
Region check

chrisradek:
Adds a check to make sure we aren't hard-coding regions in our code-base. There is a whitelist in for cases where is is necessary. Would like to eventually move to using an AST...

Also, will squash commits.


chrisradek:
Close in favor of #1860

PR #1863, opened 2018-01-09
chore: switch the coverage tool to nyc

kt3k:
Istanbul's cli is moved to [nyc](https://github.com/istanbuljs/nyc), and istanbul module itself is deprecated in favor of it. (See [readme](https://github.com/gotwarlost/istanbul/blob/master/README.md) of istanbul repository.)

This PR switches the cli tool to nyc.


jeskew:
Hi @kt3k,

I understand that new development is occurring on the `nyc` project, but `istanbul` has been working fine with the SDK for some time. I'm not sure I see enough value in upgrading this build-time only dependency to justify auditing the differences between [`nyc`'s dependency graph](http://npm.anvaka.com/#/view/2d/nyc) and [that of `istanbul`](http://npm.anvaka.com/#/view/2d/istanbul).

kt3k:
Hello @jeskew,

Thanks for the feedback.

A major difference between istanbul (v0.4.x) and nyc is the support of es6 (and above). Old istanbul does not instrument the code correctly when we use es6. Perhaps that doesn't affects this project immediately, but in future when we dropped the support of <= 0.12.x the use of istanbul will block us using es6.

> I'm not sure I see enough value in upgrading this build-time only dependency to justify auditing the differences

How do you audit the dependencies? Isn't the scanning tool like [nsp](https://www.npmjs.com/package/nsp) enough for this project?

jeskew:
The main thing I would need to audit is the license on each package. I appreciate the effort you've put into this, but for backwards compatibility, this library can never use ES6-only language features. I'll leave this open in case another reason to migrate comes up, but we're not likely to make this change in the current major version of the SDK.

Y-LyN-10:
@jeskew, I understand the attitude of "why changing _x_ if it works", but with the time it really get's harder to maintain the codebase with outdated dependencies. 

Currently "npm audit" on the master branch reports 18 vulnerabilities (some of them critical) and the only fix for them is to make an update with breaking-changes. Also, there are multiple deprecation warnings from node, running the latest LTS version. Sure, these vulnerabilities are only for the **dev dependencies**, which doesn't matter for the package users, but I think that we shouldn't hesitate some changes :) 

Perhaps, some of the open PRs already fixed that, I don't know, but I support the idea of switching to **nyc**

kt3k:
Closing for now. Please mention me if this change is necessary in future.

PR #1864, opened 2018-01-09
Stop testing on unsupported node versions

kt3k:
node v5 and v7 (and any other odd numbered versions) are basically for node.js contributors, not for the users in general and they are expired now. So I dropped v5 and v7 from travis settings. Node v4 is a supported version, but the older minor versions are not supported, so dropped v4.2 and v4.4 and added v4 instead. (The current LTS version in 4.x line is v4.8.7).

See https://github.com/nodejs/Release for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,9 @@
 language: node_js
 node_js:
-#  - "0.8"
   - "0.10"
   - "0.12"
-  - "4.2"
-  - "4.4"
-  - "5"
+  - "4"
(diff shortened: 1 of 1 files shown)


jimmywarting:
I think you should stop support old platforms and remove old legacy code.
v6 will drop soon in april so maybe just support node >= 8 for the next major version change.

srchase:
We're electing to continue testing against these versions.  Thanks for the contribution.

PR #1880, opened 2018-01-16
Replace dependency on create-hash and create-hmac with bundled hash implementations

jeskew:
This PR is a follow-up to #1857. `create-hash` and `create-hmac` bring in a transitive dependency that some customers have reported as causing build problems (e.g., https://github.com/aws/aws-sdk-js/pull/1857#issuecomment-357347958). This passes the browser and react-native tests locally using karma + phantomjs, but it's worth trying in a few different browsers.


jeskew:
The integration tests run without issue in IE, Edge, Chrome, and Firefox on my computer, though there is an unrelated failure running the unit tests in IE (same error both with master and this PR).

PR #1905, opened 2018-02-01
Allow chained profiles from config (fixes #1891)

### LINKED ISSUE #1891: Recursive config profiles don't work.
I can set up my `~/.aws/config` like
```ini
; matching aws credentials in ~/.aws/credentials
[profile base]
region = ap-southeast-2
output = text

[profile first]
role_arn = arn:aws:iam::00000:role/first-role
source_profile = base

[profile second]
role_arn = arn:aws:iam::00000:role/second-role
source_profile = first
```

then set `AWS_PROFILE=second`. Botocore, and hence the AWS CLI, will then use base to assumeRole into first, which is then used in turn to assumeRole into second. This chain can be as long as you like, but I've kept it short in this issue.

This should be supported in the JS SDK as it is in botocre, but currently this doesn't work. Any chance of allowing this?

akdor1154:
The Python SDK allows chained profiles, e.g. 
```ini
; matching aws credentials in ~/.aws/credentials
[profile base]
region = ap-southeast-2
output = text

[profile first]
role_arn = arn:aws:iam::00000:role/first-role
source_profile = base

[profile second]
role_arn = arn:aws:iam::00000:role/second-role
source_profile = first
```
(See #1891)
This PR adds this functionality to the JS SDK as well. Implementation is nice and simple - instead of directly constructing a `Credentials` with the access and secret key of the `source_profile` of a given profile, `SharedIniFileCredentials` just creates a parent `SharedIniFileCredentials` referring to the source profile, and uses this for its `STS` call.
Net code (not including tests) actually decreases as a result of this change.

Couple of questions/notes just related to testing:
 - spying on the SharedIniFileCredentials constructor caused some weird issues (such as breaking the constructor), are these known about?
 - Because of the above, I spied on STS.assumeRole instead. However Is there a better way of getting to this method? `STS.prototype.assumeRole` does not seem to work, hence the roundabout `(new STS()).constructor.prototype.assumeRole` .

Cheers
Jarrad

### DIFF
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -47,6 +47,8 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
    * @option options disableAssumeRole [Boolean] (false) True to disable
    *   support for profiles that assume an IAM role. If true, and an assume
    *   role profile is selected, an error is raised.
+   * @option options preferStaticCredentials [Boolean] (false) True to
+   *   prefer static credentials to role_arn if both are present.
    */
   constructor: function SharedIniFileCredentials(options) {
     AWS.Credentials.call(this);
@@ -56,6 +58,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
     this.filename = options.filename;
     this.profile = options.profile || process.env.AWS_PROFILE || AWS.util.defaultProfile;
     this.disableAssumeRole = Boolean(options.disableAssumeRole);
+    this.preferStaticCredentials = Boolean(options.preferStaticCredentials);
     this.get(function() {});
   },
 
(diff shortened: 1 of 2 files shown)


akdor1154:
Hi @jeskew, sorry to nag but is there someone I can ping to review this?

jeskew:
I am indeed the person to ping to review this.

There's a subtle incompatibility with the CLI, which is that the presence of both a `role_arn` and static credentials have different meanings depending on how many profiles have been visited. For the first profile processed, `role_arn` takes precedence over any static credentials, but for all subsequent profiles, static credentials are used if present, and only in their absence will the profile's `source_profile` and `role_arn` keys be used to load another set of credentials.

akdor1154:
Awesome, thanks for looking into this.  By static you mean keys saved into the credentials file? (Not env)

jeskew:
I did some manual testing to verify that everything is hunky dory. Thanks for your work on this, @akdor1154!

PR #1923, opened 2018-02-09
Implement usage of credential_process from AWS ini credentials file if it exists

dheffx:
Implements a function that takes the credential_process value from the AWS credentials ini config, executes it, and uses its response JSON as the credentials.

Makes the following assumptions:
1) The value of the credential_process is the path to an executable
2) Assumes the executable's response data conforms with the specifications found here: 
https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#sourcing-credentials-from-external-processes


dheffx:
I am having trouble understanding why my request failed the second time around. The step "npm run coverage" passes for me on repeated attempts, and passed during my first pull, but now fails after I only made fixes for eslint.

I poked around on other user's pull request failures and saw a few that had what appeared to be the same failure. I am hoping that means its an issue with the CI or it is easy to identify the problem.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -105,9 +106,16 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
         return;
       }
 
-      this.accessKeyId = profile['aws_access_key_id'];
-      this.secretAccessKey = profile['aws_secret_access_key'];
-      this.sessionToken = profile['aws_session_token'];
+      if (profile['credential_process']) {
```
jeskew: To match [current AWS CLI behavior](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L86), static credentials should take higher precedence than a `credential_process`.

dheffx:
Thanks for looking at my request and giving this feedback! I've made changes and added tests based on your response. I believe the patch should address both cases.

Also, a side question, is it intentional that CredentialProviderChain in node_loader simply checks if an error wasn't thrown, and moves on? It threw me off for a bit, since it effectively hid the error messages thrown by SharedIniFileCredentials.

`.resolve(function(err, creds) {
      if (!err) credentials = creds;
    });`
https://github.com/dheffx/aws-sdk-js/blob/501a7fe0e1286ac4f8e74d4e395c68e822438bef/lib/node_loader.js#L66

jeskew:
It is intentional, but the original error should be available as the `originalError` property of the final error thrown.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -105,9 +109,14 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
         return;
       }
 
-      this.accessKeyId = profile['aws_access_key_id'];
-      this.secretAccessKey = profile['aws_secret_access_key'];
-      this.sessionToken = profile['aws_session_token'];
+      if (this.useCredentialProcess) {
```
jeskew: Since `useCredentialProcess` is set to `true` in the default chain, this is still incompatible with the CLI. Given a profile with like the following:
```
[foo]
credential_process = /path/to/bin --flag
aws_access_key_id = asdf
aws_secret_access_key = hjkl
```

the SDK for JS would execute `/path/to/bin --flag`, whereas the AWS CLI would not.

Rather than adding a constructor flag, this could be made compatible by changing the order of the branches in the if statement:

```javascript
if (profile['aws_access_key_id'] || profile['aws_secret_access_key']) {
    // use the profile as static creds
} else if (profile['credential_process']) {
    // execute the credential process
}
```

dheffx: I am not sure that is true given that I added an additional instance to the defaultProvider list, so the instance without passing useCredentialProcess will execute first, and if the credentials do not exist in the ini then that instance will error out, and go to the instance in the chain with it set to true.

The example you provided for should be handled in one of the test cases I added to the credential_provider_chain.spec.js

`AWS.CredentialProviderChain.defaultProviders = [
  function () { return new AWS.EnvironmentCredentials('AMAZON'); },
  function () { return new AWS.SharedIniFileCredentials(); },
  function () { return new AWS.SharedIniFileCredentials({ useCredentialProcess: true }); },
  function () {
    if (AWS.ECSCredentials.prototype.isConfiguredForEcsCredentials()) {
      return new AWS.ECSCredentials();
    }
    return new AWS.EC2MetadataCredentials();
  }
];`

That said, I will test it out tonight and make the adjustments you mentioned if it is necessary.

jeskew: Ah my mistake; I misread the change to the default chain and missed that it was an addition rather than an update.

In any case, just changing the order of the `if/else` block seems simpler and has fewer moving parts.

dheffx: Fair enough. 
I had tried to mimic the example you had cited from boto where they " SharedCredentialProvider" followed by "ProcessProvider" in the providers list. I thought that's what you were originally asking for, but if preferred I can switch it to the simpler implementation you mentioned.

dheffx:
I pushed the changes you requested, and moved the tests to credentials.spec.js from credentials_provider_chain.spec.js since provider chain ended up not changing.

The CI failed but it appears to be a common timeout issue and seemingly unrelated to my change, unless I am mistaken.

dheffx:
@jeskew Hey, just curious if you had a chance to look at this again, and if it looked all good? Thanks.

lorengordon:
Any updates on supporting this feature? Would be very handy.

thebenwaters:
Python, Ruby, and Java support this, why can't javascript (since a high percentage of users are using javascript)...

srchase:
@dheffx 

Working to get this functionality pulled in with https://github.com/aws/aws-sdk-js/pull/2559.

Appreciate your contribution!  Apologize we didn't close the loop sooner on getting this pulled in.

PR #1951, opened 2018-02-27
Correct sns.d.ts message type to allow json as well as string

chmontgomery:
Based on the documentation: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SNS.html#publish-property


chmontgomery:
sorry, just realized this needs to be a stringified json object, not a real object 😢

PR #1987, opened 2018-03-26
Fix typo in s3 error property extendedRequestId

cjlarose:


chrisradek:
@cjlarose 
Thanks for the contribution! Merging.

PR #2079, opened 2018-06-07
Adds types for eventstream shapes

chrisradek:
/cc @AllanFly120


chrisradek:
I actually do want to include the model changes with this PR. Removed them from #2074

PR #2126, opened 2018-07-04
Support shared ini file mfa_serial with fn to fetch token code

### LINKED ISSUE #1543: SDK doesn't work with roles requiring MFA
Used SDK version: 2.60.0
Node.js version: v6.9.5

We have a role like one below defined in our account. We're working on to get MFA in use with both aws cli and with the node.js SDK.
```
  AccountAdministratorRole:
     Type: "AWS::IAM::Role"
     Properties:
       ManagedPolicyArns:
        - arn:aws:iam::aws:policy/AdministratorAccess
       AssumeRolePolicyDocument:
         Version: "2012-10-17"
         Statement:
           -
             Effect: "Allow"
             Principal:
              AWS: !Sub arn:aws:iam::${AWS::AccountId}:root
             Action:
               - "sts:AssumeRole"
             Condition:
              Bool:
                aws:MultiFactorAuthPresent: true
       Path: "/"
```

And the ~/.aws/config:
```
[profile account]
region = eu-central-1
[profile account-admin]
role_arn = arn:aws:iam::<accountnumber>:role/AccountAdministratorRole
region = eu-central-1
source_profile = account
mfa_serial = arn:aws:iam::<accountnumber>:mfa/ville.hartikainen
```

When using aws-cli, it prompts properly for the MFA code and caches the temporary credentials under ~/.aws/cli/cache/. The node.js SDK is not sharing this functionality at all and seems like the support for MFA is completely missing.

This issue was discussed already in #993, and @andreineculau presented a potential (though ugly) workaround for it. The original issue was closed in #1391, but the problem presented with MFA tokens usage still remained. For the ease of readers, I will just quickly post the slightly edited version of it below, which seem to work with latest tagged version of SDK:
```
const ini = require('ini');
const fs=require('fs');
fs.readFileSync(
      `${process.env.HOME}/.aws/config`,
      'utf-8'
    )
let awsProfile = process.env.AWS_PROFILE || process.env.AWS_DEFAULT_PROFILE;
if (awsProfile) {
  try {
    let configIni = ini.parse(fs.readFileSync(
      `${process.env.HOME}/.aws/config`,
      'utf-8'
    ));
    let awsProfileConfig = configIni[`profile ${awsProfile}`];
    if (awsProfileConfig && awsProfileConfig.role_arn) {
      let roleArn = awsProfileConfig.role_arn.replace(/:/g, '_').replace(/[^A-Za-z0-9\-_]/g, '-');
      let awsCliCacheFilename = `${awsProfile}--${roleArn}`;
      let awsCliCache =
          JSON.parse(fs.readFileSync(
            `${process.env.HOME}/.aws/cli/cache/${awsCliCacheFilename}.json`,
            'utf-8'
          ));
      let sts = new AWS.STS();
      AWS.config.credentials = sts.credentialsFrom(awsCliCache,awsCliCache);
    }
  } catch (_err) {
    console.log(_err)
  }
}
```

The workaround is utilizing the temporary credentials cached by aws cli which seems like an adequeate option for us, but I'd be happy to see some support for the MFA built in the SDK itself

jpb:
This change supports use of `mfa_serial` in the `SharedIniFileCredentials` provider. 

`tokenCodeFn` will be called if `mfa_serial` is found in the active profile of the shared ini file. This function will be called with the value of `mfa_serial` and should invoke the callback with the MFA token or an error. This token will be provided to the STS assume role call.

This change also prevents the `SharedIniFileCredentials` from refreshing multiple times concurrently (and thus having `tokenCodeFn` be called more than one time concurrently), similar to how the [MetadataService does it](https://github.com/aws/aws-sdk-js/blob/master/lib/metadata_service.js#L99-L100).

Example usage of this change, to prompt the user for a MFA token in a CLI application:

```
const inquirer = require('inquirer');
const aws = require('./index.js');

const f = (serial, cb) => {
  const resp = inquirer.prompt(
    {
      name: 'token',
      type: 'input',
      default: '',
      message: `MFA token for ${serial}:`
    }).then((r) => {
      cb(null, r.token);
    }).catch((e) => {
      console.log('error:', e);
      cb(e);
    });
};

const creds = new aws.SharedIniFileCredentials({
  tokenCodeFn: f,
  profile: 'profile-requiring-mfa'
});

(async () => {
  await creds.getPromise().then(function(e) {
    console.log(creds);
  }).catch(function(e) {
    console.log(creds, 'error:', e);
  });
})();
```

Using this with the credentials provider chain could be done by configuring `SharedIniFileCredentials` in the example [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html#defaultProviders-property).

Fixes #1543

### DIFF
--- a/lib/credentials/shared_ini_file_credentials.d.ts
+++ b/lib/credentials/shared_ini_file_credentials.d.ts
@@ -10,5 +10,5 @@ interface SharedIniFileCredentialsOptions {
     profile?: string
     filename?: string
     disableAssumeRole?: boolean
-    tokenCodeFn?: (mfaSerial: string, callback: (err?: string, token?: string) => void) => void
+    tokenCodeFn?: (mfaSerial: string, callback: (err?: Error, token?: string) => void) => void
 }
--- a/lib/credentials/shared_ini_file_credentials.d.ts
+++ b/lib/credentials/shared_ini_file_credentials.d.ts
@@ -10,5 +10,5 @@ interface SharedIniFileCredentialsOptions {
     profile?: string
     filename?: string
     disableAssumeRole?: boolean
-    tokenCodeFn?: (mfaSerial: string, callback: (err: string|null|undefined, token: string) => void) => void
+    tokenCodeFn?: (mfaSerial: string, callback: (err?: string, token?: string) => void) => void
 }
--- a/lib/credentials/shared_ini_file_credentials.d.ts
+++ b/lib/credentials/shared_ini_file_credentials.d.ts
@@ -10,4 +10,5 @@ interface SharedIniFileCredentialsOptions {
     profile?: string
     filename?: string
     disableAssumeRole?: boolean
+    tokenCodeFn?: (mfaSerial: string, callback: (err: string|null|undefined, token: string) => void) => void
 }
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -225,9 +225,15 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       roleParams.SerialNumber = mfaSerial;
       self.tokenCodeFn(mfaSerial, function(err, token) {
         if (err) {
+          var message;
+          if (err instanceof Error) {
+            message = err.message;
+          } else {
+            message = err;
+          }
           self.callRefreshCallbacks(
             AWS.util.error(
-              new Error('Error fetching MFA token: ' + err),
+              new Error('Error fetching MFA token: ' + message),
               { code: 'SharedIniFileCredentialsProviderFailure' }
             ));
           return;
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -49,7 +49,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
    *   role profile is selected, an error is raised.
    * @option options preferStaticCredentials [Boolean] (false) True to
    *   prefer static credentials to role_arn if both are present.
-   * @option options tokenCodeFn [Function] (noop function) Function to provide
+   * @option options tokenCodeFn [Function] (null) Function to provide
    *   STS Assume Role TokenCode, if mfa_serial is provided for profile in ini
    *   file. Function is called with value of mfa_serial and callback, and
    *   should provide the TokenCode or an error to the callback in the format
@@ -65,9 +65,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
     this.profile = options.profile || process.env.AWS_PROFILE || AWS.util.defaultProfile;
     this.disableAssumeRole = Boolean(options.disableAssumeRole);
     this.preferStaticCredentials = Boolean(options.preferStaticCredentials);
-    this.tokenCodeFn = options.tokenCodeFn || function(mfaSerial, callback) {
-      callback();
-    };
+    this.tokenCodeFn = options.tokenCodeFn || null;
     this.get(function() {});
   },
 
@@ -221,7 +219,7 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       roleParams.ExternalId = externalId;
     }
 
-    if (mfaSerial) {
+    if (mfaSerial && self.tokenCodeFn) {
       roleParams.SerialNumber = mfaSerial;
       self.tokenCodeFn(mfaSerial, function(err, token) {
         if (err) {
--- a/lib/credentials/shared_ini_file_credentials.js
+++ b/lib/credentials/shared_ini_file_credentials.js
@@ -49,17 +49,36 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
    *   role profile is selected, an error is raised.
    * @option options preferStaticCredentials [Boolean] (false) True to
    *   prefer static credentials to role_arn if both are present.
+   * @option options tokenCodeFn [Function] (noop function) Function to provide
+   *   STS Assume Role TokenCode, if mfa_serial is provided for profile in ini
+   *   file. Function is called with value of mfa_serial and callback, and
+   *   should provide the TokenCode or an error to the callback in the format
+   *   callback(err, token)
    */
   constructor: function SharedIniFileCredentials(options) {
     AWS.Credentials.call(this);
 
     options = options || {};
 
+    this.refreshCallbacks = [];
     this.filename = options.filename;
     this.profile = options.profile || process.env.AWS_PROFILE || AWS.util.defaultProfile;
     this.disableAssumeRole = Boolean(options.disableAssumeRole);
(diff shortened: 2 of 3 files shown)


jpb:
I'm not sure that the related changes have caused the test failure, but rather surfaced a broken test:

https://github.com/aws/aws-sdk-js/blob/a6b3ff2cac81edc6fb479b37cecfa619ac273c68/test/credentials.spec.js#L450-L459

I can't find where the error message `Credentials not set in source_profile ... using profile ...` would be generated.

jpb:
Thanks @chrisradek! 

I dug a bit more while I was adding some additional tests and discovered that the error message generated (on master and this branch) didn't match what the test was expecting. It appears that the exception handling in `SharedIniFileCredentials` catches the exception thrown by `assert`, which masked the error. I'm still not 100% sure why the test behaves differently with these changes, but I have fixed the error message.

lib/credentials/shared_ini_file_credentials.d.ts
```diff
@@ -10,4 +10,5 @@ interface SharedIniFileCredentialsOptions {
     profile?: string
     filename?: string
     disableAssumeRole?: boolean
-}
\ No newline at end of file
+    tokenCodeFn?: (mfaSerial: string, callback: (err?: string, token?: string) => void) => void
```
chrisradek: Can you make the `err` in your callback an `Error`? I see that you're treating this as sort of a 'reason', but I'd like to have this be an actual error so it's consistent with the rest of our callback APIs (and general node pattern of passing an error object as the first parameter to a callback).

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -49,19 +49,38 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
...
     AWS.Credentials.call(this);
 
     options = options || {};
 
+    this.refreshCallbacks = [];
     this.filename = options.filename;
     this.profile = options.profile || process.env.AWS_PROFILE || AWS.util.defaultProfile;
     this.disableAssumeRole = Boolean(options.disableAssumeRole);
     this.preferStaticCredentials = Boolean(options.preferStaticCredentials);
+    this.tokenCodeFn = options.tokenCodeFn || function(mfaSerial, callback) {
```
chrisradek: Instead of defining a default function if one isn't provided, can you update the `if (mfaSerial)` check in `loadRoleProfile` to also check for the existence of `this.tokenCodeFn`? That will save us from having to create an extra no-op function.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -49,19 +49,38 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
...
+      callback();
+    };
     this.get(function() {});
   },
 
+  /**
+   * @api private
+   */
+  callRefreshCallbacks: function (err) {
```
chrisradek: Cool update!

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -199,17 +221,44 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
       roleParams.ExternalId = externalId;
     }
 
+    if (mfaSerial) {
```
chrisradek: Comment made above to also check for existence of the `this.tokenCodeFn`.

lib/credentials/shared_ini_file_credentials.js
```diff
@@ -199,17 +221,44 @@ AWS.SharedIniFileCredentials = AWS.util.inherit(AWS.Credentials, {
...
+      roleParams.SerialNumber = mfaSerial;
+      self.tokenCodeFn(mfaSerial, function(err, token) {
+        if (err) {
+          self.callRefreshCallbacks(
+            AWS.util.error(
+              new Error('Error fetching MFA token: ' + err),
```
chrisradek: Assuming `err` is an Error object, I think it'd be fair to update this to use `err.message` instead of `err`.

test/credentials.spec.js
```diff
@@ -580,6 +580,56 @@
...
+        tokenCodeFn = function(serial, callback) {
+          tokenCodeFnSpy(serial);
+          callback(null, '123456');
+          done();
+        };
+        creds = new AWS.SharedIniFileCredentials({
+          profile: 'withmfa',
+          tokenCodeFn: tokenCodeFn
+        });
+        expect(tokenCodeFnSpy.calls.length).to.equal(1);
```
chrisradek: Can you move these expect statements to inside of your `tokenCodeFn`, before `done()`? Right now `done()` is getting called before these expect statements do, so this test will never report a failure.

test/credentials.spec.js
```diff
@@ -580,6 +580,56 @@
...
+        return expect(tokenCodeFnSpy.calls[0]["arguments"][0]).to.equal('serial');
+      });
+      it('callback receives tokenCodeFns error', function(done) {
+        var creds, tokenCodeFn, mock;
+        helpers.mockHttpResponse(200, {}, '<AssumeRoleResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">\n  <AssumeRoleResult>\n    <Credentials>\n      <AccessKeyId>KEY</AccessKeyId>\n      <SecretAccessKey>SECRET</SecretAccessKey>\n      <SessionToken>TOKEN</SessionToken>\n      <Expiration>1970-01-01T00:00:00.000Z</Expiration>\n    </Credentials>\n  </AssumeRoleResult>\n</AssumeRoleResponse>');
+        mock = '[default]\nrole_arn = arn\nmfa_serial = serial\naws_access_key_id = key\naws_secret_access_key = secret\n'
+          + '[profile withmfa]\nrole_arn = arn\nmfa_serial = serial\nsource_profile = default';
+        helpers.spyOn(AWS.util, 'readFileSync').andReturn(mock);
+        tokenCodeFn = function(serial, callback) {
+          callback('tokenCodeFn error');
```
chrisradek: Can you update this to return a new Error?

jpb:
@chrisradek I believe I've addressed your comments (including adding the additional test case).

chrisradek:
Merging! Thank you for the contribution, and good job with your first PR to the SDK!

someswarpanda:
Hi. Is this delivered and included in aws-sdk@2.437.0? I use `role_arn`, `mfa_serial` and `source_profile` in ~/.aws/config for the child profile, but the js sdk fails with _CredentialsError: Missing credentials in config_ 

I have, 
```
AWS_SDK_LOAD_CONFIG: '1'
AWS_CONFIG_FILE: '$HOME/.aws/config'
AWS_SHARED_CREDENTIALS_FILE: '$HOME/aws/credentials'
```
My guess is that the sdk should prompt for the token while assuming the role_arn or get it the temp creds from aws cache. But that does not seem to happen. Please help!

jpb:
@someswarpanda you must provide your own prompt – see `tokenCodeFn` https://github.com/aws/aws-sdk-js/pull/2126#issue-199310808. `tokenCodeFn` will be called when the SDK needs a new MFA token, and it is up to you to do what is appropriate for your application to get a new token.

PR #2183, opened 2018-08-08
Type FunctionError in Lambda InvocationResponse

LinusU:
Change the type from a generic string into an enum of the only two possible values.

### DIFF
--- a/.changes/next-release/bugfix-lambda-efd79fb0.json
+++ b/.changes/next-release/bugfix-lambda-efd79fb0.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "lambda",
+  "description": "Improve typing of FunctionError in InvocationResponse"
+}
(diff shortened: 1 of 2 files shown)


LinusU:
ping @chrisradek, any chance of getting this in? ☺️

srchase:
This suggested change is on code generated from the resource model, which is managed by the Service Team. Those models cannot be changed from this repository. Thank you very much for this contribution, but we will need to close out this request.

srchase:
@LinusU 

I've opened an internal request for the Lambda service team to review and consider updating their resource model to make this an enum.  

You're welcome to open your own request on their [AWS Developer Forum](https://forums.aws.amazon.com/forum.jspa?forumID=186).  If you link to that posting, I can update the internal request.

srchase:
@LinusU 

This would be a breaking change and is not something that Lambda will be able to change on their API model.   For consistency between the SDKs and how those models are shared, we are not able to get make this an enum.

Appreciate you working with us on this and apologize that we didn't get it resolved sooner.

Thanks!

LinusU:
I see, thanks for running it by the team ❤️

PR #2207, opened 2018-08-16
Updated EC2 API for TagSpecifications to make the parameter names cor…

aryajur:
…rect: resourceType->ResourceType, key-> Key and value->Value


srchase:
This suggested change is on the resource model, which is managed by the Service Team. Those models cannot be changed from this repository.  Thank you very much for this contribution, but we will need to close out this request.

Issue #2308, opened 2018-10-17
CloudFront signer issue dealing with single quote in content-disposition


lsholubka:
There seems to be an issue in CloudFront signer when passing content-disposition header with single quotes in it.

If I sign url with `'` not encoded (encodeURIComponent doesn't encode `'`):

``` `?response-content-disposition=${encodeURIComponent(`attachment; filename="${fileName}"`)}` ```

Then latest Chrome, Safari and Firefox get access denied, because those automatically encode `'` as `%27` in all requests. Edge/IE works ok.

If I sign url with `'` encoded:

``` `?response-content-disposition=${encodeURIComponent(`attachment; filename="${fileName}"`).replace(/'/g, '%27')}` ```

Then it works in Chrome, Safari and Firefox, but the downloaded file name will contain `%27` instead of `'`, because seems like CloudFront signer is not decoding it when signing.

S3 signer works as expected.

**Comment by srchase** [comment]
@lsholubka,

Can you share the code snippet of how you are passing in that content-disposition header?

**Comment by lsholubka** [comment]
@srchase 

```
    const url = `${configs.downloads.url}/${s3Path}/${file}` +
      `?response-content-disposition=${encodeURIComponent(`attachment; filename="${fileName}"`)}`;
    const signer = new AWS.CloudFront.Signer(
      config.key_pair_id,
      config.private_key,
    );
    const expires = 3000;
    return signer.getSignedUrl({
      url,
      expires,
    });
```

**Comment by srchase** [comment]
Thanks for responding with this.  How are you using content-disposition with the S3 Signer?

**Comment by lsholubka** [comment]
```
    const downloadIframe = $('<iframe>', { src: signResponse.signedRequest });
    downloadIframe.hide().appendTo('body');
```

**Comment by lsholubka** [comment]
sorry, that was how I use it in the browser. Here is s3 signer:

```
    const s3Params = {
      Bucket: bucket,
      Key: `${s3Path}/${file}`,
      Expires: 30000,
      ResponseContentDisposition: `attachment; filename="${fileName}"`,
    };
    const s3 = AwsHelpers.getS3(configs);
    return s3.getSignedUrl('getObject', s3Params);
```

srchase:
@lsholubka 

It seems there are two issues we need to resolve:

1.  The signer should handle encoding a plain url, which it does, but results in an invalid signature when the url includes response-content-disposition:

Passing in this url (double quotes):
[...]

PR #2314, opened 2018-10-18
improved documentation for #2308

srchase:
note that query params must be encoded.


lib/cloudfront/signer.js
```diff
@@ -157,7 +157,8 @@ AWS.CloudFront.Signer = inherit({
      *
      * @param options [Object]          The options to create a signed URL.
      * @option options url [String]     The URL to which the signature will grant
-     *                                  access. Required.
+     *                                  access. Any query params included with
```
chrisradek:
```suggestion
     *                                  access. Any query params included with
```
Looks like a trailing space is causing the linter to complain.

PR #2325, opened 2018-10-25
Cleans up connectTimeout timers when request errors

### LINKED ISSUE #2322: node: timers leaked on HTTP connection errors
If the `connectTimeout` http option has been supplied, timers are used to manage the connection timeout. The timers are cleared when a connection is established, however, the timers are *not* cleared if an error happens. 

The end result is that timers are leaked for every attempt to make an HTTP connection that does not succeed.

https://github.com/aws/aws-sdk-js/blob/master/lib/http/node.js#L60

## Reproduction

1. Create the following application:
    ```

    const dbOptions = {
      region: 'us-east-1',
      accessKeyId: 'accessKeyId',
      secretAccessKey: 'secretAccessKey',
      endpoint: new AWS.Endpoint('http://localhost:12345'),
      httpOptions: {
        connectTimeout: 10000
      }
    }
    const dynamo = new AWS.DynamoDB(dbOptions)

    console.log('about to start query')

    dynamo.listTables({}).promise()
      .then(() => {
        console.log('query complete')

        console.time('waiting for leaked timers')

        process.on('exit', () => {
          console.timeEnd('waiting for leaked timers')
        })
      })
      .catch((err) => {
        console.log('error executing query', err)
      })
    ```
2. Run the node application
3. A couple seconds after starting the application, fire up local dynamo:

    `docker run -it --rm -p 12345:8000 --name dynamo amazon/dynamodb-local`

### Expected

`waiting for leaked timers: ` reports < 1 millisecond.

### Unexpected

Application hangs for multiple seconds after seeing "query complete".

`waiting for leaked timers: ` reports multiple seconds (approximately `10 - <time to start dynamo>` seconds)

chrisradek:
##### Checklist

- [x] `npm run test` passes
- [x] changelog is added

Fixes #2322

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -55,9 +55,10 @@ AWS.NodeHttpClient = AWS.util.inherit({
 
     // connection timeout support
     if (httpOptions.connectTimeout) {
+      var connectTimeoutId;
       stream.on('socket', function(socket) {
         if (socket.connecting) {
-          var timeoutId = setTimeout(function connectTimeout() {
+          connectTimeoutId = setTimeout(function connectTimeout() {
             if (stream.didCallback) return; stream.didCallback = true;
 
             stream.abort();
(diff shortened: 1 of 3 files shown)


lib/http/node.js
```diff
@@ -55,9 +55,10 @@ AWS.NodeHttpClient = AWS.util.inherit({
 
     // connection timeout support
     if (httpOptions.connectTimeout) {
+      var connectTimeoutId;
```
mpareja: Based on the usage of `connectTimeoutId` in the stream `error` handler below, do you think it would be clearer if this declaration were moved outside of the `if (httpOptions.connectTimeout)` block? I don't think it will have a functional difference given `var` is function scoped - just clarity.

Issue #2341, opened 2018-11-07
Redundant files shipped to npm


SimonSchick:
Related to #2247

I believe the following files don't really need to be shipped to npm:

```
.changes
.github
CHANGELOG.md
CONTRIBUTING.md
SERVICES.md
UPGRADING.md
dist-tools
scripts
```

You might want to use the `files` property inside `package.json` instead as it's opt in rather than opt-out.

srchase:
Thanks for opening this issue.

We're getting this fixed.

We kept the scripts and dist-tools.  The browser-builder is shipped in dist-tools and there are some users of the scripts.

srchase:
Closing out this issue as the changes have been merged and this will get fixed with the next release.

PR #2342, opened 2018-11-07
updated npmignore, fix for issue #2341

srchase:
Update do .npmignore, fix for #2341

##### Checklist

- [x] `npm run test` passes

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,8 @@
+.changes
 .yard*
 .eslintrc
 .travis.yml
+.github
 .gitignore
 .jshintrc
 .npmignore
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -1,13 +1,17 @@
+.changes
```
SimonSchick: As suggested in #2341 you _might_ want to use the `files` property in `package.json`, this should prevent you from ever shipping unwanted files again :)

srchase: @SimonSchick 

Thanks for the suggestion.  We're opting to stick with only using .npmignore for now.  We'll keep that option in mind though.

Issue #2390, opened 2018-11-30
s3.upload treats tags differently for small and large files


sliem:
Using `s3.upload` with tags in the options object behaves differently for small and large files. For small files tag values can be numbers, e.g. `{Key: "k", Value: 10}`, while for large files they must be strings.

With large files I mean files that triggers a multi-part upload where the tagging is done in a separate [`putObjectTagging` call](https://github.com/aws/aws-sdk-js/blob/61f89137708f199bec5dfe7bf8f4754719324501/lib/s3/managed_upload.js#L639) (to resolve #1412.)

The workaround is easy, always convert the number to string. But it is a bit non-obvious that the api works differently depending on the size of the file.

I'm using node.
```bash
$ node --version
v8.11.3
$ grep -A1 aws-sdk package-lock.json
    "aws-sdk": {
      "version": "2.336.0",
      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.336.0.tgz",
      [...]
```
Here is a minimal example that reproduces the issue for me.

```javascript
var aws = require("aws-sdk");
var crypto = require("crypto");

var s3 = new aws.S3({apiVersion: "2006-03-01"});

var bucket = SPECIFY_YOUR_BUCKET_HERE;

tags = [
    {Key: "stringTag", Value: "smallfile"}, 
    {Key: "numberTag", Value: 100}
];

// upload small file, works fine
s3.upload({
    Bucket: bucket,
    Key: "smallfile",
    Body: crypto.randomBytes(20).toString('hex')
  }, {
    tags: [
        {Key: "stringTag", Value: "smallfile"}, 
        {Key: "numberTag", Value: 100}
    ]
  }, function(err, data) {
    if (err) {
      console.log(err.message);
      return;
    }
    console.log(JSON.stringify(data));
    console.log(`download with: aws s3 mv s3://${data.Bucket}/${data.Key} .`);
  });

// upload large file, fails with 
// "Expected params.Tagging.TagSet[1].Value to be a string"
s3.upload({
    Bucket: bucket,
    Key: "largefile",
    Body: crypto.randomBytes(50000000).toString('hex')
  }, {
    tags: [
        {Key: "stringTag", Value: "notsmallfile"}, 
        {Key: "numberTag", Value: 100}
    ]
  }, function(err, data) {
    if (err) {
      console.log(err.message);
      return;
    }
    console.log(JSON.stringify(data));
[...]

PR #2410, opened 2018-12-06
Standardize managed uploader behavior

srchase:
##### Checklist

- [x] `npm run test` passes


srchase:
change for https://github.com/aws/aws-sdk-js/issues/2390

srchase:
@AllanFly120 

Yes.  I tested against S3 and you can set tags like this now on either a small file, or one large enough to do a multipart-upload:

```
tags = [
    {Key: "stringTag", Value: "fooString"}, 
    {Key: "numberTag", Value: 100}
];
```

PR #2573, opened 2019-03-08
Correctly remove whitespace from custom policy

kkopachev:
Currently it incorrectly replaces whitespace characters with policy itself, making it invalid json.

##### Checklist
- [x] `npm run test` passes


cerber717:
wow, how was this missed! 🔔

Issue #2571, opened 2019-03-08
Use Connection: keep-alive to improve application performance


milancermak:
By default, NodeJS [sets](https://github.com/nodejs/node/issues/10774) `Connection: close` and not `Connection: keep-alive` HTTP header which results in sub-optimal performance. Instead of reusing an already established connection between the client and the service, a new one has to be created for every operation.

It has been reported before (#146), recently [publicized](https://theburningmonk.com/2019/02/lambda-optimization-tip-enable-http-keep-alive/) by an AWS Serverless Hero and it's also [recommended by AWS for DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/encryption.howitworks.html) when using encryption at rest. DynamoDB itself responds with a `Connection: keep-alive` header.

Even though all these examples pertain to DynamoDB, I imagine other services will benefit from this change as well. For what it's worth, boto3, the official AWS SDK for Python does not explicitly set a `Connection: close` header.

Please consider setting keep-alive header the on the SDK level so that going forward, the whole community can benefit from this improvement.

srchase:
Changing the default behavior here could introduce issues for existing users of the SDK.  If this were enabled by default, it would require a major version bump.

It is possible to enable keep-alive through configuration.  

That can be done on the global config:
```
const AWS = require('aws-sdk')
const https = require('https');
const agent = new https.Agent({
  keepAlive: true
})

AWS.config.update({
  httpOptions: {
    agent: agent
  }
})
```

Or on a per-client basis:
```
const AWS = require('aws-sdk')
const https = require('https');
const agent = new https.Agent({
  keepAlive: true
});

const DDB = new AWS.DynamoDB({
  httpOptions: {
    agent: agent
  }
});
```

PR #2595, opened 2019-03-26
enable keep alive

achrafsouk:
Keep alive flag in the http agent, keeps sockets around even when there are no outstanding requests, so they can be used for future requests without having to reestablish a TCP connection. Without this flag, every request to Dynamo will be on a separate connection. I have overriden the default http agent in my application code as per below , and the impact in terms of latency/execution time is considerable.

const https = require('https');
const keepAliveAgent = new https.Agent({ keepAlive: true});
const ddb = new AWS.DynamoDB({apiVersion: '2012-10-08', httpOptions: {agent: keepAliveAgent}});

##### Checklist

- [ ] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -142,7 +142,7 @@ AWS.NodeHttpClient = AWS.util.inherit({
     var https = require('https');
 
     if (!AWS.NodeHttpClient.sslAgent) {
-      AWS.NodeHttpClient.sslAgent = new https.Agent({rejectUnauthorized: true});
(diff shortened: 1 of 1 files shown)


achrafsouk:
related to https://github.com/aws/aws-sdk-js/issues/2571

srchase:
Thanks for submitting this PR.

As indicated in #2571, we do not want to impact existing users of the SDK who are not expecting keep-alive to get set.  As you point out, there's a way to configure this by supplying an https.Agent with keepAlive enabled.  We are considering this as the default for [v3 of the SDK](https://github.com/aws/aws-sdk-js-v3).

PR #2652, opened 2019-05-03
Increase in Karma Version to fix Lodash vulnerability

jeremiahwitt:
As indicated by issue #2651, versions of karma lower than 4.0.1 can bring in a version of lodash with a security vulnerability. Bumping up the version of karma to fix this problem.

- [X] `npm run test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "jasmine": "^2.5.3",
     "jasmine-core": "^2.5.2",
     "json-loader": "^0.5.4",
-    "karma": "^2.0.0",
+    "karma": "^4.1.0",
     "karma-chrome-launcher": "2.2.0",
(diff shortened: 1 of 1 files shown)


trivikr:
Thank you @jeremiahwitt for your first contribution!
Looks like the email ID you used for the commit https://github.com/aws/aws-sdk-js/pull/2652/commits/c1e447efd3040021199ef6cf890fb24bf5b4fcdf is not registered with Github

Can you either register it or overwrite the commit history in your fork with the email ID which is already registered?

jeremiahwitt:
@trivikr @AllanFly120 Thanks for the review. I've updated with the email I used for this github account. Would you be able to re-review and merge if appropriate?

Thanks!

PR #2664, opened 2019-05-10
Added example for generating the documentation

### LINKED ISSUE #2499: ImageExists event parameter says invalid for filter = "image-id"
Hi,

When we use the sdk to trigger some functionality of image being available or exists, the filter being passed gives error while testing.

For example:

```
var params = {
  // ... input parameters ...
};
ec2.waitFor('imageExists', params, function(err, data) {
  if (err) console.log(err, err.stack); // an error occurred
  else     console.log(data);           // successful response
});
```

The documentation [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#imageExists-waiter) states that the `Filters — (Array<map>)` can have:

> image-id - The ID of the image.

But when we trigger the follow:
```
var params = {
   Filters: [{
       "image-id": "ami-xxxxxxxxx"
   }]
};
ec2.waitFor('imageExists', params, function(err, data) {
  if (err) console.log(err, err.stack); // an error occurred
  else     console.log(data);           // successful response
});
```

The error states the key `image-id` is not a valid parameter.

tan31989:
Update CONTRIBUTING.md with steps for generating documentation

##### Checklist


apis/ec2-2016-11-15.normal.json
```diff
@@ -12006,7 +12006,7 @@
         },
         "ImageIds": {
           "shape": "ImageIdStringList",
-          "documentation": "<p>The image IDs.</p> <p>Default: Describes all images available to you.</p>",
+          "documentation": "<p>The image IDs. Example: <code>{Name: 'image-id', Values: ['ami-xxxxxxxxxxx']}</code></p> <p>Default: Describes all images available to you.</p>",
```
srchase: Unfortunately, any changes to this or any of the APIs files will be overwritten with the next release.

These API models come from the individual services.   They are consumed by this SDK to generate the clients and documentations.  As such, we take this part of the PR.

CONTRIBUTING.md
```diff
@@ -96,6 +96,24 @@ specific service, for example:
...
+Once the bundler is installed, you need to execute the following commands, to install rake:
+
+```
+bundle install
+```
+
+Once done, please run this command to get the documentation in (doc/latest/_index.html)
+
+```
+bundle exec rake docs:api
```
srchase: This addition is helpful.  It is also noted in the PR template: https://github.com/aws/aws-sdk-js/blob/master/.github/pull_request_template.md

CONTRIBUTING.md
```diff
@@ -96,6 +96,24 @@ specific service, for example:
 AWS_REGION=[set region] node node_modules/cucumber/bin/cucumber.js --tag @s3
 ```
 
+### Testing the documentation changes:
+
+To test the documentation changes, please ensure you have the following requirements installed:
+
+* `gem install bundler`
```
srchase: Perhaps it's worth mentioning that a recent version of Ruby is needed to use gem, but perhaps that's obvious enough.

CONTRIBUTING.md
tan31989: Agreed, that will only be seen while we go to raise a PR. So thought it would be good for contribution folks to see this in Contribution file :)

CONTRIBUTING.md
tan31989: Will still mention it 👍

apis/ec2-2016-11-15.normal.json
tan31989: Then how can we make sure that we make the API models to generate these? Is that in the same repository?

srchase: The API Documentation is managed by the service teams.  The models are shared by all of the SDKs. They won't be able to include this kind of specific example.  It'll need to be removed from this PR before the changes to CONTRIBUTING.md can be accepted.

tan31989: Done, have updated the PR to remove this entry.

tan31989:
@srchase have updated the PR and also the message in commit and PR.

srchase:
@tan31989 

Thanks for the PR!

Issue #2688, opened 2019-05-30
targetInService fails to fetch detail for some targets


shubhamkes:
**Describe the bug**
I have more than 10 targets but targetInService can not fetch details for two targets out of those list and these targets having instances attached to it.

**SDK version number**
2.465.0

**Is the issue in the browser/Node.js?**
Node.js

**Details of the browser/Node.js version**
v10.15.3

**Current behavior**
Callback is never invoked incase of two targets.

**Expected behavior**
Should have return target detail and instances attached to it.

klaytaybai:
Hi @shubhamkes, this operation will not do the callbacks for unhealthy targets.  Can you verify that the targets are in a healthy state via the Console or other SDK operations?

shubhamkes:
Yup, Sorry forgot to mention all those instances are in healthy status. And instance count is more than 2

shubhamkes:
By mistake i pasted wrong sdk version, Have updated now in the issue
2.465.0

ajredniwja:
Hey @shubhamkes,

Apologies for the late reply, are you still facing the issue?

shubhamkes:
I am using alternative method to fetch instance detail 
Thanks

ajredniwja:
I will go ahead and close the issue, please reach out if you have any additional questions.

PR #2690, opened 2019-05-30
doc: add commands for getting SDK version in Bug report Issue template

trivikr:
Two changes:
* Moved SDK version number question after Node.js/browser
* Added hints on how to find SDK version

This change was added after examining issue #2688 where customer shared Node version instead of SDK version in Issue template


trivikr:
Confirmed that Issue template now shows instructions to get SDK version
image

PR #2744, opened 2019-06-26
doc: add lock.yml

trivikr:
We've successfully tested lock bot in v3 for a month ([lock.yml](https://github.com/aws/aws-sdk-js-v3/blob/master/.github/lock.yml)), and would like to introduce it in v2
##### Checklist


trivikr:
Lock App was enabled for aws-sdk-js
Installation link: https://github.com/apps/lock/installations/1059737

PR #2780, opened 2019-07-23
feat: S3 add an option to enable/disable "Expect: 100-continue" headers

ThomasVille:
Since https://github.com/aws/aws-sdk-js/pull/437, when using Riak-CS v2.1.1 as backend, uploading files larger than 1MB using streams on NodeJS leads to corrupted data stored in Riak-CS.

This PR adds an option to prevent the `Expect: 100-continue` header from being added to the content of the file when uploading a large file.

Fixes https://github.com/basho/riak_cs/issues/1327.

##### Checklist

- [ ] `npm run test` passes
- [x] `.d.ts` file is updated
- [ ] ~~run `npm run integration` if integration test is changed~~


trivikr:
@ThomasVille Looks like the email address in the git config for this PR is not registered with GitHub.
Either register that email address with GitHub or override your git history with your registered email address ([details](https://help.github.com/en/articles/why-are-my-contributions-not-showing-up-on-my-profile))

ThomasVille:
@trivikr I tried overriding the history of the branch with my registered email address by rebasing then amending my own commits however it seems that I now appear as the commiter of all the commits of the `awstools` account...

trivikr:
Yup, PR shows conflicts need to be resolved.
You can try fixing it by overriding just your commits. An easy option would be to cherry-pick your commits to new branch (as there are only four of them) and create new PR.

ThomasVille:
Thank you @trivikr for your help.

New PR created here https://github.com/aws/aws-sdk-js/pull/2783.

PR #2790, opened 2019-07-30
Mark CustomErrorResponse props as required

shaharmor:
When any of the props of CustomErrorResponse are not passed, the API returns an `IllegalUpdate` error with the message: "The specified list of custom error responses does not exist or is not valid."

##### Checklist

- [x] `.d.ts` file is updated


ajredniwja:
Hey @shaharmor thank-you for opening a PR, I believe this is generated from the service team and [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudFront.html) mentions this not being required. I can reach out to the team to ask for the details and make a change if necessary.

PR #2795, opened 2019-08-02
Fix to define AWSError as interface, not class

mazamachi:
In a d.ts file, `AWSError` was defined as class. 
However, there is no class or function named `AWSError` in javascript.
So I fix to define it as interface.

This will protect typescript users from mistekenly treating it as class like
```typescript
const result = await this.ses.sendRawEmail(request).promise();
if (result.$response.error instanceof AWSError) {  // AWSError is not an object, so `instaceof` fails!!
    console.error("Error!");
}
```

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/bugfix-typings-60c8932c.json
+++ b/.changes/next-release/bugfix-typings-60c8932c.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "typings",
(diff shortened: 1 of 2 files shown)


rubengmurray:
This makes sense to me.

Issue #1977, opened 2018-03-17
zip does not support timestamps before 1980 aws


devTalha:
We are having projects created with aws codestar. these were working fine. but from today we are facing following issue:
Logs:
'Unable to upload artifact None referenced by CodeUri parameter of GetCompanyRecords resource. zip does not support timestamps before 1980'

Now when i removed aws-sdk module again it works fine. but when i add it again build fails. i am pretty much worried about this. Here is my lambda function.

GetCompanyRecords:
Type: AWS::Serverless::Function
Properties:
  Handler: index.handler
  Runtime: nodejs6.10
  Role:
    Fn::ImportValue:
      !Join ['-', [!Ref 'ProjectId', !Ref 'AWS::Region', 'LambdaTrustRole']]
  Timeout: 10
  Events:
    PostEvent:
      Type: Api
      Properties:
        Path: /getCompanyRecords
        Method: post

Here is pacakge.json
"dependencies": {
        "aws-sdk": "^2.211.0",
        "aws-serverless-express": "latest",
        "body-parser": "^1.18.2",
        "chai": "^4.1.2",
        "chai-http": "^3.0.0",
        "cors": "^2.8.4",
        "country-list": "^1.1.0",
        "express": "^4.16.2",
        "mocha": "^5.0.0",
        "promise": "^8.0.1",
        "provinces": "^1.11.0",
        "underscore": "^1.8.3",
        "uuid": "^3.2.1"
    }

Thanks in advance

zimny:
+1

abajowski:
+1

gbegher:
+1 -- related: https://github.com/feross/ieee754/issues/17

devTalha:
I am facing this issue explicitly with 'aws-sdk'. When i add it in node modules i get the exception but when i remove it's all fine.

devTalha:
@zimny , @abajowski , @gbegher  at the moment following patched fixed my issue:

I added following lines to buildspec.yml after 'npm install'

- ls $CODEBUILD_SRC_DIR
- find $CODEBUILD_SRC_DIR/node_modules -mtime +10950 -exec touch {} \;

dkrosso:
+1

chrisradek:
@devTalha 
Can you share the code that's resulting in the error you're seeing?

chrisradek:
[...]

PR #2850, opened 2019-09-16
Pin ieee754 to 1.1.13

tnir:
This is a follow-up for #1979 as the upstream fixed the problem, which confirmed at https://github.com/feross/ieee754/issues/17#issuecomment-531733308 .

Upstream diff: https://github.com/feross/ieee754/compare/v1.1.8...v1.1.13

##### Checklist
- [x] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/bugfix-Dependency-d3fdb497.json
+++ b/.changes/next-release/bugfix-Dependency-d3fdb497.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Dependency",
+  "description": "Pin the dependency to the latest"
+}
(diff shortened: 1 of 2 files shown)


trivikr:
The issue in #1977 was fixed in #1979 
What's the issue being fixed in this PR?

tnir:
@trivikr Just update the ieee754 package to the latest: https://github.com/feross/ieee754/compare/v1.1.8...v1.1.13

Some other packages are moving to the latest (namely 1.1.13) so a user of application would have only a single version of ieee754 package.

Issue #1236, opened 2016-11-22
[Feedback Requested] [Question] Both DynamoDB.putItem and DocClient.put return "There were x validation errors: InvalidParameterType"


karlwhite:
I'm using DynamoDB/DocumentCient on Lambda, and attempting a simple put operation, thusly:

```javascript
var doc = require('dynamodb-doc');
var dynamo = new doc.DynamoDB();
...
var params = {
    "TableName": "Recipients",
    "Item": {
       "recipient_id": 'XXX',
        "user_id": 'XXX'
    }
};
                        
dynamo.putItem(params, function(err, res) {
    ...
```

and also:

```javascript
var AWS = require('aws-sdk');
var docClient = new AWS.DynamoDB.DocumentClient({region: 'us-west-2'});
...
docClient.put(params, function(err, res) {
    ...
```

However both are returning the same error, as if they are expecting AttributeValue values for the Item:

```json
{
  "errorMessage": "There were 8 validation errors:\n* InvalidParameterType: Expected params.Item['recipient_id'] to be a structure\n* UnexpectedParameter: Unexpected key '0' found in params.Item['recipient_id']\n* UnexpectedParameter: Unexpected key '1' found in params.Item['recipient_id']\n* UnexpectedParameter: Unexpected key '2' found in params.Item['recipient_id']\n* InvalidParameterType: Expected params.Item['user_id'] to be a structure\n* UnexpectedParameter: Unexpected key '0' found in params.Item['user_id']\n* UnexpectedParameter: Unexpected key '1' found in params.Item['user_id']\n* UnexpectedParameter: Unexpected key '2' found in params.Item['user_id']",
  "errorType": "MultipleValidationErrors",
```

I have tried both the Lambda-included aws-sdk, as well as bundling it myself, but both yield the same results.

Can someone please tell me what silly mistake I'm making?

karlwhite:
Also, if I *do* try to use AttributeValues instead, I get the following:

```json
{
  "errorMessage": "shape.toType is not a function",
  "errorType": "TypeError",
  "stackTrace": [
    "Translator.translateScalar (/var/runtime/node_modules/aws-sdk/lib/dynamodb/translator.js:76:16)",
    "Translator.translate (/var/runtime/node_modules/aws-sdk/lib/dynamodb/translator.js:30:26)",
    "Object.<anonymous> (/var/runtime/node_modules/aws-sdk/lib/dynamodb/translator.js:42:25)",
[...]

PR #2924, opened 2019-10-27
Dynamodb translator to ignore unknown values - DocumentClient

OurNewestMember:
Addresses [Issue #1236](https://github.com/aws/aws-sdk-js/issues/1236) where an error occurs in the DocumentClient's translator module when the input native javascript object contains an unserializable property immediately under a top level property of the input object.

Example scenario:
```
const dynamoDBDocumentClient = new require('aws-sdk').DynamoDB.DocumentClient();
Object.prototype.newProp = function() { };
const paramNewJsObject = { TableName: 't',  Key: { tableKey: 'keyValueJsString' }};
// ? behavior due to "paramNewJsObject.Key.newProp" ?
// without fix: throws TypeError 'shape.toType is not a function'
// with fix: dynamodb request is successful despite "newProp"
dynamoDBDocumentClient.get(paramNewJsObject);
```

Without fix, this results in a runtime exception in the dynamodb document client translator.  The proposed solution is to ensure these input values are treated similar as in other parts of the translator logic -- they are ignored from output and/or emitted as javascript/json null in the output of the DocumentClient translator.

Typical use case shown in example above: dynamodb document client user has modified Object.prototype to include a function property before creating and passing a javascript object to a dynamodb DocumentClient API method.  Without the fix, the request will fail during input translation to the native API request parameter object.  With the fix, the dynamodb request proceeds as expected.

##### Checklist
- [x] `npm run test` passes (npm run unit)

### DIFF
--- a/.changes/next-release/bugfix-dynamodb-documentClient-translator-9d00ba98.json
+++ b/.changes/next-release/bugfix-dynamodb-documentClient-translator-9d00ba98.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "dynamodb-documentClient-translator",
+  "description": "Fixes runtime exception in the documentClient translator when a native javascript input object contains a top-level property with a non-serializable value (eg, when the input object was created after Object.prototype was modified to include a property whose value is a function) -- see Issue #1236"
+}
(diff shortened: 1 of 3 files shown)


OurNewestMember:
Currently reviewing test results following latest change in lib code...

PR #2964, opened 2019-11-19
fix: signed IMDS workflow

trivikr:
##### Checklist

- [x] `npm run test` passes
- [ ] `.d.ts` file is updated


trivikr:
To unblock this PR, I will add update to type definitions in a separate PR

PR #2967, opened 2019-11-20
chore: npmignore eslintignore and vscode

G-Rath:
These files don't need to be published.

I also question if `scripts`, and `dist-tools` also need to be published - on disk `scripts` is about 100kb alone.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,12 +1,14 @@
 .changes
 .yard*
 .eslintrc
+.eslintignore
 .travis.yml
 .github
 .gitignore
 .jshintrc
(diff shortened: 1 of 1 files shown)


trivikr:
Thank you @G-Rath for the contribution

> I also question if scripts, and dist-tools also need to be published - on disk scripts is about 100kb alone.

Please [create an issue](https://github.com/aws/aws-sdk-js/issues/new/choose) to discuss this question

G-Rath:
@trivikr thanks - I've made #3159 to track this question :)

Issue #2968, opened 2019-11-21
Crash on node 12 with certain Http Error Responses


griffinmyers:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js/issues) and didn't find any solution
  
**Describe the bug**
Modern versions of node (at least `>10`, confirmed on `v12.13.0`) do not handle `Error` objects where the `name` property is a non-string properly. As described in https://github.com/nodejs/node/issues/30572, the following simple program will crash when running on node `12.13.0`:

```node
const err = new Error('bloop');
err.name = 404;
console.log(err);
```

with the message: 

```
internal/util/inspect.js:880
      name.endsWith('Error') &&
           ^

TypeError: name.endsWith is not a function
    at formatError (internal/util/inspect.js:880:12)
    at formatRaw (internal/util/inspect.js:681:14)
    at formatValue (internal/util/inspect.js:569:10)
    at inspect (internal/util/inspect.js:223:10)
    at formatWithOptions (internal/util/inspect.js:1651:40)
    at Object.Console.<computed> (internal/console/constructor.js:272:10)
    at Object.log (internal/console/constructor.js:282:61)
    at Object.<anonymous> (/Users/williammyers/projects/js/error-12.js:3:9)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
```

It turns out, the AWS SDK will set the `name` property on certain errors to be the numeric HTTP status code of a failed HTTP request/response. See [lib/protocol/query.js#L55-L57](https://github.com/aws/aws-sdk-js/blob/master/lib/protocol/query.js#L55-L57) and [lib/util.js#L592-L597](https://github.com/aws/aws-sdk-js/blob/master/lib/util.js#L592-L597). 

I can simulate this edge case with the following test program:

##### `aws-test.js`
```node
const AWS = require('aws-sdk/lib/core');
[...]

PR #2978, opened 2019-11-26
fix: console.log ing certain errors throws exception

griffinmyers:
Fixes: https://github.com/aws/aws-sdk-js/issues/2968
Ref: https://github.com/nodejs/node/issues/30572

Without the change, the new tests throw with the following errors on node `12.13.0`: 

```
  1) AWS.Protocol.Query extractError returns an empty error when the body is blank:
     TypeError: name.endsWith is not a function
      at formatError (internal/util/inspect.js:914:13)
      at formatRaw (internal/util/inspect.js:703:14)
      at formatValue (internal/util/inspect.js:591:10)
      at Object.inspect (internal/util/inspect.js:221:10)
      at Context.<anonymous> (test/protocol/query.spec.js:147:21)
      at processImmediate (internal/timers.js:439:21)
      at process.topLevelDomainCallback (domain.js:131:23)

  2) AWS.Protocol.Query extractError returns an empty error when the body cannot be parsed:
     TypeError: name.endsWith is not a function
      at formatError (internal/util/inspect.js:914:13)
      at formatRaw (internal/util/inspect.js:703:14)
      at formatValue (internal/util/inspect.js:591:10)
      at Object.inspect (internal/util/inspect.js:221:10)
      at Context.<anonymous> (test/protocol/query.spec.js:157:21)
      at processImmediate (internal/timers.js:439:21)
      at process.topLevelDomainCallback (domain.js:131:23)
```

##### Checklist

- [x] `npm run test` passes

### Alternative Idea

As I call out in https://github.com/aws/aws-sdk-js/issues/2968, there's an argument to be made for this being a breaking api change. To wit, the following code would break in some circumstances: 

```js
// ...

try {
  await sqs.sendMessage({ MessageBody: messageBody, QueueUrl: queueUrl }).promise()
} catch (e) {
  if (e.name === 500) { // predicate no longer true after upgrade
    // log somewhere special? retry logic?
  }
}
```

If this were an absolute show-stopper, we could consider an alternative, which is to declare a [custom inspect function](https://nodejs.org/api/util.html#util_custom_inspection_functions_on_objects). 

This special function is invoked whenever the error is passed to `util.inspect` (for instance, inside of `console.log`) -- there we can do some internal type coercion before and after calling the native `util.inspect` so that the bug is averted, while preserving state observers may depend on. 

The downside is it skews "clever". Here's an example implementation that also passes the new tests: 

```diff
diff --git a/lib/protocol/query.js b/lib/protocol/query.js
index eddb7aa6..ce54d176 100644
--- a/lib/protocol/query.js
+++ b/lib/protocol/query.js
@@ -3,6 +3,24 @@ var util = require('../util');
 var QueryParamSerializer = require('../query/query_param_serializer');
 var Shape = require('../model/shape');
 var populateHostPrefix = require('./helpers').populateHostPrefix;
+var nodeUtil = require('util');
+
+var inspect = Symbol.for('nodejs.util.inspect.custom');
+
+var customInspect = function() {
+  var name = this.name;
+
+  this.name = String(name);
+  this[inspect] = undefined;
+
+  var result = nodeUtil.inspect(this);
+
+  this.name = name;
+  this[inspect] = customInspect;
+
+  return result;
+};
+

 function buildRequest(req) {
   var operation = req.service.api.operations[req.operation];
@@ -48,17 +66,17 @@ function extractError(resp) {
   if (data.Error) data = data.Error;
   if (data.Code) {
     resp.error = util.error(new Error(), {
       code: data.Code,
       message: data.Message
     });
   } else {
     resp.error = util.error(new Error(), {
       code: resp.httpResponse.statusCode,
       message: null
     });
   }
+
+  resp.error[inspect] = customInspect;
 }
```


SimonWoolf:
For reference there was already an open PR in this repo to fix this bug: https://github.com/aws/aws-sdk-js/pull/2963 , though mine does it in a slightly different way.

griffinmyers:
> For reference there was already an open PR in this repo to fix this bug: #2963 (in a slightly different way)

Doh! Wish I had caught that in my search before filing/fixing myself!

Fwiw, node has merged a fix as well: https://github.com/nodejs/node/issues/30572.

griffinmyers:
Closing since https://github.com/aws/aws-sdk-js/pull/2963 was merged.

PR #3121, opened 2020-02-27
Have presigner preserve accesssKeyIds with colons in them

tipabu:
If a user supplies an accessKeyId like `test:tester`, the v2 signer will (correctly) craft an Authorization header like

    Authorization: AWS test:tester:<signature>

When creating a presigned URL, however, we previously would generate a query string like

    ?AWSAccessKeyId=test&Expires=<expiry>&Signature=tester

which changes the accessKeyId used and completely ignores the calculated signature. Now, have the presigner generate a query string like

    ?AWSAccessKeyId=test%3Atester&Expires=<expiry>&Signature=<signature>

##### Checklist

- [x] `npm run test` passes

### DIFF
--- a/.changes/next-release/bugfix-Signers.Presign-6cf33bd5.json
+++ b/.changes/next-release/bugfix-Signers.Presign-6cf33bd5.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Request Signing",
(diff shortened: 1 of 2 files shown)


tipabu:
Anything else I could do to help this along?

ajredniwja:
@tipabu

tipabu:
Thanks, @ajredniwja! Anything else needed to actually have it merge?

PR #3168, opened 2020-03-19
chore: npmignore scripts folder

G-Rath:
Adding `scripts` folder to `.npmignore` per discussion in #3159.

This might be breaking, and so needs to be checked with the team.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -26,7 +26,7 @@ Gemfile.lock
 features
 karma.conf.js
 Rakefile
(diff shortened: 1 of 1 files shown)


G-Rath:
@ajredniwja `services-table-generator.ts` is included in the package, so that can at least be ignored if not the whole folder :)

ajredniwja:
I dont believe this change will be made. Thanks for submitting the PR. 

Closing it now.

Issue #3189, opened 2020-03-31
MaxRetriesExceeded set to 1 when final retry succeeds


DRNagar:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js/issues) and didn't find any solution
  
**Describe the bug**

The maxRetriesExceeded metric is set to 1 when the API call is successful.

This occurs if the call is successful on the final retry. This is due to the associated responses retryCount being equal to the maxRetries. [This sets the MaxRetriesExceeded to 1 as retryCount >= maxRetries](https://github.com/aws/aws-sdk-js/blob/master/lib/service.js#L435-L441).

**Research on the bug**

The responses retryCount is incremented in two places: 
1. A request is [restarted] (https://github.com/aws/aws-sdk-js/blob/master/lib/event_listeners.js#L196) due to failing in the state [build or afterBuild](https://github.com/aws/aws-sdk-js/blob/master/lib/request.js#L45-L46).
2. After a retry[retry] (https://github.com/aws/aws-sdk-js/blob/master/lib/event_listeners.js#L462-L484)

In both these cases the retryCount is incremented after a failing call on the condition that retryCount<maxRetries. This implies that on the final retry the retryCount is equal to the maxRetries.

**Is the issue in the browser/Node.js?**
The issue should be in both

**If on Node.js, are you running this on AWS Lambda?**
Yes

**Details of the Node.js version**
v12.13.1

**SDK version number**
v2.597.0

**To Reproduce (observed behavior)**
The observed behavior in evident in our metrics
```
{ 
   ...
   MaxRetriesExceeded: 1,
   StatusCode2xx: 1
   ...
}
```

Using the test framework, adding the following case to the unit tests succeeds:
```
    {
      'description': 'Test if final retry succeeds maxRetriesExceeded is not set',
      'configuration': {
        'accessKey': 'myaccesskey',
        'region': 'us-west-2',
        'environmentVariables': {
          'AWS_CSM_ENABLED': 'true'
        },
[...]

PR #3196, opened 2020-04-02
Set MaxRetriesExceeded on monitoring event only when the final retry fails

DRNagar:
#### Related Issue
https://github.com/aws/aws-sdk-js/issues/3189

##### Checklist
- [x] `npm run test` passes


trivikr:
Thank you @DRNagar for your contribution - for identifying, diving deep and fixing this issue in Monitoring 🎉

PR #3289, opened 2020-06-01
Build for node

tmathern:
Adds a script, `node-builder.js`, similar to `browser-builder.js`, to build only parts of the SDK. We needed this to reduce the size of our dependencies, since we use only some services. I therefore added a script that builds a version for node only (+refactored so that node and browser builds use the same underlying functions, but with different parameters for the browserify part). 

##### Checklist

- [x] `npm run test` passes
- [ ] `.d.ts` file is updated (no change needed)
- [ ] run `bundle exec rake docs:api` and inspect `doc/latest/index.html` if documentation is changed (no change)
- [ ] run `npm run integration` if integration test is changed (no change)

### DIFF
--- a/.changes/next-release/feature-Build-910cd6ed.json
+++ b/.changes/next-release/feature-Build-910cd6ed.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "Build",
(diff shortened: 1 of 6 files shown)


ckant:
Hi @tmathern, were you able to get this working?

PR #3298, opened 2020-06-09
Fix nextPage typing in requests

gruenem:
The fixes the return type of the `nextPage()` method. Before, it would
return `Request<D, E>|void` which is hard to work with, because you
couldn't runtime test for `void` (everything can be void). This change
splits up the definition into two overloads: one without a callback
provided that returns `Request<D, E> | null` and a second for the
callback version that always returns `void`. This also matches the
documentation in the implementation now.

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/bugfix-Paginator-887637f3.json
+++ b/.changes/next-release/bugfix-Paginator-887637f3.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Paginator",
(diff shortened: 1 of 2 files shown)


kirananto (approved):

siddsriv (approved):

kiran-esper:
Much needed!

PR #3316, opened 2020-06-17
Ensure we don't finish ManagedUpload while still reading

### LINKED ISSUE #3311: ManagedUpload sometimes truncates stream payloads
  
**Describe the bug**
Sometimes ManagedUpload will prematurely set `isDoneSending` before the final bytes have been read from the stream in `fillStream`. This results in truncated files being uploaded to S3.

Node.js

No

v12.18.0

**SDK version number**
2.697.0

The behavior is non-deterministic unfortunately – I guess due to network timing, etc. But most times, this will fail (EDIT: updated with larger file and it fails more regularly):
```js
const assert = require('assert')
const tar = require('tar')
const S3 = require('aws-sdk/clients/s3')

// In my case, this ends up as a ~190MB stream
const stream = tar.create({ gzip: true }, ['my_large_file.bin'])
const upload = new S3().upload({ Bucket: bucket, Key: key, Body: stream })

await upload.promise()

assert.equal(upload.totalChunkedBytes, upload.totalBytes)
```

**Expected behavior**
The above should not error, as it implies that `totalChunkedBytes` was modified after `totalBytes` had been calculated. This means that `totalBytes` wasn't the correct length of the stream, and everything ended prematurely.

**Additional context**
I suspect the problem is this:

1. `fillStream` calls `self.body.read()` here: https://github.com/aws/aws-sdk-js/blob/1bb6b43ae8336c6a8e95eb33042291c1697422cd/lib/s3/managed_upload.js#L422-L423
2. *If* the stream ends during that read, then `end` is emitted, which is managed here: https://github.com/aws/aws-sdk-js/blob/1bb6b43ae8336c6a8e95eb33042291c1697422cd/lib/s3/managed_upload.js#L189-L192
3. As can be seen there ^ this causes *another* call to `fillStream` (via `fillQueue`). So if this is all synchronous, then our current stack looks like this:
```
upload.fillStream()
  upload.body.read()
    upload.body.emit('end')
      upload.body.on('end')
        upload.fillStream()
```
4. Here's where the problem surfaces: this nested call to `fillStream()` will assume that everything has finished (because `isDoneChunking` is now true), and mark it as such here: https://github.com/aws/aws-sdk-js/blob/1bb6b43ae8336c6a8e95eb33042291c1697422cd/lib/s3/managed_upload.js#L448-L455
5. BUT, everything hasn't finished – the `read()` call that triggered the `end` event still hasn't returned its payload. When the stack unwinds and it does return, it will be added to the buffers (and `totalChunkedBytes`): https://github.com/aws/aws-sdk-js/blob/1bb6b43ae8336c6a8e95eb33042291c1697422cd/lib/s3/managed_upload.js#L422-L428
6. But by then it's too late, because both `isDoneChunking` and `isDoneSending` have been set to true by the nested call, so this final chunk will never be sent.

So the root cause is that the `fillStream()` calls become nested. There are few ways to address this – one would be to assume that not everything is done if you receive a non-null buffer from `read`. The other would be (I think) to make the `end` handler asynchronous so you don't end up with nested `fillStream()` calls in this case.

mhart:
Fixes #3311

Ensure that we don't call `fillStream` from within another `fillStream` when we're handling the `end` event.

##### Checklist

- [x] `npm run test` passes


mhart:
Not sure if this is the approach you'd want to go with to fix #3311 – but it's one approach nevertheless

PR #3323, opened 2020-06-23
Add typing to `AWSError` starting with DynamoDB (draft)

rubengmurray:
Draft proposal to add better typing to `AWSError` for specific modules. I feel like this addition provides good value to the SDK (which is proven for my use case at least).

However, there are some `.d.ts` structure / placement topics to discuss here as currently `AWSError` (which is actually a class - previous PR proposes to change this [here](https://github.com/aws/aws-sdk-js/pull/2795) which I have incorporated) is just one file.

I wonder whether including a separate `error.d.ts` inside each respective `/lib/xx` that extends the generic AWSError would be a good structure to proceed with?

##### Checklist

- [ ] `npm run test` passes
- [x] `.d.ts` file is updated


rubengmurray:
Looks like enums and declaration files don't mix particularly well, causing the tests to fail on this.

I've added a string type for `DynamoDBError` in an `error.d.ts` inside `lib/dynamodb`.

Still open to suggestions on this, wondering if now the tests aren't failing there will be some feedback :)

gordonmleigh:
@rubengmurray regarding enums, I think you want `export declare enum DynamoDBErrorCode {` (note added `declare` keyword)

rubengmurray:
Yep @gordonmleigh I had already tried that in the first attempt which failed the tests on this repo:

https://github.com/aws/aws-sdk-js/pull/3323/commits/de69e26eb06a600288243c72dc1f1e6f4147c028 .. :(

kirananto (approved):

PR #3397, opened 2020-08-14
Add fallback to use the Code attribute for error codes

priyajeet:
Allow the error code extraction to happen also when the body has `Code` instead of `code` as the attribute. Preference is still maintained as before. This makes it consistent with how the message is being extracted where it checks for both `message` and `Message`. This also makes it consistent with how the Go SDK processes these attributes: https://github.com/aws/aws-sdk-go/blob/6ca8a5496cb4723ad4ef1e6709ed593ea81ce000/private/protocol/restjson/unmarshal_error.go#L131-L134

Issue encountered when trying to parse a response from the Amazon Interactive Video service API, which has an error body like below

```
{
  "Code" : "SomeException",
  "Message" : "ivs.SomeException: Some message",
  "RequestID" : "some guid"
}
```

##### Checklist

- [X] `npm run test` passes
- [ ] ~`.d.ts` file is updated~
- [ ] ~run `bundle exec rake docs:api` and inspect `doc/latest/index.html` if documentation is changed~
- [ ] ~run `npm run integration` if integration test is changed~
- [ ] ~non-code related change (markdown/git settings etc)~

### DIFF
--- a/.changes/next-release/bugfix-protocol-2b15d604.json
+++ b/.changes/next-release/bugfix-protocol-2b15d604.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "protocol",
+  "description": "Allow error code extraction from pascal cased Code attribute in the JSON body response"
+}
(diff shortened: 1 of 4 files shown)


ajredniwja (approved):

ajredniwja:
Thanks for your contribution.

PR #3398, opened 2020-08-15
Deep typescript client imports no longer import all clients

### LINKED ISSUE #3034: Too much memory used for typescript compilation
- [ ] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js/issues) and didn't find any solution
  
**Describe the bug**
Too much memory used for typescript compilation. Duplicate of https://github.com/aws/aws-sdk-js/issues/2846. 

Bringing this up again since I think it is a serious enough problem that we can't compile even the hello world of the SDK in a EC2 linux t3.micro machine. Running it with --max-old-space-size=4096 and 1 GB swap works, but I don't think it should take so much memory to compile just this small program. 

Node.js

No

v12.13.0

**SDK version number**
Example: 2.597.0

Just trying to compile this one line TS file results in a out of memory error on a t3.micro AWS instance with nothing else running on it!
```
import  CloudFront from "aws-sdk/clients/cloudfront";
```

**Expected behavior**
I wanted the program to compile

**Screenshots**

**Additional context**

JakeGinnivan:
Separates global config from service config

ConfigurationServicePlaceholders references all clients, this means when a single client is deep imported it still transitively references all clients.

The impact of this is TypeScript needs to parse all type definitions in the SDK, and there is no way for a project to avoid this.

By separating client config from global config clients can be deep imported without pulling all type definitions in, greatly reducing build time

Fixes #3034

I created a simple repo demonstrating the issue at https://github.com/JakeGinnivan/aws-sdk-deep-imports. I then used `patch-package` to apply this change directly to node modules, giving me the before/after.

For ease of reviewing I have separated the regeneration of the client definitions from the main changes, if you check out the first commit it contains the actual changes.

## Example program

``` ts
import S3 from 'aws-sdk/clients/s3'
```

### Before
```
Files:                         259
Lines:                      439618
Memory used:               375717K
Parse time:                  2.19s
Program time:                2.62s
```

### After

```
Files:                         22
Lines:                      31187
Memory used:               54651K
Parse time:                 0.36s
Program time:               0.43s
```

##### Checklist

- [x] `.d.ts` file is updated


trivikr:
@JakeGinnivan Thank you for the PR!

AWS JS SDK team will prioritize reviewing this PR today (8/25)

ankon:
FWIW: Before this change one could import `ConfigurationOptions` from `aws-sdk/lib/config`, now this is exported from `config-base` instead. Same for some other types.

This broke one of our users of the aws-sdk, and I had to explicitly bump the dependency in that user.

JakeGinnivan:
Ah apologies, did not think about that. I can put up another PR which re-exports the types from config-base in config fixing the breaking change

PR #3432, opened 2020-09-02
Remove coffeescript

mhart:
It seems it's only needed for a couple of files that can be converted into JS.

It's also misleading to have all the "Generated by CoffeeScript" comments in JS files now that the corresponding *.coffee files no longer exist – it implies they aren't the source of truth (they are now) and it also misleads tools like GitHub into thinking the files are still generated and so it doesn't display diffs for them by default.

##### Checklist

- [x] `npm run test` passes


mhart:
Yup, still an issue afaict

PR #3438, opened 2020-09-04
Fix types for fetching Credentials from AWS.config

workeitel:
Making the error optional since it's only set if there is an error and
exposing optional second parameter with the actual parameter on
`getCredentials` method. That means the `CredentialsOptions` needs to be
exported as well in order to make use of it.

https://github.com/aws/aws-sdk-js/blob/master/lib/config.js#L386

##### Checklist

- [x] `npm run test` passes
- [X] `.d.ts` file is updated


alexforsyth (approved):
LGTM!

PR #3447, opened 2020-09-14
fix: profile assume-role credentials ignore region

rix0rrr:
This PR changes three behaviors of the SDK around AssumeRole credentials
to match boto/AWS CLI.

* AssumeRole credentials uses the region in the `[default]` profile,
  ignoring the region specified in the requested profile.
* AssumeRole credentials fail if `AWS_STS_REGIONAL_ENDPOINTS=regional`
  and no region has been configured (SDK defaults to `us-east-1` here).

Had to make 2 larger-scope changes to get this to work:

* INI loading: when a profile exists both in `credentials` and in
  `config`, all the values in `config` would be overwritten with
  all the values in `credentials` (they are now merged).
* Test helpers: `createSpy` would wreck the `this` in the called spy
  function: it would be set to be the object that `spyOn` was originally
  called on, which is not helpful when mocking a class method. Changed
  to leave `this` intact in the spy call. Hope this doesn't break
  any tests.

##### Checklist

- [ ] `npm run test` passes
- [ ] `.d.ts` file is updated


rix0rrr:
fwiw not all tests pass on my machine, but half of them failed to start with before I started making the change... :(

ajredniwja (approved):

PR #3489, opened 2020-10-10
Adds name to all CredentialProviderChain default providers in node_loadr

joaomilho:
While debugging some auth issues I've noticed the functions set in the CredentialProviderChain were anonymous, making it harder to debug and understand the chain. Loggin the providers (`AWS.config.credentialProvider.providers`) one would see:

```
[
  [Function],
  [Function],
  [Function],
  [Function],
  [Function],
  [Function],
  [Function]
]
```

Sure, it's easy to simply map these with a `toString`, therefore getting:

```
[
  "function () { return new AWS.EnvironmentCredentials('AWS'); }",
  "function () { return new AWS.EnvironmentCredentials('AMAZON'); }",
  'function () { return new AWS.SharedIniFileCredentials(); }',
  'function () { return new AWS.ECSCredentials(); }',
  'function () { return new AWS.ProcessCredentials(); }',
  'function () { return new AWS.TokenFileWebIdentityCredentials(); }',
  'function () { return new AWS.EC2MetadataCredentials(); }'
]
```

These are quite easy to read, but that's just because the function's content is quite trivial. Having a proper name would not depend on that while also not requiring people to inspect with toString, improving the API's transparency and understanding to developers.

With this change the simple log will look like:

```
[
  [Function: EnvironmentCredentialsAWS],
  [Function: EnvironmentCredentialsAMAZON],
  [Function: SharedIniFileCredentials],
  [Function: ECSCredentials],
  [Function: ProcessCredentials],
  [Function: TokenFileWebIdentityCredentials],
  [Function: EC2MetadataCredentials]
]
```

Better name could be set by using `Object.defineProperty(fn, "name", value: "Better names")`, but this fails in the pipeline for nodes < 4.

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/feature-CredentialProviderChain-be037bf6.json
+++ b/.changes/next-release/feature-CredentialProviderChain-be037bf6.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "CredentialProviderChain",
+  "description": "Names all default functions in the node_loader"
+}
--- a/lib/node_loader.js
+++ b/lib/node_loader.js
@@ -60,13 +60,41 @@ require('./credentials/process_credentials');
 // AWS.CredentialProviderChain.defaultProviders in
 // credentials/credential_provider_chain.js
 AWS.CredentialProviderChain.defaultProviders = [
-  function () { return new AWS.EnvironmentCredentials('AWS'); },
-  function () { return new AWS.EnvironmentCredentials('AMAZON'); },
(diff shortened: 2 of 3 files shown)


joaomilho:
I'm not sure how to solve the CodeBuild problem, it seems it doesn't allow external access?

joaomilho:
Should I close this?

Issue #2202, opened 2018-08-15
S3.putObject doesn't support non-node ReadableStream


aspectgfg:
S3 docs claim that S3.putObject supports ReadableStream

```
Body — (Buffer, Typed Array, Blob, String, ReadableStream)
Object data.
```

But when used in amplify-js in a React Native app, Readable Stream is not supported.

```
StorageClass - error uploading: Error: Unsupported body payload object
    at ManagedUpload.self.fillQueue
```

Currently, streams are only supported in NodeJS environments.

`https://github.com/aws/aws-sdk-js/blob/f7728f45dd80c5d4919652720a773d343aaab8af/lib/s3/managed_upload.js#L178`
```
if (self.sliceFn) {
      self.fillQueue = self.fillBuffer;
    } else if (AWS.util.isNode()) {
      var Stream = AWS.util.stream.Stream;
      if (self.body instanceof Stream) {
```

Please expand stream support for ReadableStream as defined by:
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream

Or provide a general definition or interface of ReadableStream.

Feature necessary since streams keep IO at low memory footprint. Currently, loading entire HD videos into memory for an upload is significantly suboptimal.

chrisradek:
@kidcosmic 
This documentation was written long before browsers supported streams, and we likely won't be able to support them in browsers until a major version bump when we can look at using `fetch` instead of `XmlHttpRequest`. We can definitely update our documentation to make it clear that this only works with node.js streams today.

Are you able to use Blobs? I'm not sure how they differ in react native versus in the browsers, but if they work like the [File](https://developer.mozilla.org/en-US/docs/Web/API/File) object in browsers, that may be an alternative way to upload large files.

I didn't know react native supported ReadableStream with fetch. Is is officially supported, or provided through a 3rd party package?

kidcosmic:
ReadableStream is actually supported by the JavascriptCore, ie, Safari. We're actually shaping another stream into ReadableStream via the ReadableStream constructor:

```
function convertStreamToReadableStream(_in) {
  return new ReadableStream({
    start(controller) {
      _in.open()
      _in.onData((chunk) => {
        controller.enqueue(chunk)
      })
      _in.onError((error) => {
        controller.error(error)
      })
      _in.onEnd(() => {
        controller.close()
      })
    }
  })
}
```

chrisradek:
[...]

Issue #2819, opened 2019-08-22
Use Fetch in preference of XMLHttpRequest (on supported platforms)


newton3:
Hey guys,
I am trying to make aws calls in a service worker for reliability and performance reasons but seems like it is not supported due to dependency on XMLHttpRequest (Similar issue here - https://github.com/aws/aws-sdk-js/issues/1902)

While looking for alternates I could not find any aws recommended way of using either - 
1. REST Api directly so that I can use fetch, OR
2. Override the default mechanism in the aws-sdk to use fetch
3. OR any other suitable alternate without incurring additional costs (like using api gateway is very expensive for us)

It would be nice to know what are the team's thought on this and are there any future plans to address this.

We considered running on main UI thread but here are the problems - 
1. The SDK itself is 230kb and adds immensely to the bundle.
2. In order to ensure reliability for the Kinesis calls, we need to wait until the calls succeed, and we need the code to run on a separate thread to be more performant, like service-workers.

MickL:
Im running in the same issue like you an #1902. I wonder why #1902 has been closed:

Any request that I want to make results in an error: `XMLHttpRequest is not defined` in my serverless environment (Cloudflare Worker). This should be the case for Amazon Lambda, too.

Example:

```
const s3 = new S3({
  // credentials
});

this.s3.getObject({
          Bucket: this.s3Config.bucket,
          Key:    id,
        },
        (error, data) => {
             // Error: XMLHttpRequest is not defined
        }
);
```

zzzxtreme:
this is not an issue with aws sdk, but rather cloudflare which blocks the use of xmlhttp. like the solution from other post, just change xmlhttp to fetch

jimmywarting:
ppl are [asking](https://github.com/jimmywarting/StreamSaver.js/issues/151) how they can use my [StreamSaver](https://github.com/jimmywarting/StreamSaver.js) to download files without using much of the memory in the browser.
currently it seems like it isn't possible due to this issue

It would also have value to provide them with the actual response object instead of giving developers a blob/arrayBuffer or a stream.

developers can use the response object to convert it to any shape they like 
- .blob() // as promise
- .text() // as promise
- .arrayBuffer() // as promise
- res.body // readable stream

developer can then be able to pipe it, cache it to cache storage or save it to native fs system or any browser transformer (de/compress(gzip) & textTransformer)

[...]

PR #3505, opened 2020-10-19
Streams api version 3 - Whatwg Fetch & Streams Support

heri16:
This PR adds support for WHATWG Fetch and ReadableStream for use in ServiceWorkers and modern browsers.

`AWS.Request.createReadStream()` is now supported on the browser, allowing use with [StreamSaver.js](https://jimmywarting.github.io/StreamSaver.js/) to zip huge S3 folders within the browser.

##### Checklist

- [ ] `npm run test` passes
- [ ] `.d.ts` file is updated


heri16:
Relates to #2819

heri16:
Fixes #2202

lib/http/fetch.js
```diff
@@ -0,0 +1,159 @@
...
+    var request = new Request(href, req);
+
+    // Send request
+    fetch(request).then(function(response) {
+        if (!response.ok) throw Error(response.statusText);
+        return response;
+    }).then(function(response) {
+      emitter.statusCode = response.status;
+      emitter.headers = self.parseHeaders(response.headers);
+      emitter.abort = function() { response.body.cancel(); };
```
jimmywarting, line 59: I believe you should be using [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to abort a request/response

lib/http/fetch.js
```diff
@@ -0,0 +1,159 @@
...
+        code: 'BadReplyError'
+      }));
+    });
+
+    return emitter;
+  },
+
+  writeBody: function writeBody(emitter, response) {
+    var body = response.body;
+    var totalBytes = parseInt(response.headers.get('Content-Length'), 10);
```
jimmywarting, line 95: just a heads up: if the response is encoded with gzip, br, deflate or anything then that will be what the content-length will also be (not the actual file size) so when you read the all the chunks then you will only increase the loadedBytes with uncompressed data.

which means loadedBytes might be higher than content-length in some scenarios

lib/http/fetch.js
heri16: Yes indeed. 😄 I'm just concerned on the level of browser support for AbortSignal. Need research as the pollyfills I've seen so far don't do it correctly: https://www.npmjs.com/package/abortcontroller-polyfill

lib/http/fetch.js
heri16: This issue is very tricky indeed due to the fetch implementation, and I don't have a solution yet.

See discussion here:
https://github.com/w3c/ServiceWorker/issues/339#issuecomment-372304884

heri16: Any suggestions are welcomed. :smiling_face_with_three_hearts:

jimmywarting:
```js
    const k = 1024;
    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
```
Think you should change the size array in your codepen to: `['Bytes', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB', 'YiB']`

lib/http/fetch.js
jimmywarting: There is also this observer issue: https://github.com/whatwg/fetch/issues/607 but nothing have happened on that front for a long time now.

You could read the content-encoding as well and if it's null or "identity" then the content-length is the equivalent to the actual file size and it can be ok to use the progress observer. it's also possible to send a `accept-encoding: identity` request header to ask for raw bytes.  but i might best be done with a HEAD request
I don't know exactly how the aws sdk/api looks like, but there looks to be a size metadata when you fetch the file list that may be accurate.

I would probably just ditch the hole progress stuff and leave that up to the developer to handle b/c it can be unreliable. or give them the option to choose between XHR and Fetch (at least until https://github.com/whatwg/fetch/issues/607 have been implemented)

Also if you use the size option in streamsaver than you don't need to be subscribed to any progress event listener since browser native UI will show one for you. but it needs to be the uncompressed size as well - can't be the same content-length

jimmywarting:
Im just a little bummed that you can't just get the original Response object. You can do so much more with it.
like calling blob, arraybuffer, json, formData and text you are also able to cache the Response to the Cache store if you want. 

(yes responding with a multipart formData response from the server might seem odd, but it can also be a grate hack to send multiple binary files to the client without having to include a hole zip decoding library - `await response.formData()`)

heri16:
> Im just a little bummed that you can't just get the original Response object. You can do so much more with it.
> like calling blob, arraybuffer, json, formData and text you are also able to cache the Response to the Cache store if you want.
> 
> (yes responding with a multipart formData response from the server might seem odd, but it can also be a grate hack to send multiple binary files to the client without having to include a hole zip decoding library - `await response.formData()`)

If the project maintainers are alright with it, I would love to add a AWS.request.fetch().

The current implementation emphasizes compatibility with the existing documented APIs and follows internal design patterns of AWS.NodeHttpClient to ensure nothing breaks, but we know that is not optimal.

See: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-a-response-event-handler.html

heri16:
The only reason we are not using S3 presignedUrl is because we have a Cloudfront distribution fronting our bucket, and we prefer signatures embedded within the http headers because our Cloudfront signedUrls are already pretty large to fit within the queryString limits.

See:
https://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html

jimmywarting:
> If the project maintainers are alright with it, I would love to add a AWS.request.fetch().

I would preferred that too! 
I'm not a huge fan of bundling nodes core modules and bringing them into browser when they aren't needed (includes streams, buffer, eventemitter among others) 
Wrote a guideline before of how to best write cross platform applications https://github.com/cross-js/cross-js

heri16:
> > If the project maintainers are alright with it, I would love to add a AWS.request.fetch().
> 
> I would preferred that too!
> I'm not a huge fan of bundling nodes core modules and bringing them into browser when they aren't needed (includes streams, buffer, eventemitter among others)
> Wrote a guideline before of how to best write cross platform applications https://github.com/cross-js/cross-js

Nice guide!

Bundle size is definitely a concern for our less performant code that uses [stream-browserify](https://github.com/browserify/stream-browserify) or [webpack node-config](https://github.com/webpack/node-libs-browser) to monkeypatch `AWS.util.stream`: https://gist.github.com/heri16/e54ef791e5c9b049941724e315663b71

Our Codepen demo bundle size is much smaller.

PR #3510, opened 2020-10-23
Fix typings for credentials provider chain

workeitel:
`Credentials[]|providers[]` means either an array with `Credentials` or
an array with `providers`. But in reality it's a array with both.

##### Checklist

- [X] `npm run test` passes
- [X] `.d.ts` file is updated


ajredniwja (approved):

ajredniwja:
Thank you for the contribution.

Issue #2855, opened 2019-09-18
Typescript interface definition wrong for Request onAsync events


iamatypeofwalrus:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js/issues) and didn't find any solution
  
**Describe the bug**
I'm writing an application that is putting data to CloudWatch via the `PutMetricData` API in Lambda. The code itself is written in Typescript. My payloads are large so I am trying to take advantage of gzip compression by following along with the suggestions in #2297, specifically [this comment](https://github.com/aws/aws-sdk-js/issues/2297#issuecomment-429405253).

The Typescript compiler won't allow me to to use a function with a signature like `(req, done) => {...}` because [the interface described in the typescript interface file does not include a doneCallback](https://github.com/aws/aws-sdk-js/blob/41c9d5a2ca0c455bea55c98bb99e647eed2c8781/lib/request.d.ts#L201). According to [this code async listeners are passed a `doneCallback`](https://github.com/aws/aws-sdk-js/blob/cb1604ca89a077ffdb86127884292d3b18c8b4df/lib/sequential_executor.js#L102)

**Is the issue in the browser/Node.js?**
Node.js + Typescript

**If on Node.js, are you running this on AWS Lambda?**
Yes.

**Details of the browser/Node.js version**
```
$ node --version
v10.16.0
```

**Typescript version**
```
"typescript": "^3.3.4000"
```

**SDK version number**
```
"aws-sdk": "^2.427.0",
```

**To Reproduce (observed behavior)**
n/a

**Expected behavior**
Expect Typescript interface definitions for Request.onAsync methods to have a signature that includes the `doneCallback`.

ajredniwja:
Hey @iamatypeofwalrus,

Thank-you for reaching out. I have reached to the SDK team, they will try to work on it. It will be great if you can open a [PR](https://github.com/aws/aws-sdk-js/compare), we always welcome PRs.

mhart:
[...]

PR #3511, opened 2020-10-24
Fixes Request onAsync() listener types by adding doneCallback()

jeromecovington:
This PR adds typing for the second `doneCallback()` argument for request `onAsync()` event listener callbacks. The issue is described in more detail here: https://github.com/aws/aws-sdk-js/issues/2855

##### Checklist

- [x] `npm run test` passes
- [x] `npm run tstest` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/lib/request.d.ts
+++ b/lib/request.d.ts
@@ -54,7 +54,7 @@ export class Request<D, E> {
      * @param {function} listener - Callback to run when the request is being validated.
      * @param {boolean} prepend - If set, prepends listener instead of appending.
      */
-    on(event: "validate", listener: (request: Request<D, E>) => void, prepend?: boolean): Request<D, E>;
+    on(event: "validate", listener: (request: Request<D, E>, doneCallback?: () => void) => void, prepend?: boolean): Request<D, E>;
     /**
      * Adds a listener that is triggered when the request payload is being built.
      *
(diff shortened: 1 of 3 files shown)


lib/request.d.ts
```diff
@@ -54,23 +54,23 @@ export class Request<D, E> {
      * @param {function} listener - Callback to run when the request is being validated.
      * @param {boolean} prepend - If set, prepends listener instead of appending.
      */
-    on(event: "validate", listener: (request: Request<D, E>) => void, prepend?: boolean): Request<D, E>;
+    on(event: "validate", listener: (request: Request<D, E>, doneCallback?: () => void) => void, prepend?: boolean): Request<D, E>;
```
jeromecovington, line 57: I added `doneCallback?` as an optional second param, trying to be sensitive to not breaking anybody's current code that may not use the param. Although I'm divided on whether or not it should probably actually be a required argument.

ts/request.ts
```diff
@@ -18,6 +18,10 @@ request.send(function(err, data) {
 request.on('error', function(err, response) {
 
 });
+request.onAsync('validate', function(request, done) {
+  console.log(request.httpRequest.body);
+  if (done) done()
```
jeromecovington, line 23: This is the pattern that most likely will be necessary, with `doneCallback?` annotated as optional. It might be better to make the param required so that the `if` check will not be necessary (depending on the typescript config).

jeromecovington:
The issue that I referenced this to has been closed as "stale", but from what I can tell, applying this diff would still have value. It's "companion" PR has already been merged.

PR #3514, opened 2020-10-27
Fix typings for AWSError

workeitel:
AWSError is not a class. It's based on Error but with additional properties on top:

https://github.com/aws/aws-sdk-js/blob/master/lib/util.js#L570-L603

By wrongly declaring a class statements like `if (error instanceof AWSError)` would compile but during runtime `AWSError` is `undefined` since it's just a type. Fixing the types prevents this issue. In addition many properties are optional. Declaring them as required can cause type errors and makes mocking more difficult.

##### Checklist

- [X] `npm run test` passes
- [X] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/bugfix-Types-3007af51.json
+++ b/.changes/next-release/bugfix-Types-3007af51.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Types",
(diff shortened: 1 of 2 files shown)


ajredniwja (approved):

ajredniwja:
Thanks for your contribution.

PR #3515, opened 2020-10-27
Update NodeJS versions in travis build

workeitel:
Everything older than NodeJS 10 is deprecated according to https://nodejs.org/en/about/releases/

In addition I added 12 and 14 as the current active LTS versions.

The travis build takes ages. I hope by dropping the ancient node versions it could make it faster.

##### Checklist

- [ ] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
-  - "4"
-  - "6"
   - "8"
   - "10"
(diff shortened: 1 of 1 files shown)


ajredniwja:
Will be done as part of https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-node-js-runtimes-10-x-in-the-aws-sdk-for-javascript-v2/

Issue #3504, opened 2020-10-19
Timeout documentation appears to be incorrect


julrich1:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js/issues) and didn't find any solution
  
**Describe the bug**
In the export of the HTTPOptions interface there are several comments regarding timeouts along with comments describing their functionality. Along with these the SDK documentation also has comments that do not match the TypeScript comments.
In my testing, neither of these is correct. The timeout is the actual transaction timeout when waiting for a response from Dynamo, nothing related to inactivity of the connection. When tuning timeouts and retries it is very important for this functionality to be well documented, otherwise the default value of two minutes for a timeout can cause extremely long transaction times.

TypeScript comments:
`    /**
     * The number of milliseconds to wait before giving up on a connection attempt.
     * Defaults to two minutes (120000).
     */
    timeout?: number;`

SDK comments (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB.html):
`timeout [Integer] — Sets the socket to timeout after timeout milliseconds of inactivity on the socket. Defaults to two minutes (120000).`

**Is the issue in the browser/Node.js?**
Node.js

**If on Node.js, are you running this on AWS Lambda?**
No

**Details of the browser/Node.js version**
v12.14.1

**SDK version number**
2.769.0

**Expected behavior**
I would expect proper documentation on this functionality so that tuning retries and timeouts behaves as expected.

ajredniwja:
Hey @julrich1 Thanks for opening this issue, 

My findings: 

<details>

<summary>
Setting timeout lower than the time it needs to get a response from the service i.e 2ms:
</summary>

I get error: 

```
  message: 'Connection timed out after 2ms',
  code: 'TimeoutError',
  time: 2020-10-30T22:54:33.282Z,
  region: 'us-west-2',
  hostname: 'dynamodb.us-west-2.amazonaws.com',
  retryable: true
```

</details>


<details>

[...]

PR #3516, opened 2020-10-30
Documentation changes for timeout in HTTP options

ajredniwja:
Related to #3504 A part of it was to fix the comments in the code which is not generated, next step would be to request changes in the API reference. 

##### Checklist
- [ x ] non-code related change (markdown/git settings etc)

### DIFF
--- a/lib/config-base.d.ts
+++ b/lib/config-base.d.ts
@@ -62,7 +62,7 @@ export interface HTTPOptions {
      */
     connectTimeout?: number;
     /**
-     * The number of milliseconds to wait before giving up on a connection attempt.
+     * The number of milliseconds to wait before giving up on a requested response from a service.
      * Defaults to two minutes (120000).
      */
     timeout?: number;
--- a/lib/config-base.d.ts
+++ b/lib/config-base.d.ts
@@ -62,7 +62,7 @@ export interface HTTPOptions {
      */
     connectTimeout?: number;
     /**
-     * The number of milliseconds to wait before giving up on a requested response from a service.
+     * The number of milliseconds a request can take before automatically being terminated.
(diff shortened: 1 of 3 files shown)


lib/config-base.d.ts
```diff
@@ -62,7 +62,7 @@ export interface HTTPOptions {
      */
     connectTimeout?: number;
     /**
-     * The number of milliseconds to wait before giving up on a connection attempt.
+     * The number of milliseconds to wait before giving up on a requested response from a service.
```
trivikr, line 65: Use the [XMLHttpRequest.timeout terminology from MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout).
```suggestion
     * The number of milliseconds a request can take before automatically being terminated.
```

Code which shows where httpOptions.timeout is used:
* Set in XMLHttpRequest.timeout in: https://github.com/aws/aws-sdk-js/blob/c62b6ed76c2f9676247e165bd3ed66daebec8f1c/lib/http/xhr.js#L70-L72
* Set in request.setTimeout in: https://github.com/aws/aws-sdk-js/blob/c62b6ed76c2f9676247e165bd3ed66daebec8f1c/lib/http/node.js#L82

Node.js docs:
* [request.setTimeout(timeout[, callback])](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback)
* [socket.setTimeout(timeout[, callback])](https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback)

trivikr (approved):

trivikr (changes requested):
Build failure (trailing spaces) needs to be fixed.

trivikr (approved):

PR #3537, opened 2020-11-19
Fix createPresignedPost dual-callback error

### LINKED ISSUE #3486: createPresignedPost triggers the callback twice and crashes when credentials are not provided
  
**Describe the bug**
When calling `createPresignedPost` with a callback and invalid credentials, my callback is called and node also crashes.

Node.js

No

v12.18.3

**SDK version number**
v2.763.0

In an environment *with no configured credentials*:

```
import { S3 } from 'aws-sdk';
try {
  const s3 = new S3();
  s3.createPresignedPost({}, (err, data) => {
    console.log('sup dog');
  });
} catch (err: Error) {
  console.log('KABOOM!');
}
```

Running this causes the following output:

```
sup dog
./node_modules/aws-sdk/lib/services/s3.js:1241
      throw new Error('Unable to create a POST object policy without a bucket,'
      ^

Error: Unable to create a POST object policy without a bucket, region, and credentials
    at features.constructor.preparePostFields (./node_modules/aws-sdk/lib/services/s3.js:1241:13)
    at finalizePost (./node_modules/aws-sdk/lib/services/s3.js:1204:22)
    at ./node_modules/aws-sdk/lib/services/s3.js:1221:24
    at finish (./node_modules/aws-sdk/lib/config.js:386:7)
    at ./node_modules/aws-sdk/lib/config.js:428:9
    at Object.<anonymous> (./node_modules/aws-sdk/lib/credentials/credential_provider_chain.js:111:13)
    at Object.arrayEach (./node_modules/aws-sdk/lib/util.js:516:32)
    at resolveNext (./node_modules/aws-sdk/lib/credentials/credential_provider_chain.js:110:20)
    at ./node_modules/aws-sdk/lib/credentials/credential_provider_chain.js:126:13
    at ./node_modules/aws-sdk/lib/credentials.js:124:23
error Command failed with exit code 1.
```

I looked into the code and what's happening is that my callback is being called twice.  Specifically, it is being called [once to pass the credential error](https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/services/s3.js#L1081), and then it is being [called a second time](https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/services/s3.js#L1084).

The second call does not actually complete, because `finalizePost` calls `preparePostFields` which then [throws an error](https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/services/s3.js#L1104) due to invalid configuration. 

**Expected behavior**
I would expect the error to be passed to the callback directly (using `err`) without terminating the process.

**Screenshots**
N/A

**Additional context**
The partial solution is to update [L1080](https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/services/s3.js#L1081) to be `return callback(err);` instead of just `callback(err);`

The fully defensive solution would be to ALSO wrap [L1084](https://github.com/aws/aws-sdk-js/blob/4fd99d087f245f23b164219fba62156af060770e/lib/services/s3.js#L1084) in a try catch block:

```
try {
        callback(null, finalizePost());
} catch (err) {
        callback(err);
}
```

slifty:
This fixes a bug where using the callback version of createPresignedPost without proper AWS credentials configured would result in an un-catchable error and crash the node process.

Note: as discussed in #3486, I was not able to figure out a way to write a test to capture this error due to the nature of the error.  In the original code, the callback is called correctly at first, and only after that does the crash occur (when the callback is called a second time).

Resolves #3486

##### Checklist

- [x] `npm run test` passes

### DIFF
--- a/.changes/next-release/bugfix-S3-9d19dd09.json
+++ b/.changes/next-release/bugfix-S3-9d19dd09.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "S3",
+  "description": "fixed a bug where createPresignedPost could result in a process crash."
+}
--- a/lib/services/s3.js
+++ b/lib/services/s3.js
@@ -1079,9 +1079,13 @@ AWS.util.update(AWS.S3.prototype, {
       config.getCredentials(function (err) {
         if (err) {
           callback(err);
+        } else {
+          try {
+            callback(null, finalizePost());
+          } catch (err) {
+            callback(err);
+          }
         }
-
-        callback(null, finalizePost());
       });
     } else {
       return finalizePost();


ajredniwja:
Can you please address the build failure @slifty ?

slifty:
@ajredniwja I believe that failure is because I force pushed to correct a typo in my change log that I noticed right after opening the PR, which meant the branch didn't exist any more (the error in that failure is `
reference not found for primary source and source version 4b64673d0e391567ab0e3a37f7bc5a437a06378e`)

The `fd99318` commit passed -- which is indeed the actual commit in the current PR.

Is there a way for me to trigger a re-build, or is that good enough?

ajredniwja:
Hey @slifty we would need a successful build to review and merge this what you can do is rebase from master and push which would trigger rebuild. 

```
git fetch upstream
git rebase upstream/master
git push --force-with-lease
```

slifty:
@ajredniwja Just following up on whether there are any other changes you would like?

ajredniwja (approved):

ajredniwja:
Apologies for late reply, Thanks for your late reply.

PR #3545, opened 2020-11-22
Allow core-only loading (no service classes loading)

### LINKED ISSUE #3544: Endpoint throws TypeError if imported from aws-sdk/lib/core
  
**Describe the bug**
Class `Endpoint` throws `TypeError` if imported from `aws-sdk/lib/core`.

```
Uncaught TypeError: Cannot read property 'parse' of undefined
    at Object.urlParse (/Users/cw-tatsuno/IdeaProjects/aws-sdk-scalajs-facade/node_modules/aws-sdk/lib/util.js:60:21)
    at new Endpoint (/Users/cw-tatsuno/IdeaProjects/aws-sdk-scalajs-facade/node_modules/aws-sdk/lib/http.js:58:36)
```

Node.js

YES

v14.15.1

**SDK version number**
v2.797.0

Run following script where `aws-sdk` installed.

```js
const core = require("aws-sdk/lib/core")
const customEndpoint = new core.Endpoint("http://localhost");
```

**Expected behavior**

Class `Endpoint` can be constructed as similar as when invoked as `AWS.Endpoint`.
I want to import `Endpoint` from `aws-sdk/lib/core` to minimize bundle size by tree shaking.

exoego:
This PR closes #3544 .
Users will be able to load "core" modules with no service classes.
This would be helpful for minimizing bundle by tree shaking.

```js
const coreOnlyAWS = require('aws-sdk/aws_core_loader');
const endpoint = new coreOnlyAWS.Endpoint('http://localhost');
```

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/lib/aws.js
+++ b/lib/aws.js
@@ -1,6 +1,5 @@
-require('./node_loader');
-
-var AWS = require('./core');
+// Load core-only
+var AWS = require('./aws_core_loader');
 
 // Load all service classes
 require('../clients/all');
--- a/lib/aws_core_loader.js
+++ b/lib/aws_core_loader.js
@@ -1,0 +1,8 @@
+require('./node_loader');
+
+var AWS = require('./core');
+
+/**
+ * @api private
+ */
+module.exports = AWS;
--- a/test/core.spec.js
+++ b/test/core.spec.js
@@ -1,0 +1,19 @@
+(function() {
+  var AWS;
+
+  if (typeof window === 'undefined') {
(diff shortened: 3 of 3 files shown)


exoego:
This PR is not necessary.
The below works.

```js
const coreOnlyAWS = require('aws-sdk/lib/node_loader');
const endpoint = new coreOnlyAWS.Endpoint('http://localhost');
```

Issue #3550, opened 2020-11-25
ie11 upload image larger than 20m error


hkdg:
version：2.725.0

**`ie11` upload image larger than 20m error**
The error report is as follows

```
"No access to ETag property on response. Check CORS configuration to expose ETag header." 
```
![image](https://s3.ax1x.com/2020/11/26/DwhSgO.png)

**Problem orientation: ` rawHeaders.split (/ \ r? \ n /) ` problems caused by incompatibility of regular expressions**

```javascript
// lib/http/xhr.js  source code
parseHeaders: function parseHeaders(rawHeaders) {
    var headers = {};
    AWS.util.arrayEach(rawHeaders.split(/\r?\n/), function (line) {
      var key = line.split(':', 1)[0];
      var value = line.substring(key.length + 2);
      if (key.length > 0) headers[key.toLowerCase()] = value;
    });
    console.log(rawHeaders)
    console.log(headers)
    return headers;
  },
```
Rawheaders before `ie11` printing and parsing

```javascript
access-control-allow-credentials: true
access-control-allow-origin: *
access-control-expose-headers: Date, Etag, Server, Connection, Accept-Ranges, Content-Range, Content-Encoding, Content-Length, Content-Type, X-Amz*, X-Amz*, *
content-disposition: attachment;filename=
content-encoding: gzip
content-security-policy: block-all-mixed-content
content-type: application/xml
date: Wed, 25 Nov 2020 07:57:06 GMT
etag: W/"7518960c8d4a92075f3d969d011ad045-5"
status: 200
strict-transport-security: max-age=15724800; includeSubDomains
vary: Accept-Encoding, Origin
x-amz-request-id: 164AB12DAF04D1D2
x-xss-protection: 1; mode=block
```
`ie11` print the parsed headers, the analysis is confused
![image](https://s3.ax1x.com/2020/11/26/DwhQbj.png)

The header method after `chrome` printing and parsing is correct

![image](https://s3.ax1x.com/2020/11/26/Dwhm28.png)

**Solution**
Replace it ` rawHeaders.split (/ \ r? \ n /)  `, use the finite state machine to parse the response header as follows

```javascript
// lib/http/xhr.js  source code
parseHeaders: function parseHeaders(rawHeaders) {
  // rawHeaders.split(/\r?\n/) method has a bug in ie11
// Using finite state machine to solve ie11 parsing head problem
var responseChunkedParse = {
  receive: function(data) {
    //  the response header
    this.header = {}
[...]

PR #3551, opened 2020-11-26
🐛 solve ie11 parsing head problem

hkdg:


hkdg:
[ie11 upload image larger than 20m error](https://github.com/aws/aws-sdk-js/issues/3550)

PR #3597, opened 2021-01-11
Add Honeycode client to list of clients.

ncb000gt:
Honeycode is a supported service, but hasn't yet been added to the default
list of available clients. This commit fixes that.

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated


ncb000gt:
Nevermind, it is there in the clients list. Just not alphabetized- must have had a typo when I searched for it.

PR #3492, opened 2020-10-12
Fix: do not require credentials when using config file

thomasmichaelwallace:
Resolves #3436 

##### Checklist

- [x] `npm run test` passes
- [ ] ~`.d.ts` file is updated~
- [x] changelog is added, `npm run add-change`
- [ ] ~run `bundle exec rake docs:api` and inspect `doc/latest/index.html` if documentation is changed~
- [ ] ~run `npm run integration` if integration test is changed~
- [ ] ~non-code related change (markdown/git settings etc)~

PR #3684, opened 2021-03-28
fix: Missing credentials file with config file

JadenSimon:
When the environment variable `AWS_SDK_LOAD_CONFIG` is set to a truthy value, the expected behavior is to suppress errors caused by a missing credentials file, as introduced in #3492. The SDK's behavior does not currently match when it tries to read a region from a profile.

The bug fix instead treats a missing credentials file as empty rather than raising an exception. A test was added to check that the fix correctly implements the expected behavior.

##### Checklist

- [x] `npm run test` passes (all but 2 that were already failing)

### DIFF
--- a/.changes/next-release/bugfix-Credentials-c59044aa.json
+++ b/.changes/next-release/bugfix-Credentials-c59044aa.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Credentials",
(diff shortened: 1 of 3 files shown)


AllanZhengYP (approved):
It looks good to me. :shipit:

AllanZhengYP (approved):

PR #3690, opened 2021-03-31
fix: move next-release json file to correct location

trivikr:
Summary: The release on 3/31 failed as release json was inside `bugfix-S3` folder instead of being directly in `next-release` folder. More details in JS-2530 (internal)

##### Checklist


alexforsyth (approved):
+1

PR #3709, opened 2021-04-13
Extend package entry points for Node.js esm support

adrai:
This change would help to import the aws-sdk natively in Node.js like this:

```javascript
// import entire SDK
import AWS from 'aws-sdk';
// import AWS object without services
import AWS from 'aws-sdk/global';
// import individual service
import S3 from 'aws-sdk/clients/s3';
```

Like described in the README.md for TypeScript: https://github.com/aws/aws-sdk-js#in-nodejs-1

Node.js doc: https://nodejs.org/api/packages.html#packages_package_entry_points

##### Checklist

- [x] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,11 @@
     "./lib/core.js": "./dist/aws-sdk-core-react-native.js",
     "xml2js": "./dist/xml2js.js"
   },
+  "exports": {
+    ".": "./lib/aws.js",
+    "./*": "./*.js",
+    "./clients/*": "./clients/*.js"
+  },
   "directories": {
     "lib": "lib"
   },


adrai:
@AllanZhengYP @trivikr @alexforsyth what's your opinion to this?

trivikr:
> what's your opinion to this?

The AWS SDK for JavaScript (v2) is going to be put on deprecation path at some point, as AWS SDK for JavaScript (v3) is generally available and supports ESM. We would advise customers to switch to v3 instead.

adrai:
> > what's your opinion to this?
> 
> The AWS SDK for JavaScript (v2) is going to be put on deprecation path at some point, as AWS SDK for JavaScript (v3) is generally available and supports ESM. We would advise customers to switch to v3 instead.

I know, but in the meantime (while v3 gets more stable) what prevents us from adding those exports in the package.json?

PR #3739, opened 2021-05-02
Retry http request on ECONNRESET error

workeitel:
> ECONNRESET (Connection reset by peer): A connection was forcibly
closed by a peer. This normally results from a loss of the connection on
the remote socket due to a timeout or reboot. Commonly encountered via
the http and net modules.

This error can for example happen when fetching Fargate credentials from
metadata endpoint.

##### Checklist

- [x] `npm run test` passes

### DIFF
--- a/.changes/next-release/bugfix-retry-c648c8ae.json
+++ b/.changes/next-release/bugfix-retry-c648c8ae.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "retry",
+  "description": "Retry http request on ECONNRESET error"
+}
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -93,7 +93,7 @@ AWS.NodeHttpClient = AWS.util.inherit({
         connectTimeoutId = null;
       }
       if (stream.didCallback) return; stream.didCallback = true;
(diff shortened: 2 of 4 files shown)


trivikr (changes requested):
The error [`ECONNRESET`](https://nodejs.org/api/errors.html#errors_common_system_errors) is specific to Node.js, and it should be moved to Node.js http handler: https://github.com/aws/aws-sdk-js/blob/d3ce713151b6d1085a9e19e0a79181aeda390c85/lib/http/node.js#L90-L97

There are other timeout errors with code `EPIPE` and `ETIMEDOUT` which also need to be handled.

Example diff:
```diff
       stream.abort();
     });
 
-    stream.on('error', function() {
+    stream.on('error', function(err) {
       if (connectTimeoutId) {
         clearTimeout(connectTimeoutId);
         connectTimeoutId = null;
       }
       if (stream.didCallback) return; stream.didCallback = true;
-      errCallback.apply(stream, arguments);
+      if (['ECONNRESET', 'EPIPE', 'ETIMEDOUT'].includes(err.code)) {
+        errCallback(AWS.util.error(err, {code: 'TimeoutError'}));
+      } else {
+        errCallback(err);
+      }
     });
 
     var expect = httpRequest.headers.Expect || httpRequest.headers.expect;
```

trivikr (approved):

PR #3785, opened 2021-05-31
Update uuid dependency to 8.3.2

TikiTDO:
Resolves deprecation warning from uuid module:

```
warning aws-sdk > uuid@3.3.2: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

The only usage of this module is [here](https://github.com/aws/aws-sdk-js/blob/e26ac616265a6f59913248c060f02146cd126d10/lib/util.js#L925).

As per the [uuid docs](https://github.com/uuidjs/uuid#upgrading-from-uuid3) there is no need for any code changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "querystring": "0.2.0",
     "sax": "1.2.1",
     "url": "0.10.3",
-    "uuid": "3.3.2",
+    "uuid": "8.3.2",
     "xml2js": "0.4.19"
   },
   "main": "lib/aws.js",


TikiTDO:
Currently the `uuid` module is failing on older versions of node. I have submitted a PR for that repo [here](https://github.com/uuidjs/uuid/pull/574) which should make newer versions of uuid work with older versions of node.

ajredniwja:
Can you address the build failures?

TikiTDO:
Ah sorry, this is not going to work until the AWS SDK drops support for older versions of node in November. The newer version of the `uuid` module no longer support any version of node before 10, and there is not appetite on the side of the `uuid` developers to merge in any fixes for deprecated code versions. 

Until this version conflict is resolved the only thing I can recommend is to set the `resolutions` block in `package.json` to an up-to-date version of the `uuid` lib. The API is fully compliant, as long as it's running in an sufficiently recent version of node.

PR #3805, opened 2021-06-22
docs: fix invokeAsync description

joscha:
##### Checklist

- [x] `.d.ts` file is updated

### DIFF
--- a/clients/lambda.d.ts
+++ b/clients/lambda.d.ts
@@ -254,11 +254,11 @@ declare class Lambda extends Service {
    */
   invoke(callback?: (err: AWSError, data: Lambda.Types.InvocationResponse) => void): Request<Lambda.Types.InvocationResponse, AWSError>;
   /**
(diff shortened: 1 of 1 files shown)


clients/lambda.d.ts
```diff
@@ -254,11 +254,11 @@ declare class Lambda extends Service {
    */
   invoke(callback?: (err: AWSError, data: Lambda.Types.InvocationResponse) => void): Request<Lambda.Types.InvocationResponse, AWSError>;
   /**
-   *  For asynchronous function invocation, use Invoke.  Invokes a function asynchronously.
```
joscha, line 257: Possible other improvements:

- [ ] Remove double space `  ` between the sentences
- [ ] Swap the order, so we describe first what this function does and then the synchronous alternative, which would make more sense semantically.

ajredniwja (approved):

PR #3826, opened 2021-07-09
Fix CloudFront signer RFC5987 support

### LINKED ISSUE #2952: Cloudfront Signer not works with UTF-8 filenames in content-disposition
  
**Describe the bug**
I want to sign url by Cloudfront.Signer but filename could contain special or diacritic characters. 
On S3 file is stored on random name, but I want to download it in proper name so need to use "response-content-dispostion" query parameter.
I tried to sign it using this code:
```js
function urlEncode(str) {
        return encodeURIComponent(str)
            .replace(/!/g, '%21')
            .replace(/'/g, '%27')
            .replace(/\(/g, '%28')
            .replace(/\)/g, '%29')
            .replace(/\*/g, '%2A')
            .replace(/~/g, '%7E');
    }
const filename = 'abc def.mp4'; // for simple example, not works for this as well
const s = new AWS.CloudFront.Signer(....);
const signedUrl = s.getSignedUrl({
                expires: ~~(Date.now() / 1000) + expirationSeconds,
                url: `https://my-url.com/folder/file.mp4?response-content-disposition=${this.urlEncode(`attachment;filename*=UTF-8''${urlEncode(filename)}`)}`,
            });
```
In result, i get signed url with not works - cloudfront return AccessDenied message.
```
https://my-url.com/folder/file.mp4?response-content-disposition=attachment%3Bfilename*%3DUTF-8''abc%2520def.mp4&Expires=1573653786&Key-Pair-Id=....&Signature=....
```
There is not encoded `*` `'` (and maybe more?) chars.
In PHP version of sdk everything works.
```php
$client = new Aws\CloudFront\CloudFrontClient([
    'region' => 'eu-central-1',
    "version" => "2019-03-26"
]);

$resourceKey = 'https://my-url.com/folder/file.mp4?response-content-disposition='.urlencode('attachment;filename*=UTF-8\'\''.rawurlencode('abc def.mp4'));
$expires = time() + 300;

// Create a signed URL for the resource using the canned policy
echo($client->getSignedUrl([
    'url' => $resourceKey,
    'expires' => $expires,
    'private_key' => 'key.pem',
    'key_pair_id' => '...'
])."\n");
```
returns:
```
https://my-url.com/folder/filemp4?response-content-disposition=attachment%3Bfilename%2A%3DUTF-8%27%27abc%2520def.mp4&Expires=1573643427&Signature=...&Key-Pair-Id=...
```

Node.js

No

12.12.0

**SDK version number**
2.548.0

see above

**Expected behavior**
SDK should return proper encoded url.

**Screenshots**

**Additional context**
Workaroud which works is additional replace (to revert original response-content-disposition) after generating signed url:
```js
const s = new AWS.CloudFront.Signer(...);

        const preparedUrl = `https://my-url.com/folder/file.mp4?response-content-disposition=${this.urlEncode(`attachment;filename*=UTF-8''${urlEncode(filename)}`)}`
        const signedUrl = s.getSignedUrl({
                expires: ~~(Date.now() / 1000) + expirationSeconds,
                url: preparedUrl,
            }).replace(/\?response-content-disposition=(.*?)&/, (matched) => {
                return '?response-content-disposition=' + (preparedUrl.match(/\?response-content-disposition=(.*?)(?:&|$)/) || [matched]).pop() + '&';
            });
```

realies:
`url.parse(someUrl, true)` in combination with nulling its `search` property has a side-effect when re-encoding using `url.format`.
https://github.com/aws/aws-sdk-js/blob/821e9a11f3917551e26d5cbc794bf319979ceb00/lib/cloudfront/signer.js#L183
https://github.com/aws/aws-sdk-js/blob/821e9a11f3917551e26d5cbc794bf319979ceb00/lib/cloudfront/signer.js#L188

It causes URL queries existing before the signing process to get re-encoded improperly during the `url.format` stage here:
https://github.com/aws/aws-sdk-js/blob/821e9a11f3917551e26d5cbc794bf319979ceb00/lib/cloudfront/signer.js#L197-L198

This breaks the usage of some special characters in URL queries, even when they have been properly URL encoded. One example is setting the filename in a specific charset using the `response-content-disposition` query field. For example, the `response-content-disposition`  query value ([RFC5987 examples](https://datatracker.ietf.org/doc/html/rfc5987#section-3.2.2)) may contain:
```
attachment%3Bfilename%2A%3DUTF-8%27%27abc%2520def.mp4
```
Although the current functionality returns a signed URL that has the query value formatted as:
```
attachment%3Bfilename*%3DUTF-8''abc%2520def.mp4
```

This invalidates the generated signature and CloudFront replies with:
```
<Error>
<Code>AccessDenied</Code>
<Message>Access denied</Message>
</Error>
```

To replicate the issue in isolation:
```
> var u = url.parse('https://asdf.cloudfront.net/932e9e5d-cbdf-4c73-a3dc-e07758bd3adb?response-content-disposition=attachment%3Bfilename%2A%3DUTF-8%27%27abc%2520def.mp4', true)
> url.format(u)
'https://asdf.cloudfront.net/932e9e5d-cbdf-4c73-a3dc-e07758bd3adb?response-content-disposition=attachment%3Bfilename%2A%3DUTF-8%27%27abc%2520def.mp4'
> u.search=null
> url.format(u)
"https://asdf.cloudfront.net/932e9e5d-cbdf-4c73-a3dc-e07758bd3adb?response-content-disposition=attachment%3Bfilename*%3DUTF-8''abc%2520def.mp4"

```
The example input URL query works fine in `aws-sdk-php` and `boto3`.

Fixes #2952.

### DIFF
--- a/lib/cloudfront/signer.js
+++ b/lib/cloudfront/signer.js
@@ -180,22 +180,27 @@ AWS.CloudFront.Signer = inherit({
             return handleError(err, cb);
         }
 
-        var parsedUrl = url.parse(options.url, true),
-            signatureHash = Object.prototype.hasOwnProperty.call(options, 'policy')
-                ? signWithCustomPolicy(options.policy, this.keyPairId, this.privateKey)
-                : signWithCannedPolicy(resource, options.expires, this.keyPairId, this.privateKey);
+        var signatureHash = Object.prototype.hasOwnProperty.call(options, 'policy')
+            ? signWithCustomPolicy(options.policy, this.keyPairId, this.privateKey)
+            : signWithCannedPolicy(resource, options.expires, this.keyPairId, this.privateKey);
 
-        parsedUrl.search = null;
-        for (var key in signatureHash) {
-            if (Object.prototype.hasOwnProperty.call(signatureHash, key)) {
-                parsedUrl.query[key] = signatureHash[key];
-            }
+        var signatureHashUrlQuery = Object.keys(signatureHash).map(function (key) {
+            return key + '=' + signatureHash[key];
+        }).join('&');
+        var urlHasQuery = resource.split('/').pop().includes('?');
+
+        var formattedUrl = resource;
+        if (urlHasQuery) {
+            formattedUrl += '&';
+        } else {
+            formattedUrl += '?';
         }
+        formattedUrl += signatureHashUrlQuery;
 
         try {
             var signedUrl = determineScheme(options.url) === 'rtmp'
-                    ? getRtmpUrl(url.format(parsedUrl))
-                    : url.format(parsedUrl);
+                    ? getRtmpUrl(formattedUrl)
+                    : formattedUrl;
         } catch (err) {
             return handleError(err, cb);
         }


realies:
@ajredniwja @AllanZhengYP could you please review this and potentially release it?

realies:
@ajredniwja @AllanZhengYP this is more than a year old now

realies:
review it

kellertk:
Thank you for your efforts, but this project is no longer accepting contributions as part of our plan to end support for AWS SDK for JavaScript v2. This method is overridable in your own code at runtime, if you prefer.

PR #3861, opened 2021-08-22
Include port number on ENOTFOUND_ERROR error message

dav1app:
This pull request includes the port number for the errors returned by `ENOTFOUND_ERROR`. 

This is important because sometimes there are multiple services running in the `hostname` but without the port number it is impossible to identify which one is not being found. 

##### Checklist

None of the items apply, since it is a simple message that is returned by `AWS.util.error()`


ajredniwja (approved):

ajredniwja:
Thanks for your contribution.

PR #3904, opened 2021-09-30
ManagedUpload.send's callback may return null instead of an Error

jaller94:
This callback's TypeScript definition says it would always return an AWSError, however, by looking at the actual code and JSDocs one notices that the `err` may also be `null`.

https://github.com/aws/aws-sdk-js/blob/a42ad578686801607b88358b7a85da48a8952bdc/lib/s3/managed_upload.js#L159

I don't intent to maintain this PR – I just though a code change would be more descriptive than filing an issue. Please edit this or commit it yourself if you must not have contributors without CLA signature in your code.

##### Checklist

- [ ] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/lib/s3/managed_upload.d.ts
+++ b/lib/s3/managed_upload.d.ts
@@ -16,7 +16,7 @@ export class ManagedUpload {
     /**
      * Initiates the managed upload for the payload.
      */
-    send(callback?: (err: AWSError, data: ManagedUpload.SendData) => void): void;
+    send(callback?: (err: AWSError|null, data: ManagedUpload.SendData) => void): void;
     /**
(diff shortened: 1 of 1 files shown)


jaller94:
This type error still exists.

Based on recent reviews, @kuhe, could you have a look at this please?

jaller94:
This is still an issue and even the TypeScript types now have this incorrect type assumption.

https://github.com/aws/aws-sdk-js/blob/49f5d76b417615de1f928d6cfdf9af60e6cc9ca0/lib/s3/managed_upload.d.ts#L19

kuhe (approved):

PR #3957, opened 2021-11-11
Add useDualstackEndpoint configuration

trivikr:
This PR:
* Add useDualStack configuration.
* Reads from env/shared ini if it's not passed in client configuration.
* Enables dualstack, if legacy `useDualstack` is passed for S3/S3Control.
* Adds endpoint tests for useDualStackEndpoint.

##### Checklist

- [x] `npm run test` passes


AllanZhengYP (approved):
Solid PR, 🚢 it!

PR #3996, opened 2021-12-25
Make maxSockets predictable

Kikobeats:
Not sure why this value is hard-coded to be 50, but it isn't a predictable thing.

I was facing with hidden infrastructure issue because of this, and I didn't be possible to find the root cause of problems until I randomly read a blog post related to that:

https://plaid.com/blog/how-we-parallelized-our-node-service-by-30x

I'll happy to reflect the changes on the testing suite if I have ACK of this will be merge 🙂


Kikobeats:
Closing because v3 design is good
https://github.com/aws/aws-sdk-js-v3/blob/b0e1422630/packages/node-http-handler/src/node-http-handler.ts#L45

PR #3997, opened 2021-12-26
Fix rest-json test case, which was not being run

danopia:
I have been running the protocol fixtures against a different codebase, and one of the newer fixtures doesn't seem to pass properly there. I determined that protocol fixture cases without a `body` field weren't being fully evaluated in this project. So I first fix the test helper to properly run the bad case, and then fix the case. The first commit "should" be red if CI runs on it.

##### Checklist
- [ ] `npm run test` passes


danopia:
I suppose y'all are ok with the test assertion not working 🤷🏻

PR #4032, opened 2022-02-18
Use fetch instead of XMLHttpRequest

yiukamsum:
This is my frist PR in GitHub. I don't know the rul of PR. Please help me if I am doing wrong things.
And this code just copy from issuse #2807. Thanks @zzzxtreme.

Related issues: #2807 #2819

##### Checklist

- [ ] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/lib/http/xhr.js
+++ b/lib/http/xhr.js
@@ -6,82 +6,85 @@ require('../http');
  * @api private
  */
 AWS.XHRClient = AWS.util.inherit({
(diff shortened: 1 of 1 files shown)


kuhe:
Appreciate the contribution, however we do not need to migrate to fetch in JSv2. The AWS SDK for JavaScript v3 does use `fetch` by default in browsers.

PR #4055, opened 2022-03-18
fix: update redirect logic to also update path

CaerusKaru:
When a 307 redirect comes in, it contains the new, full URL for the location change.
The current logic respects this for the `Host` header that gets sent, but it does not
then update the path, leading to possibly invalid redirects. The new logic pulls the
path out of the endpoint and updates the outgoing request.

##### Checklist

- [x] `npm run test` passes

Ref: Ticket ID V544628989

### DIFF
--- a/.changes/next-release/bugfix-event-listeners-536ce342.json
+++ b/.changes/next-release/bugfix-event-listeners-536ce342.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "event listeners",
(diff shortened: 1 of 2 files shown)


AllanZhengYP (commented):
Thanks for the fix. It makes sense to me. Can you add a changelog so we can release it?

CaerusKaru:
Please do not close. Internal issue that should be addressed.

kuhe (approved):

PR #4060, opened 2022-03-25
chore: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.
##### Checklist

- [x] `npm run test` passes
- [ ] `.d.ts` file is updated

### DIFF
--- a/lib/event-stream/parse-message.js
+++ b/lib/event-stream/parse-message.js
@@ -101,11 +101,11 @@ function parseHeaders(headers) {
                 position += 16;
                 out[name] = {
                     type: UUID_TAG,
-                    value: uuidChars.substr(0, 8) + '-' +
-                        uuidChars.substr(8, 4) + '-' +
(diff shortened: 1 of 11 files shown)


elizatlawy (approved):

CommanderRoot:
The PR is still relevant

kellertk:
Thank you for your efforts, but this project is no longer accepting contributions as part of our plan to end support for AWS SDK for JavaScript v2.

PR #4125, opened 2022-06-13
fix(util): update ini parsing

kuhe:
Update the parsing method for ini file. This does not change the expected behavior. Covered by existing unit tests. 

- [x] `npm run test` passes
- n/a `.d.ts` file is updated
- n/a changelog is added, `npm run add-change`
- n/a run `npm run integration` if integration test is changed

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -225,10 +225,15 @@ var util = {
             );
           }
         } else if (currentSection) {
-          var item = line.match(/^\s*(.+?)\s*=\s*(.+?)\s*$/);
-          if (item) {
+          var expression = line.split('=');
+          var isAssignment = expression.length >= 2 && !!expression[0];
+
+          if (isAssignment) {
+            var name = expression[0].trim();
+            var value = expression.slice(1).join('=').trim();
+
             map[currentSection] = map[currentSection] || {};
-            map[currentSection][item[1]] = item[2];
+            map[currentSection][name] = value;
           }
         }
       });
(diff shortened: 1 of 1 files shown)


kuhe:
related https://github.com/aws/aws-sdk-js-v3/pull/3682

lib/util.js
```diff
@@ -225,10 +225,15 @@ var util = {
...
           }
         } else if (currentSection) {
-          var item = line.match(/^\s*(.+?)\s*=\s*(.+?)\s*$/);
-          if (item) {
+          var expression = line.split('=');
+          var isAssignment = expression.length >= 2 && !!expression[0];
+
+          if (isAssignment) {
+            var name = expression[0].trim();
+            var value = expression.slice(1).join('=').trim();
```
trivikr, line 233: Use [indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) and [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) instead of split?

```suggestion
          var indexOfEqualsSign = line.indexOf("=");

          if (indexOfEqualsSign !== -1) {
            var name = line.substring(0, indexOfEqualsSign).trim();
            var value = line.substring(indexOfEqualsSign + 1).trim();
```

trivikr (changes requested):

trivikr (approved):

PR #3914, opened 2021-10-07
feat: update serialization of Rest-JSON API input and Content-Type

trivikr:
### Issue

Internal JS-2854

### Description
* Rest-JSON API input now only supports Content-Type `application/json` for all services except finspace.
  * ~~The customization for finspace will be removed once the changes are done on the service side.~~ Removed.
* The Content-Type will be sent for all REST-JSON requests sending structured content in the body.
* For operations modeled to send a JSON payload, if the requests are empty due to all arguments being optional, SDK sends `{}` in the body.

### Testing

* Verified that all protocol tests are successful. The protocol tests now remove the hacks for rest-json, and checks for body and headers only if present in output as discussed internally.
* Verified that paginator keys are not ignored after removing https://github.com/aws/aws-sdk-js/pull/2562 and subsequent changes. The revert of https://github.com/aws/aws-sdk-js/pull/2589 does not require testing, as the code path will not reach after previous revert.

<details>
<summary>Paginator keys are not ignored</summary>

```console
$ const client = new AWS.PinpointEmail({ region: "us-west-2" });

$ client.listEmailIdentities().promise().then(console.log)
aws-sdk> {
  EmailIdentities: [
    {
      IdentityType: 'EMAIL_ADDRESS',
      IdentityName: 'bar@example.com',
      SendingEnabled: false
    },
    {
      IdentityType: 'EMAIL_ADDRESS',
      IdentityName: 'foo@example.com',
      SendingEnabled: false
    },
    {
      IdentityType: 'EMAIL_ADDRESS',
      IdentityName: 'baz@example.com',
      SendingEnabled: false
    }
  ],
  NextToken: null
}

$ client.listEmailIdentities({PageSize: 1}).promise().then(console.log)
aws-sdk> {
  EmailIdentities: [
    {
      IdentityType: 'EMAIL_ADDRESS',
      IdentityName: 'bar@example.com',
      SendingEnabled: false
    }
  ],
[...]

PR #4219, opened 2022-09-27
fix: unset content length in rest-json when payload is undefined

trivikr:
Unsets content length in rest-json when payload is undefined

Previous PR: https://github.com/aws/aws-sdk-js/pull/3914
Internal JS-3610

##### Checklist

- [x] `npm run test` passes

### DIFF
--- a/lib/protocol/rest_json.js
+++ b/lib/protocol/rest_json.js
@@ -4,6 +4,16 @@ var Json = require('./json');
 var JsonBuilder = require('../json/builder');
 var JsonParser = require('../json/parser');
 
+function unsetContentLength(req) {
+  var payloadMember = util.getRequestPayloadShape(req);
+  if (
+    payloadMember === undefined &&
+    ['GET', 'HEAD', 'DELETE'].indexOf(req.httpRequest.method) >= 0
+  ) {
+    delete req.httpRequest.headers['Content-Length'];
+  }
+}
+
 function populateBody(req) {
   var builder = new JsonBuilder();
   var input = req.service.api.operations[req.operation].input;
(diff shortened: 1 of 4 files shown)


lib/protocol/rest_json.js
```diff
@@ -4,6 +4,16 @@ var Json = require('./json');
 var JsonBuilder = require('../json/builder');
 var JsonParser = require('../json/parser');
 
+function unsetContentLength(req) {
+  var payloadMember = util.getRequestPayloadShape(req);
+  if (
+    payloadMember === undefined &&
+    ['GET', 'HEAD', 'DELETE'].indexOf(req.httpRequest.method) >= 0
```
kuhe, line 11: any need to add HEAD/OPTIONS?

lib/protocol/rest_json.js
```diff
@@ -4,6 +4,16 @@ var Json = require('./json');
...
+  ) {
+    delete req.httpRequest.headers['Content-Length'];
```
kuhe, line 13: is this ever any other case than `Aaaa-Aaaa`?

lib/protocol/rest_json.js
trivikr: No as per internal RFC. On line 53, we avoid sending payload for GET, HEAD and DELETE

lib/protocol/rest_json.js
trivikr: Not that I know of. What's `Aaaa-Aaaa` case?

lib/protocol/rest_json.js
trivikr: Created a reusable array in commit [`2329b60` (#4219)](https://github.com/aws/aws-sdk-js/pull/4219/commits/2329b60df0850adcbac009375a23706b9a48fed4) to reduce duplicate code.

PR #4324, opened 2023-01-12
Fixed CognitoIdentityServiceProvider UserType definition

danielfx90:
Hi everyone!

`CognitoIdentityServiceProvider.UserType` definition has a property wrongly defined as `Attributes` when it is actually `UserAttributes`. That's it!

Source: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CognitoIdentityServiceProvider.html#getUser-property

Thank you!

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/.changes/next-release/bugfix-CognitoIdentityServiceProvider-1dbbc277.json
+++ b/.changes/next-release/bugfix-CognitoIdentityServiceProvider-1dbbc277.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "CognitoIdentityServiceProvider",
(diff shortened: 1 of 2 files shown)


aBurmeseDev:
Hi @danielfx90 - thanks for submitting the PR and apology for long silence. 

Upon reviewing PR, we cannot accept PR as this needs to be updated upstream API by service team and this PR would only change type declaration and not model which is owned by service team. 

I'm going to submit a request to service team to notify and close this for now.

PR #4332, opened 2023-02-02
chore(util): use custom MD5 checksum instead of native for FIPS environments

lukiano:
Logic copied from aws-sdk-js-v3

Let me know your thoughts on this. the `browserMd5` file also has a similar logic.

The checksum is used for integrity checks, at least in SQS and S3. Because this isn't about ensuring the data hasn't been tampered with, it should be fine in a FIPS environment. Besides, the V3 of the `aws-sdk` does the same.

##### Checklist

- [X] `npm run test` passes

### DIFF
--- a/.changes/next-release/feature-util-a775abff.json
+++ b/.changes/next-release/feature-util-a775abff.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "util",
+  "description": "Use custom Md5 checksum calculation instead of one based on crypto modules as the latter won't work in a FIPS-compliant environment"
+}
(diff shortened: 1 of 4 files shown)


lukiano:
Regretfully the test file https://github.com/aws/aws-sdk-js-v3/blob/main/packages/md5-js/src/index.spec.ts isn't straightforward to translate.

nguyenpc:
Should this be reviewed and merged? The S3 SSE-C approach won't work in FIPS mode without this.

kuhe:
For now, please overwrite the `AWS.util` function in question at runtime with your custom MD5 implementation if you need it in your environment.

aBurmeseDev:
@lukiano - thanks for submitting this PR. This's something we need to implement it across-SDK for consistency and we haven't decided which route we want to take. 

Anyone that comes across this, please refer to [above comment](https://github.com/aws/aws-sdk-js/pull/4332#issuecomment-1740276252) for workaround.

PR #4345, opened 2023-02-10
chore: add maintenance mode message

kuhe:
Adds a maintenance mode message for 2023.

I still wanted to add a suppression method, but not in the spec-derived config files or as an ENV variable. This suppression method is specific to JSv2, just like the message itself.

##### Checklist

- [n/a] `npm run test` passes
- [n/a] `.d.ts` file is updated
- [n/a] changelog is added, `npm run add-change`
- [n/a] run `bundle exec rake docs:api` and inspect `doc/latest/index.html` if documentation is changed
- [n/a] run `npm run integration` if integration test is changed
- [n/a] non-code related change (markdown/git settings etc)

### DIFF
--- a/lib/aws.js
+++ b/lib/aws.js
@@ -4,6 +4,7 @@ var AWS = require('./core');
 
 // Load all service classes
 require('../clients/all');
+require('./maintenance_mode_message');
 
 /**
  * @api private
--- a/lib/aws.js
+++ b/lib/aws.js
@@ -4,7 +4,6 @@ var AWS = require('./core');
 
 // Load all service classes
 require('../clients/all');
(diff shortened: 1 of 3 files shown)


siddsriv (approved):

trivikr (approved):

DavidWells:
This message is messing with a lot of consuming CLIs. I'd consider reverting this or giving users a way to suppress this message for consumers of their CLIs

https://github.com/serverless/serverless/issues/11753

kuhe:
There is a suppression method in this PR. I will mention it in https://github.com/serverless/serverless/issues/11753.

PR #4364, opened 2023-03-09
Upgrade buffer dependency to 6.0.3

solancer:
Using 4.9.2 version of the buffer package package causes an error on Lambda

Since it's a console.error, this logs to AWS CloudWatch every time a request is made.  This problem exists only on Lambda, on local, it doesn't show any such errors.

`ERROR    (node:8) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
`

Screenshot 2023-03-10 at 1 41 43 AM
Screenshot 2023-03-10 at 1 40 08 AM
Screenshot 2023-03-10 at 1 39 56 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "webpack": "^1.15.0"
   },
   "dependencies": {
-    "buffer": "4.9.2",
+    "buffer": "6.0.3",
     "events": "1.1.1",
     "ieee754": "1.1.13",
     "jmespath": "0.16.0",


kellertk:
Thank you for your efforts, but this project is no longer accepting contributions as part of our plan to end support for AWS SDK for JavaScript v2. Looking into this further, it doesn't look as though this is a critical security issue, instead merely a warning. We encourage you to migrate to v3 of the JavaScript SDK.

PR #4404, opened 2023-04-26
fix(sqs): adapt to protocol change in sqs

kuhe:
Modify SQS tests to handle either XML or JSON protocol.

The service itself also appears to need a fix (see diff). This indicates there might be additional unforeseen consequences in the protocol change. 

##### Checklist

- [x] `npm run test` passes
- n/a `.d.ts` file is updated
- n/a run `npm run integration` if integration test is changed


lib/services/sqs.js
```diff
@@ -118,7 +118,8 @@ AWS.util.update(AWS.SQS.prototype, {
    * @api private
    */
   buildEndpoint: function buildEndpoint(request) {
-    var url = request.httpRequest.params.QueueUrl;
+    var params = request.httpRequest.params || this.config.params;
+    var url = params.QueueUrl;
```
kuhe, line 122: in JSON protocol, request.httpRequest.params no longer contained any QueueUrl in testing when the client was initialized with

```js
new AWS.SQS({
  params: { QueueUrl: ... }
});
```

kuhe: The change on this line is needed if this is a way that users pass QueueUrl to client operations, and not merely a testing mechanism.

kuhe: After investigation, I believe this change is good to go as-is. 

The query protocol assigns client init params to each httpRequest but the JSON protocol does not. Rather than modify the protocols behaviors, modifying the behavior of the SQS customizations is the better option for limiting potential impact.

trivikr (approved):

PR #4405, opened 2023-04-26
fix(sqs): add queryCompatibility to json protocol

kuhe:
##### Checklist

- [x] `npm run test` passes
- n/a `.d.ts` file is updated

Ran `npx cucmber-js -t @sqs` this time for both current and new model JSON.

### DIFF
--- a/lib/services/sqs.js
+++ b/lib/services/sqs.js
@@ -5,7 +5,7 @@ AWS.util.update(AWS.SQS.prototype, {
    * @api private
    */
   setupRequestListeners: function setupRequestListeners(request) {
-    request.addListener('build', this.buildEndpoint.bind(this));
+    request.addListener('build', this.buildEndpoint);
 
     if (request.service.config.computeChecksums) {
       if (request.operation === 'sendMessage') {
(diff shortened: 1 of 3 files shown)


trivikr (approved):

lib/services/sqs.js
```diff
@@ -118,8 +118,27 @@ AWS.util.update(AWS.SQS.prototype, {
...
+    var self = this;
+
+    var params = (function () {
+      // Query protocol client or request level parameters
+      if (request.httpRequest.params && request.httpRequest.params.QueueUrl) {
+        return request.httpRequest.params;
+      }
+      // JSON protocol with config QueueUrl.
+      if (self.config.params && self.config.params.QueueUrl) {
+        return self.config.params;
```
kuhe, line 130: But if users from the Query version put QueueUrl into params, it does not get serialized into the body, it only exists on this config.params object. 

In contrast, a client param in Query is serialized into its outgoing requests.

kuhe: resolved in revision: added query compat customization to JSON protocol

trivikr (approved):

kuhe:
`npx cucumber-js -t @sqs` is also passing with both JSON and Query protocols.

PR #4421, opened 2023-05-15
feat(errors): add error.body with original JSON error body

kuhe:
Add `error.body` hidden field with the original JSON body of the error response.

##### Checklist
- [x] `npm run test` passes
- n/a `.d.ts` file is updated


trivikr (changes requested):
Blocked on checklng behavior in other SDKs which does not support modeled errors

trivikr (approved):

PR #4422, opened 2023-05-17
fix(IAM): endpoint in aws-iso partition

pinak:
Fix endpoint for IAM for us-iso partition

IAM is partitional service, previously it was being treated as regional.

##### Checklist

- [ ] `npm run test` passes -> tests don't even pass in master branch for me locally

### DIFF
--- a/lib/region_config_data.json
+++ b/lib/region_config_data.json
@@ -46,7 +46,7 @@
     "us-iso-*/iam": {
       "endpoint": "{service}.us-iso-east-1.c2s.ic.gov",
       "globalEndpoint": true,
-      "signingRegion": "us-east-1"
+      "signingRegion": "us-iso-east-1"
     },
 
     "us-gov-*/iam": "globalGovCloud",
--- a/lib/region_config_data.json
+++ b/lib/region_config_data.json
@@ -43,8 +43,13 @@
       "globalEndpoint": true,
       "signingRegion": "cn-north-1"
     },
-    "us-gov-*/iam": "globalGovCloud",
+    "us-iso-*/iam": {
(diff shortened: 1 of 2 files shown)


lib/region_config_data.json
```diff
@@ -43,8 +43,13 @@
       "globalEndpoint": true,
       "signingRegion": "cn-north-1"
     },
-    "us-gov-*/iam": "globalGovCloud",
+    "us-iso-*/iam": {
+      "endpoint": "{service}.us-iso-east-1.c2s.ic.gov",
+      "globalEndpoint": true,
+      "signingRegion": "us-east-1"
```
trivikr, line 49: Should this be us-iso-east-1?
```suggestion
      "signingRegion": "us-iso-east-1"
```

trivikr: Verified that it should be `us-iso-east-1` from Endpoints 2.0 tests https://github.com/aws/aws-sdk-js-v3/blob/7ed7101dcc4e81038b6c7f581162b959e6b33a04/codegen/sdk-codegen/aws-models/iam.json#L2270-L2291

trivikr:
Verified that correct endpoint is logged for IAM

```js
import AWS from "../aws-sdk-js/lib/aws.js";

const region = "us-iso-west-1";
const client = new AWS.IAM({ region });

// Logs iam.us-iso-east-1.c2s.ic.gov
console.log(`JS v2 endpoint for region '${region}':`, client.endpoint.host);
```

lib/region_config_data.json
pinak: Yes, thanks for catching that!

Issue #4303, opened 2022-12-20
Cognito unauthenticated user not authorized to perform: ssm:GetParameter because no session policy allows


hienbx-0488:
### Describe the bug

I use cognito identity pool with enable access to unauthenticated identities .
For Unauthenticated role, i created new role with policy AmazonS3FullAccess and inline policy for get SSM Parameter store
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "ssm:*"
            ],
            "Resource": "*"
        }
    ]
}
```

I use blow code to get list bucket S3 and SSM Paramter Store
```js
var AWS = require('aws-sdk');

var IDENTITY_POOL_ID = 'my_identity_pool_id';
var REGION = 'ap-northeast-1';

// Initialize the Amazon Cognito credentials provider
AWS.config.region = REGION; // Region
AWS.config.credentials = new AWS.CognitoIdentityCredentials({IdentityPoolId: IDENTITY_POOL_ID,});


// Create S3 service object
s3 = new AWS.S3({apiVersion: '2006-03-01'});

// Call S3 to list the buckets
s3.listBuckets(function(err, data) {
  if (err) {
    console.log("Error", err);
  } else {
    console.log("Success", data.Buckets);
  }
});


var ssm = new AWS.SSM({});
var params = {
  Name: 'abc', /* required */
  WithDecryption: false
};
ssm.getParameter(params, function(err, data) {
  if (err) console.log(err, err.stack); // an error occurred
  else     console.log(data);           // successful response
});
```

After run code, list bucket S3 return success, but SSM parameter store return error
```
{
  code: 'AccessDeniedException',
  time: 2022-12-20T08:05:41.500Z,
  requestId: 'bd8d4f26-7fd3-4bd3-ae40-dfaab59e62fb',
  statusCode: 400,
  retryable: false,
  retryDelay: 78.72879079767343
} AccessDeniedException: User: arn:aws:sts::****:assumed-role/bc-CognitoAnonymousGroupRol-***/CognitoIdentityCredentials is not authorized to perform: ssm:GetParameter on resource: arn:aws:ssm:ap-northeast-1:****:parameter/abc because no session policy allows the ssm:GetParameter action
```

### Expected Behavior

Return success SSM parameter store

### Current Behavior

Error when get SSM parameter store

### Reproduction Steps

```
var AWS = require('aws-sdk');

[...]

PR #4455, opened 2023-06-22
docs(cognito_identity_credentials): Explain limitation of CognitoIden…

RanVaknin:
This method uses the simplified authflow, which means users cannot explicitly set a session policy. This results in STS appending the Default Session Policy to the request, which can lead to limiting permissions even if they are explicitly attached to a role.

The doc change highlights the limitation of the connivence method since it uses that Enhanced (simplified) flow "under the hood" which limits permissions on the assumed role because of the intersection of multiple policies resulting in the least permissive one being applied to the credentials returned from STS.

[Related ticket ](https://github.com/aws/aws-sdk-js/issues/4303)

### DIFF
--- a/lib/credentials/cognito_identity_credentials.js
+++ b/lib/credentials/cognito_identity_credentials.js
@@ -27,7 +27,7 @@ var STS = require('../../clients/sts');
  * implementation calls Cognito's `getId()` and `GetCredentialsForIdentity()`.
  * In this flow there is no way to explicitly set a session policy, resulting in
  * STS attaching the default policy and limiting the permissions of the federated role.
- * To be able to explicitly set a session policy, do not use this convinience method. 
+ * To be able to explicitly set a session policy, do not use this convinience method.
  * Instead, you can use the Cognito client to call `getId()`, `GetOpenIdToken()` and then use
  * that token with your desired session policy to call STS's `AssumeRoleWithWebIdentity()`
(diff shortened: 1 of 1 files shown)


kuhe (approved):

lib/credentials/cognito_identity_credentials.js
```diff
@@ -23,6 +23,15 @@ var STS = require('../../clients/sts');
  * identity providers. See {constructor} for an example on creating a credentials
  * object with proper property values.
  *
+ * DISCLAIMER: This convinience method leverages the Enhanced (simplified) Authflow. The underlying
```
kuhe, line 26:
```suggestion
 * DISCLAIMER: This convenience method leverages the Enhanced (simplified) Authflow. The underlying
```

lib/credentials/cognito_identity_credentials.js
```diff
@@ -23,6 +23,15 @@ var STS = require('../../clients/sts');
...
+ * implementation calls Cognito's `getId()` and `GetCredentialsForIdentity()`.
+ * In this flow there is no way to explicitly set a session policy, resulting in
+ * STS attaching the default policy and limiting the permissions of the federated role.
+ * To be able to explicitly set a session policy, do not use this convinience method.
```
kuhe, line 30:
```suggestion
 * To be able to explicitly set a session policy, do not use this convenience method.
```

PR #4529, opened 2023-11-17
chore: deprecate macie for macie2

kuhe:
removes the macie client.


trivikr (approved):

kuhe:
https://docs.aws.amazon.com/macie/latest/userguide/what-is-macie.html

PR #4541, opened 2023-12-02
Default empty array for previously flattened lists and maps in query compatible services

siddsriv:
##### Checklist


RanVaknin (commented):
I think the changes need to be made in the JSON deserializer since we are trying to mimic the behavior of the XML deserializer, not change it.

PR #4542, opened 2023-12-03
fix: adding default value for list members for awsQueryCompatible for…

RanVaknin:
- [x] `npm run test` passes
- [x] `.d.ts` file is updated

### DIFF
--- a/lib/json/parser.js
+++ b/lib/json/parser.js
@@ -23,12 +23,17 @@ function translateStructure(structure, shape) {
 
   var struct = {};
   var shapeMembers = shape.members;
+  var isAwsQueryCompatible = shape.api && shape.api.awsQueryCompatible;
   util.each(shapeMembers, function(name, memberShape) {
     var locationName = memberShape.isLocationName ? memberShape.name : name;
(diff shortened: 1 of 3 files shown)


kuhe (approved):

siddsriv (approved):

lib/json/parser.js
```diff
@@ -23,12 +23,17 @@ function translateStructure(structure, shape) {
...
   util.each(shapeMembers, function(name, memberShape) {
     var locationName = memberShape.isLocationName ? memberShape.name : name;
     if (Object.prototype.hasOwnProperty.call(structure, locationName)) {
       var value = structure[locationName];
       var result = translate(value, memberShape);
       if (result !== undefined) struct[name] = result;
+    } else if (isAwsQueryCompatible && memberShape.defaultValue) {
+      if (memberShape.type === 'list') {
+        struct[name] = typeof memberShape.defaultValue === 'function' ? memberShape.defaultValue() : memberShape.defaultValue;
+      }
```
kuhe, line 36: the reason we only apply this to lists and not maps is because the XML behavior was only applied to lists

kstich (approved):

PR #4577, opened 2024-01-23
fix: add string fallback for S3#Expires field when not a date-time

kuhe:
See https://github.com/aws/aws-sdk-js-v3/pull/5715.

Adds an S3-only model transform that adds an additional field `ExpiresString: string` where any operation output has `Expires: Date` as an output member. This field will contain the raw value from Expires response header in all cases. 

If `Expires` response header is not a date format, this will no longer throw an exception. Users can access the `ExpiresString` field as a fallback, and `Expires` will be set to `undefined` in that case.

##### Checklist

- [x] `npm run test` passes
- [x] `.d.ts` file is updated


kuhe:
testing:

```js
// in v2/workspace/s3 folder
import S3 from "../../clients/s3";

var s3 = new S3({
  region: 'us-west-2'
});

(async () => {
  // both Expires and ExpiresString will be set on the returned object.
  const get = await s3.getObject({
    Bucket: "...",
    Key: "goodexpires",
  }).promise();

  console.log("expires and expiresString", get.Expires, get.ExpiresString);

  // this emits a warning, but does not throw. Expires becomes undefined and only ExpiresString is set.
  const get2 = await s3.headObject({
    Bucket: "...",
    Key: "badexpires",
  }).promise();

  console.log("expires and expiresString", get2.Expires, get2.ExpiresString);
})();

```

siddsriv (approved):

Issue #4580, opened 2024-02-02
ReferenceError in SageMaker invokeEndpointWithResponseStream


cshen-dev:
### Describe the bug

I created an AWS SageMaker endpoint with Mistral Jumpstarter (jumpstart-dft-hf-llm-mistral-7b-ins-20240130).

`sagemakerRuntime.invokeEndpoint` worked as a charm however when I started exploring its stream capability, I met the issue below.

```console
Error [ReferenceError]: payload is not defined
    at Request.extractData (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/protocol/rest_json.js:78:17)
    at Request.callListeners (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/request.js:688:12)
    at Request.callListeners (/Users/c.shen/Workspaces/tmp/sagemaker/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
  code: 'ReferenceError',
  time: 2024-02-02T00:39:53.693Z,
  statusCode: 200,
  retryable: false,
  retryDelay: 62.00239553897755
}
```

### Expected Behavior

```console
[ReferenceError]: payload is not defined
```

^ This error should not occur which I guess is caused by mis-mapped data structure between SageMaker endpoint's Response and the sdk client's Request.

### Current Behavior

It's throwing error and block the downstream processing

[...]

PR #4607, opened 2024-03-08
Use rules.payload when extracting data for event stream

inf3rnus:
Fixes https://github.com/aws/aws-sdk-js/issues/4580, problem is that there were undefined variables in `lib/protocol/rest_json.js`

File could use more cleanup as well, as there are also unused variables from whoever touched it last.

Without this change, you cannot use the client to get a streamed sagemaker request.

##### Checklist

- [x ] `npm run test` passes (tests seem flaky, less are failing after this change, and failures change with each run, before and after the change. Your guys' judgement call)
- [x ] run `npm run integration` if integration test is changed (tests are flaky and also just fail, always fail on my machine before and after change, didn't change them)

### DIFF
--- a/lib/protocol/rest_json.js
+++ b/lib/protocol/rest_json.js
@@ -1,3 +1,4 @@
+var AWS = require('../core');
 var util = require('../util');
 var Rest = require('./rest');
 var Json = require('./json');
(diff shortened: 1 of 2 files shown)


lib/protocol/rest_json.js
```diff
@@ -75,7 +76,7 @@ function extractData(resp) {
     var body = resp.httpResponse.body;
     if (payloadMember.isEventStream) {
       parser = new JsonParser();
-      resp.data[payload] = util.createEventStream(
+      resp.data[rules.payload] = util.createEventStream(
```
trivikr, line 79: Looks like this was a miss in https://github.com/aws/aws-sdk-js/pull/2085, likely while doing copy-paste from `rest-xml.js`

trivikr (approved):

kuhe (approved):

